PR #183, opened 2026-05-13
StratifiedSampler returns xi only

imerad:
### Description

- What does this PR do?
See title
Deals with this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=184758328)

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ## [Next release]
 
 ### ✨ Added
-<<<<<<< feat/uniform_sampler
 - `UniformSampler` for uniform random sampling
-=======
 - Example for `IPWPTDMeanEstimator`
->>>>>>> main
 - Section about IPWPTD in the user guide
 - `IPWPTDMeanEstimator` for inverse probability weighted PTD mean estimation
 - Section about `CostOptimalRandomSampler` in the user guide
@@ -21,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - `StratifiedPTDMeanEstimator` stratified extension of `PTDMeanEstimator`
 
 ### 🔄 Changed
+- `StratifiedSampler` now returns sampling indicators only
 - Refactored `glide.core.simulated_datasets` into dedicated `glide.simulators` module with separate files for each generator function
 - Replaced "semi-supervised" with "prediction-powered" everywhere in the docs and code
 
--- a/docs/deep_dive/icml_case_study.ipynb
+++ b/docs/deep_dive/icml_case_study.ipynb
@@ -212,7 +212,7 @@
     "        y_true[xi] = y_true_oracle[xi]\n",
     "        return y_true, y_proxy\n",
     "    elif workflow == \"Stratified\":\n",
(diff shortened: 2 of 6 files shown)


imerad (commented):
## Code Review

### Overview

This PR changes `StratifiedSampler.sample()` from returning a `(pi, xi)` tuple to returning only `xi`, and switches the sampling mechanism from per-sample Bernoulli draws (with probability `pi_i = n_h / N_h`) to exact without-replacement sampling (`rng.choice(..., replace=False)`). All call sites — tests, functional tests, and the notebook — are updated consistently.

This is a meaningful improvement: the new mechanism guarantees each stratum receives exactly `n_h` annotations, whereas Bernoulli sampling could produce 0 annotations from a stratum in the worst case.

---

### Issues

**1. `np.argwhere` returns 2D, should use `np.flatnonzero`**

```python
# Current (glide/samplers/stratified.py:219)
stratum_indices = np.argwhere(groups == stratum_id)   # shape (N_h, 1) — 2D
selected_samples = rng.choice(stratum_indices, size=n_h, replace=False)  # shape (n_h, 1)
xi[selected_samples] = 1  # works via fancy indexing, but opaque
```

`np.argwhere` on a 1D boolean mask returns a 2D array of shape `(N_h, 1)`. `rng.choice` samples rows from it, yielding `selected_samples` of shape `(n_h, 1)`. The assignment to `xi` happens to work because NumPy's advanced indexing flattens the index array, but it is not obvious and could silently misbehave under a refactor.

```python
# Cleaner
stratum_indices = np.flatnonzero(groups == stratum_id)   # shape (N_h,) — 1D
selected_samples = rng.choice(stratum_indices, size=n_h, replace=False)  # shape (n_h,)
xi[selected_samples] = 1
```

**2. `Raises` section in `sample()` is incomplete**

The body raises `ValueError` when a stratum receives zero allocation (lines 207–212), but this case is absent from the docstring's `Raises` section. Either document it, or — if it's considered an internal invariant that users cannot trigger — move the check earlier and make that clear.

**3. CHANGELOG.md not updated**

The checklist leaves this unchecked, but this is a breaking API change (return type `Tuple[NDArray, NDArray]` → `NDArray`). Any caller that was doing `pi, xi = sampler.sample(...)` will now get a `ValueError` at runtime. Per project conventions, this must appear under `Changed` (or `Breaking`) in CHANGELOG.md.

**4. `test_sample_proportional_strategy` only tests the all-selected edge case**

```python
# tests/unit/samplers/test_stratified.py:113-117
def test_sample_proportional_strategy(sampler, y_proxy, groups):
    xi = sampler.sample(y_proxy, groups, 8, strategy="proportional", random_seed=0)
    expected_xi = np.ones(8)
    np.testing.assert_array_equal(xi, expected_xi)
```

Budget 8 on an 8-element fixture means every sample is selected. This passes trivially for any correct implementation. The functional test covers the proportional allocation property, but a unit test with a partial budget (e.g., budget=4) would catch implementation bugs that the current test misses.

---

### Minor

- The `Tuple` import at line 1 remains used (`_validate` return type), so no spurious import — good.
- The updated class-level docstring is accurate and clear.
- Functional test rewrites (`test_proportional_matches_uniform_equal_strata`, `test_sample_rounding_sums_to_budget`) are tighter and test actual behavior rather than intermediate `pi` values — a net improvement.

gmartinon-ed (approved):

PR #204, opened 2026-05-19
Implement cost optimal sampler

### LINKED ISSUE #184: Implement CostOptimalSampler
## Scientific Background

This sampler implements **Proposition 2** from:

> Angelopoulos, A. N., Eisenstein, J., Berant, J., Agarwal, A., and Fisch, A. (2025).
> *Cost-Optimal Active AI Model Evaluation.* arXiv:2506.07949, §2.3 — Optimal Active Annotation.

---

### How This Relates to `CostOptimalRandomSampler`

`CostOptimalRandomSampler` (Proposition 1) annotates each record with the **same** probability `π` regardless
of the record's content. It is optimal among constant policies.

`CostOptimalSampler` (Proposition 2) is strictly more powerful: it annotates each record with a probability
`π(x)` that **depends on x**. It exploits *heteroscedasticity* — the fact that the cheap proxy rater G is
more wrong on some inputs than others. Records where G is likely to be wrong get a higher annotation
probability; records where G is likely to be right get a lower probability (or none at all). The gain over
the random policy is large when some records are systematically harder than others.

---

### The Key Quantity: `u(x)`

The policy is driven by a function `u(x)` defined as the **conditional mean squared error** of G given x:

```
u(x) = E[(H - G)² | X = x]
```

Intuitively, `u(x)` is the expected squared disagreement between the expensive rater H and the cheap proxy G
*when the input is x*. A high `u(x)` means G is unreliable for this kind of input, so it is worth paying for
H. A low `u(x)` means G is reliable, so we can skip H.

In practice, `u(x)` cannot be computed exactly — it must be estimated. Common choices:
- A regression model trained on a burn-in dataset to predict `(h_i - g_i)²` from `x_i`.
- A model confidence score or entropy (a rough proxy for disagreement).

**This class does not learn `u(x)` internally.** The caller pre-computes the per-record uncertainty values
and passes them as a 1D array directly to `sample()`. This avoids a hard dependency on any ML framework
(scikit-learn, PyTorch, etc.) and lets the caller choose the approach that fits their pipeline.

---

### The Optimal Policy (Proposition 2)

The policy is:

```
π_active(x) = π_clip(x; τ*)
```

where `π_clip` is the **scaled-and-clipped** policy:

```
              ⎧ γ*(τ) · sqrt(u(x))    if sqrt(u(x)) ≤ τ
π_clip(x; τ) = ⎨
              ⎩ 1                      otherwise
```

**Interpretation:** records with high uncertainty (`sqrt(u(x)) > τ`) are always annotated by H.
Records with low uncertainty are annotated with probability proportional to their uncertainty, scaled by γ*.
The scaling factor γ*(τ) is computed analytically:

```
γ*(τ) = min(
    sqrt( (c_g/c_h + P(U > τ²)) / max(Var(H) - E[U · 1{U ≤ τ²}], 0) ),
    1 / τ
)
```

where `U = u(X)` is the random variable for conditional error, and all expectations are approximated
empirically over the pool of records passed to `sample()`.

The notation `max(..., 0)` is the **positive part**, written `(...)_+` in the paper. It prevents division by
zero when the denominator would otherwise go negative (which can happen for extreme τ values). When the
positive part is zero, γ*(τ) = 1/τ (the upper bound), meaning all records in the "proportional" region get
clamped to 1, so the policy effectively becomes "always annotate".

---

### Finding the Optimal Threshold τ*

The threshold `τ*` is found by minimising the product of **expected cost per record** and **expected error**:

```
τ* = argmin over τ > 0 of:
     [c_h · E[π_clip(X; τ)] + c_g] · [Var(H) + E[U · (π_clip(X; τ)⁻¹ - 1)]]
```

Per Remark 3 of the paper, this objective is **non-convex** in general, but because it is one-dimensional,
a simple grid search suffices. The natural grid is the sorted distinct values of `sqrt(u_i)` — these are the
breakpoints at which the objective changes character (a record transitions from the proportional region to the
"always annotate" region as τ decreases through `sqrt(u_i)`). Evaluating the objective at each such breakpoint
is sufficient.

**Important:** `π_clip(x; τ)⁻¹` is the inverse probability. For records where `π_clip = 1` this is just 1,
and the term `U · (1 - 1) = 0` contributes nothing to the error. For records in the proportional region,
`π_clip = γ*(τ) · sqrt(u(x))`, so `π_clip⁻¹ = 1 / (γ*(τ) · sqrt(u(x)))`.

---

### Budget and Record Selection

Once τ* is known and all per-record probabilities `π_i = π_clip(x_i; τ*)` are computed:

1. **Expected cost per record** = `c_h · mean(π_i) + c_g`
2. **Number of affordable records**: `T = floor(budget / cost_per_record)`
3. If `T < N`: draw T records **uniformly at random without replacement** from the N-record pool.
4. If `T ≥ N`: use all N records.
5. For each selected record `i`, independently draw `ξ_i ~ Bernoulli(π_i)` — the indicator for whether H
   was actually queried.

The uniform record-level draw in step 3 is the same mechanism used by `CostOptimalRandomSampler`. The
heteroscedastic signal only affects the *per-record* annotation probability, not which records are selected.

---

### Constructor

```python
class CostOptimalSampler:
    def __init__(self) -> None:
```

### `fit` — burn-in phase

Same role as in `CostOptimalRandomSampler`: the caller passes labeled burn-in records to let the sampler
estimate `Var(H)`. The burn-in data must be fully labeled (no NaN). The burn-in arrays should be preserved
by the caller for downstream estimation (they are reusable via inverse-variance weighting with the active-phase
estimates, exactly as in the random sampler case).

```python
def fit(
    self,
    y_true: NDArray,
) -> "CostOptimalSampler":
    """
    y_true: 1D float array, no NaN.
    Estimates Var(H) = np.var(y_true, ddof=1) and stores it for use in sample().
    Raises ValueError if y_true is empty, contains NaN, or Var(H) == 0 (degenerate labels).
    """
```

Note: unlike `CostOptimalRandomSampler.fit()`, we accept only `y_true` — `y_proxy` is not needed
because the role of MSE is fully taken over by the per-record `u(x)` values supplied in `sample()`.

### `sample` — active sampling phase

All parameters are **mandatory**. `y_true_cost` and `y_proxy_cost` must be strictly positive floats.
`budget` must be a strictly positive int. `random_seed` is mandatory (no default). `uncertainties` is a
1D float array of shape `(N,)` containing the pre-computed conditional MSE estimate `u(x_i)` for each record.

```python
def sample(
    self,
    uncertainties: NDArray,
    y_true_cost: float,
    y_proxy_cost: float,
    budget: int,
    random_seed: int,
```

Returns a tuple `(indices, xi, pi)`:
- `indices` — 1D int array of shape `(T,)`: selected record positions in `X`, sorted ascending.
- `xi` — 1D float array of shape `(T,)`: 1.0 if the expensive rater was queried, 0.0 otherwise.
- `pi` — 1D float array of shape `(T,)`: per-record annotation probability for each selected record.

Unlike `CostOptimalRandomSampler`, `pi` is a **per-record array** (not a scalar), because the active policy
assigns different probabilities to different records.

---

### `glide/samplers/cost_optimal.py`

```python
import math
from typing import Tuple

import numpy as np

class CostOptimalSampler:
    """Sampler that draws elements with probabilities proportional to uncertainty scores.

    Implements the cost-optimal active annotation policy. This
    sampler assigns a *per-record* annotation probability that scales with the
    conditional mean squared error ``u(x) = E[(H - G)² | X = x]``. Records where
    the cheap proxy G is likely to be wrong receive a higher probability of being
    annotated by the expensive rater H; records where G is reliable are sampled
    less often. This concentrates the annotation budget where it matters most.

    The caller pre-computes per-record conditional error estimates and passes them
    as a 1D array to ``sample()``. This class does not learn ``u(x)`` internally.

    References
    ----------
    Angelopoulos, A. N., Eisenstein, J., Berant, J., Agarwal, A., and Fisch, A.
    (2025). Cost-Optimal Active AI Model Evaluation. arXiv:2506.07949, §2.3.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.samplers.cost_optimal import CostOptimalSampler
    >>> rng = np.random.default_rng(0)
    >>> y_true = rng.normal(loc=2.0, scale=1.0, size=20)
    >>> uncertainties = np.full(100, 0.25)
    >>> sampler = CostOptimalSampler()
    >>> _ = sampler.fit(y_true)
    >>> indices, xi, pi = sampler.sample(uncertainties, y_true_cost=10.0, y_proxy_cost=1.0, budget=50, random_seed=0)
    >>> xi.shape == indices.shape == pi.shape
    True
    """

    def _validate_fit_inputs(self, y_true: NDArray) -> None:
        if len(y_true) == 0:
            raise ValueError("'y_true' must be non-empty.")
        if np.any(np.isnan(y_true)):
            raise ValueError(
                "'y_true' contains NaN values. The burn-in dataset must be fully labeled."
            )

    def fit(
        self,
        y_true: NDArray,
    ) -> "CostOptimalSampler":
        """Estimate Var(H) from a fully-labeled burn-in dataset.

        Corresponds to Policy A2 from Angelopoulos et al. (2025): the burn-in
        records are annotated unconditionally (π = 1) to bootstrap the sampler.
        The caller must preserve these arrays for downstream estimation via
        inverse-variance weighting with the active-phase estimates.

        Parameters
        ----------
        y_true : NDArray
            1D float array of expensive-rater labels from the burn-in phase. No NaN.

        Returns
        -------
        CostOptimalSampler
            The fitted sampler (returns ``self`` for method chaining).

        Raises
        ------
        ValueError
            If ``y_true`` is empty, contains NaN, or ``Var(H) == 0`` (all labels are identical).
        """
        self._validate_fit_inputs(y_true)
        y_true_variance = np.var(y_true, ddof=1)
        if y_true_variance == 0.0:
            raise ValueError(
                "Var(H) is zero — all labels are identical. "
                "Estimation is impossible with degenerate labels."
            )
        self._y_true_variance = y_true_variance
        return self

    def _compute_gamma(
        self,
        tau: float,
        uncertainties: NDArray,
        y_true_cost: float,
        y_proxy_cost: float,
    ) -> float:
        cost_ratio = y_proxy_cost / y_true_cost
        above_mask = uncertainties > tau ** 2
        prob_above = np.mean(above_mask)
        e_u_below = np.mean(uncertainties * ~above_mask)
        denominator = max(self._y_true_variance - e_u_below, 0.0)
        gamma_uncapped = np.sqrt((cost_ratio + prob_above) / denominator) if denominator > 0.0 else np.inf
        gamma = min(gamma_uncapped, 1.0 / tau)
        return gamma

    def _compute_per_record_probabilities(
        self,
        tau: float,
        gamma: float,
        uncertainties: NDArray,
    ) -> NDArray:
        sqrt_u = np.sqrt(uncertainties)
        probabilities = np.minimum(gamma * sqrt_u, 1.0)
        return probabilities

    def _compute_objective(
        self,
        tau: float,
        uncertainties: NDArray,
        y_true_cost: float,
        y_proxy_cost: float,
    ) -> float:
        gamma = self._compute_gamma(tau, uncertainties, y_true_cost, y_proxy_cost)
        pi_values = self._compute_per_record_probabilities(tau, gamma, uncertainties)

        mean_pi = np.mean(pi_values)
        cost_term = y_true_cost * mean_pi + y_proxy_cost

        safe_pi = np.where(pi_values > 0.0, pi_values, 1.0)
        error_term = self._y_true_variance + np.mean(uncertainties * (1.0 / safe_pi - 1.0))

        objective = cost_term * error_term
        return objective

    def _find_optimal_threshold(
        self,
        uncertainties: NDArray,
        y_true_cost: float,
        y_proxy_cost: float,
    ) -> float:
        sqrt_u_values = np.sqrt(uncertainties)
        candidates = np.unique(sqrt_u_values)
        objectives = [
            self._compute_objective(tau, uncertainties, y_true_cost, y_proxy_cost)
            for tau in candidates
        ]
        optimal_tau = candidates[int(np.argmin(objectives))]
        return optimal_tau

    def sample(
        self,
        uncertainties: NDArray,
        y_true_cost: float,
        y_proxy_cost: float,
        budget: int,
        random_seed: int,
        """Sample records and draw annotation indicators under the active policy.

        Applies the cost-optimal active annotation policy (Proposition 2). The
        optimal clipping threshold τ* is found by a grid search over the empirical
        distribution of ``uncertainties`` values. Per-record probabilities
        ``π_i = π_clip(x_i; τ*)`` are then used to determine how many records can
        be afforded within ``budget`` and to draw Bernoulli annotation indicators.

        Parameters
        ----------
        uncertainties : NDArray
            1D float array of shape ``(N,)`` containing the pre-computed conditional
            MSE estimate ``u(x_i) = E[(H - G)² | X = x_i]`` for each record.
            All values must be non-negative and not all zero.
        y_true_cost : float
            Cost of one expensive-rater annotation. Must be strictly positive.
        y_proxy_cost : float
            Cost of one cheap-proxy annotation. Must be strictly positive.
        budget : int
            Total annotation budget. Must be strictly positive.
        random_seed : int
            Seed for ``numpy.random.default_rng``. Mandatory for reproducibility.

        Returns
        -------
        Tuple[NDArray, NDArray, NDArray]
            A tuple ``(indices, xi, pi)`` where:

            - ``indices`` — 1D int array of shape ``(T,)``: positions of selected
              records in ``uncertainties``, sorted ascending.
            - ``xi`` — 1D float array of shape ``(T,)``: Bernoulli indicators;
              ``1.0`` if the expensive rater was queried, ``0.0`` otherwise.
            - ``pi`` — 1D float array of shape ``(T,)``: per-record annotation
              probability for each selected record.

        Raises
        ------
        ValueError
            If ``fit()`` has not been called, if any cost or budget argument is
            non-positive, if ``uncertainties`` contains NaN or negative values,
            if all uncertainty values are zero, or if the budget is too small to
            afford a single record.
        """
        if not hasattr(self, "_y_true_variance"):
            raise ValueError("Call fit() before sample().")
        if y_true_cost <= 0.0:
            raise ValueError(f"'y_true_cost' must be strictly positive; got {y_true_cost}.")
        if y_proxy_cost <= 0.0:
            raise ValueError(f"'y_proxy_cost' must be strictly positive; got {y_proxy_cost}.")
        if budget <= 0:
            raise ValueError(f"'budget' must be strictly positive; got {budget}.")

        if np.any(np.isnan(uncertainties)):
            raise ValueError(
                "'uncertainties' contains NaN values. All conditional error estimates must be finite."
            )
        if np.any(uncertainties < 0.0):
            raise ValueError(
                "'uncertainties' contains negative values. "
                "A conditional MSE estimate u(x) cannot be negative."
            )
        if np.all(uncertainties == 0.0):
            raise ValueError(
                "All values in 'uncertainties' are zero. "
                "The proxy perfectly predicts the expensive rater on every record; "
                "there is no benefit from annotating with H."
            )

        tau_star = self._find_optimal_threshold(uncertainties, y_true_cost, y_proxy_cost)
        gamma_star = self._compute_gamma(tau_star, uncertainties, y_true_cost, y_proxy_cost)
        pi = self._compute_per_record_probabilities(tau_star, gamma_star, uncertainties)

        cost_per_record = y_true_cost * np.mean(pi) + y_proxy_cost
        T = math.floor(budget / cost_per_record)
        if T < 1:
            raise ValueError(
                f"Budget {budget} is too small to afford a single record "
                f"at cost_per_record={cost_per_record:.4f}."
            )

        N = len(uncertainties)
        if T < N:
            indices = np.sort(rng.choice(N, size=T, replace=False))
        else:
            indices = np.arange(N)

        selected_pi = pi[indices]
        xi = rng.binomial(n=1, p=selected_pi, size=len(indices)).astype(float)
        return indices, xi, selected_pi
```

### `glide/samplers/__init__.py` — updated export

```python
from glide.samplers.active import ActiveSampler
from glide.samplers.cost_optimal import CostOptimalSampler
from glide.samplers.cost_optimal_random import CostOptimalRandomSampler
from glide.samplers.stratified import StratifiedSampler

__all__ = ["ActiveSampler", "CostOptimalRandomSampler", "CostOptimalSampler", "StratifiedSampler"]
```

## Edge Cases

| Condition | Behaviour |
|---|---|
| `Var(H) == 0` in `fit()` | Raise `ValueError` — degenerate labels |
| `y_true` empty or contains NaN in `fit()` | Raise `ValueError` |
| `sample()` called before `fit()` | Raise `ValueError` |
| `y_proxy_cost <= 0` or `y_true_cost <= 0` | Raise `ValueError` |
| `budget <= 0` | Raise `ValueError` |
| `uncertainties` contains NaN | Raise `ValueError` |
| `uncertainties` contains negative values | Raise `ValueError` — a conditional MSE cannot be negative |
| All `u_i == 0` | Raise `ValueError` — proxy is perfect everywhere, no benefit from H |
| Budget too small for T=1 | Raise `ValueError` |

---

## File Structure

```
glide/
  samplers/
    __init__.py          ← add CostOptimalSampler to exports
    active.py            (unchanged)
    stratified.py        (unchanged)
    cost_optimal_random.py   (unchanged, already implemented)
    cost_optimal.py      ← NEW
tests/unit/
  samplers/
    __init__.py          (unchanged)
    test_active.py       (unchanged)
    test_stratified.py   (unchanged)
    test_cost_optimal_random.py  (unchanged, already implemented)
    test_cost_optimal.py  ← NEW
```

## Code Conventions

Follow [glide/samplers/active.py](glide/samplers/active.py) as the canonical reference:
- NumPy-style docstrings with `Parameters`, `Returns`, `Raises`, `References`, `Examples` sections.
- A runnable doctest in the `Examples` section of the class docstring and of `sample()`.
- No docstrings on private methods (`_compute_gamma`, `_compute_per_record_probabilities`, etc.).
- Type annotations using `typing` module — **not** PEP 604/585 built-ins:
  - Use `Tuple[NDArray, NDArray, NDArray]`, not `tuple[...]`.
- Never compute in a `return` statement — always assign to a named variable first.
- `from typing import Tuple`
- `import math` (for `math.floor`)

# Functional test
This sampler should coincide with the CostOptimalRandomSampler when uncertainty function yields a constant

imerad:
### Description

- What does this PR do?
See title
Deals with this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=173583418&issue=EmertonData%7Cglide%7C184) and Closes #184 
Multiple changes to original ticket : 
  - Zero uncertainties have been forbidden to preserve statistical validity and align with ActiveSampler
  - Budget allocation take a deterministic number of the first samples rather than drawing at random based on average cost. This reduce fluctuation and deviation from the prescribed budget
  - `sample()` output format changed to `xi, pi` aligning with other IPW samplers
  - the main package folder `glide` was added to UNIT_DIRS in the makefile to make sure unit tests also ran the doctests
  - Added an instruction claude.md to stop generating superfluous comments in unit tests

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
### Code review

Found 3 issues:

1. **`_compute_gamma` computes a partial sum instead of a conditional expectation** (bug in core algorithm)

`np.mean(uncertainties * ~above_mask)` divides by the total sample count `n`, producing `E[u · I(u ≤ τ²)]` rather than the required conditional expectation `E[u | u ≤ τ²]`. This underestimates `e_u_below`, inflates `denominator`, and causes suboptimal (or degenerate) sampling probabilities for any heterogeneous uncertainty distribution. Fix: `np.mean(uncertainties[~above_mask])`.

https://github.com/EmertonData/glide/blob/a5943e92d93793e9a72bc9dc15ae37881d91b693/glide/samplers/cost_optimal.py#L103-L108

2. **`np.searchsorted` uses `side="left"`, causing an off-by-one at exact budget boundaries** (bug)

When `budget` exactly equals `cumulative_costs[k]`, `side="left"` returns `k` instead of `k+1`, so the last affordable sample is excluded and the subsequent `T < 1` guard may raise a spurious `ValueError` even when one sample is affordable. The sibling `CostOptimalRandomSampler` uses `int(np.floor(budget / cost_per_sample))` which has `side="right"` semantics. Fix: `side="right"`.

https://github.com/EmertonData/glide/blob/a5943e92d93793e9a72bc9dc15ae37881d91b693/glide/samplers/cost_optimal.py#L213-L218

3. **Unit test file contains section-separator comments, violating the rule this PR introduces** ([`.claude/CLAUDE.md`](https://github.com/EmertonData/glide/blob/a5943e92d93793e9a72bc9dc15ae37881d91b693/.claude/CLAUDE.md) says "No comments in unit tests. Encode non-obvious derivations as named variables in the test body instead.")

The PR adds this rule to CLAUDE.md but the new test file has six `# --- <method> ---` separator comments.

https://github.com/EmertonData/glide/blob/a5943e92d93793e9a72bc9dc15ae37881d91b693/tests/unit/samplers/test_cost_optimal.py#L33-L37

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+
+    np.testing.assert_allclose(pi, expected_pi, atol=0.01)
+    assert xi[0] == 0.0
+    assert np.isnan(xi[1])
+
+
+def test_sample_reproducibility(fitted_sampler, uncertainties):
+    pi1, xi1 = fitted_sampler.sample(uncertainties, y_true_cost=1.0, y_proxy_cost=0.01, budget=1.5, random_seed=42)
+    pi2, xi2 = fitted_sampler.sample(uncertainties, y_true_cost=1.0, y_proxy_cost=0.01, budget=1.5, random_seed=42)
+
```
mraki-ed, line 175: le budget est défini en tant que entier, ce serait préférable d'utiliser un entier ici aussi

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,230 @@
...
+    def _find_optimal_threshold(
+        self,
+        uncertainties: NDArray,
+        y_true_cost: float,
+        y_proxy_cost: float,
+    ) -> float:
+        sqrt_u_values = np.sqrt(uncertainties)
+        candidates = np.unique(sqrt_u_values)
+        # the breakpoints where the policy changes character
+        # are the distinct values of sqrt(u_i).
```
mraki-ed: si on suit la policy de claude.md sur les commentaires, je pense que ce commentaire n'est pas nécessaire

mraki-ed (changes requested):

tests/unit/samplers/test_cost_optimal.py
imerad: Le budget ne devrait pas être contraint à être entier, je l'ai changé en float comme dans CostOptimalRandomSampler

glide/samplers/cost_optimal.py
imerad: Enlevé

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,228 @@
...
+class CostOptimalSampler:
+    """Sampler that draws elements with optimal probabilities based on uncertainty
+    scores and annotation costs on a limited budget.
+
+    Implements a cost-optimal active annotation policy. Each sample is assigned
+    an annotation probability proportional to how unreliable the proxy label is
+    expected to be for that sample, as measured by the caller-supplied per-sample
+    uncertainty scores. Samples with high expected proxy error are more likely to
+    be annotated; samples where the proxy is reliable are less likely to be
+    annotated. This concentrates the annotation budget where it matters most.
```
gmartinon-ed, line 17: The sentence "Samples..." is not symetric, thus confusing

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,228 @@
...
+            Pass ``None`` (the default) to use a non-deterministic seed.
+
+        Returns
+        -------
+        Tuple[NDArray, NDArray]
+            [0]: array of shape ``(n_samples,)``, ``pi`` with per-sample annotation probabilities
+            for selected samples and ``0.0`` for unselected samples.
+            [1]: array of shape ``(n_samples,)``, ``xi`` with Bernoulli indicators:
+            ``1.0`` if the true label was requested, ``0.0`` if only the proxy
+            was used, ``NaN`` if the sample was not selected.
```
gmartinon-ed, line 191: C'est pas -1 ? Dans le doctest c'est -1.
Je pense qu'il faut ajouter une section upfront, dans la docstring de la classe, qui explicite les sorties 0, 1, -1 et leur sens en termes d'exploitation par l'utilisateur (annotation manuelle, proxy, ou ignore). A propager dans les autres samplers

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def _validate_y_true(self, y_true: NDArray) -> None:
+        if len(y_true) == 0:
+            raise ValueError("'y_true' must be non-empty.")
+        if np.any(np.isnan(y_true)):
+            raise ValueError("'y_true' contains NaN values. The burn-in dataset must be fully labeled.")
+        if np.min(y_true) == np.max(y_true):
+            raise ValueError("'y_true' label values have zero variance.")
+
+    def _validate_uncertainties(self, uncertainties: NDArray) -> None:
```
gmartinon-ed, line 59: Bien d'anticiper les prochaines refacto. J'aurais limite envie de demander de propager maintenant, mais c'est parce que je suis impatient :-) Restons focus sur cette PR qui est déjà grosse

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,228 @@
...
+        tau: float,
+        uncertainties: NDArray,
+        y_true_cost: float,
+        y_proxy_cost: float,
+    ) -> float:
+        gamma = self._compute_gamma(tau, uncertainties, y_true_cost, y_proxy_cost)
+        pi_values = self._compute_per_sample_probabilities(tau, gamma, uncertainties)
+        mean_pi = np.mean(pi_values)
+        cost_term = y_true_cost * mean_pi + y_proxy_cost
+        error_term = self._y_true_variance + np.mean(uncertainties * (1.0 / pi_values - 1.0))
```
gmartinon-ed, line 132: Désolé je fais du cross-PR. Ce n'est pas tout à fait clair pour moi en quoi le `error_term` est bien un terme d'erreur. Possible de clarifie pour moi, et pour les utilisateurs dans le User Guide ?

glide/samplers/cost_optimal.py
```diff
@@ -0,0 +1,228 @@
...
+        if budget <= 0:
+            raise ValueError(f"'budget' must be strictly positive; got {budget}.")
+        self._validate_uncertainties(uncertainties)
+
+        tau_star = self._find_optimal_threshold(uncertainties, y_true_cost, y_proxy_cost)
+        gamma_star = self._compute_gamma(tau_star, uncertainties, y_true_cost, y_proxy_cost)
+        pi_all = self._compute_per_sample_probabilities(tau_star, gamma_star, uncertainties)
+
+        cumulative_costs = np.cumsum(y_true_cost * pi_all + y_proxy_cost)
+        T = np.searchsorted(cumulative_costs, budget, side="right")
```
gmartinon-ed: Ok pour ce changement, c'est un arbitraire moins fluctuant que le tirage au sort. Tu penses qu'il faut propager à CostOptimalRandom ? Par ailleurs, ça fait implicitement l'hypothèse que le dataset est trié par ordre chronologique. Possible de le préciser dans la docstring, que le cut de budget suit l'ordre des lignes du budget ?

tests/functional/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,38 @@
...
+    pi_cost_optimal, xi_cost_optimal = cost_optimal_sampler.sample(
+        uniform_uncertainties,
+        y_true_cost=y_true_cost,
+        y_proxy_cost=y_proxy_cost,
+        budget=budget,
+        random_seed=random_seed,
+    )
+
+    np.testing.assert_allclose(pi_cost_optimal, pi_random, atol=0.001)
+    np.testing.assert_array_equal(xi_cost_optimal, xi_random)
```
gmartinon-ed, line 36: C'est marrant que ça tienne ça, car dans l'un on tire au hasard les T observations, dans l'autre on prend les T premières. Ah mais tu dis que tu prends tout. En fait si on homogénéise entre les deux le fait de prendre les T premières, on peut avoir un test plus robuste avec un budget plus modeste ?

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+    with pytest.raises(ValueError, match="NaN"):
+        sampler.fit(np.array([1.0, np.nan]))
+
+
+def test_fit_raises_on_zero_variance_y_true(sampler):
+    with pytest.raises(ValueError, match="zero variance"):
+        sampler.fit(np.array([1.0, 1.0]))
+
+
+def test_fit_known_variance(sampler):
```
gmartinon-ed, line 53: Il manque pas un test similaire pour CostOptimalRandom ?

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def test_compute_gamma_known_output(fitted_sampler, uncertainties):
+    tau = np.sqrt(0.4)
+    gamma = fitted_sampler._compute_gamma(tau, uncertainties, y_true_cost=10.0, y_proxy_cost=1.0)
+    expected_gamma = 0.6325
+    assert gamma == pytest.approx(expected_gamma, abs=0.001)
+
+
+def test_compute_gamma_known_output_small_variance(fitted_sampler_small_variance, uncertainties):
```
gmartinon-ed, line 69: Pourquoi deux tests ici ?

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+    with pytest.raises(ValueError, match="non-positive value"):
+        fitted_sampler.sample(np.array([0.1, 0.0]), y_true_cost=10.0, y_proxy_cost=1.0, budget=5, random_seed=42)
+
+
+def test_sample_budget_too_small_raises(fitted_sampler, uncertainties):
+    with pytest.raises(ValueError, match="Budget .* is too small"):
+        fitted_sampler.sample(uncertainties, y_true_cost=10.0, y_proxy_cost=1.0, budget=1, random_seed=42)
+
+
+def test_sample_known_output(fitted_sampler, uncertainties):
```
gmartinon-ed, line 152: Homogénéiser le naming de ces tests à travers les samplers (known VS valid)

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+    with pytest.raises(ValueError, match="'budget' must be strictly positive"):
+        fitted_sampler.sample(uncertainties, y_true_cost=10.0, y_proxy_cost=1.0, budget=budget, random_seed=42)
+
+
+def test_sample_raises_on_nan_uncertainties(fitted_sampler):
+    with pytest.raises(ValueError, match="NaN"):
+        fitted_sampler.sample(np.array([0.1, np.nan]), y_true_cost=10.0, y_proxy_cost=1.0, budget=5, random_seed=42)
+
+
+def test_sample_raises_on_non_positive_uncertainties(fitted_sampler):
```
gmartinon-ed, line 147: possible de mettre ces tests spécifiques après le test du budget qui est générique aux cost optimal samplers ?

tests/unit/samplers/test_cost_optimal.py
```diff
@@ -0,0 +1,173 @@
...
+    pi, xi = fitted_sampler.sample(uncertainties, y_true_cost=10.0, y_proxy_cost=1.0, budget=10, random_seed=42)
+
+    expected_pi = np.array([0.2, 0.4])
+    expected_xi = np.array([0.0, 0.0])
+
+    np.testing.assert_allclose(pi, expected_pi, atol=0.01)
+    np.testing.assert_array_equal(xi, expected_xi)
+
+
+def test_sample_known_output_truncated_samples(fitted_sampler, uncertainties):
```
gmartinon-ed, line 162: Homogénéiser le naming avec le test similaire dans CostOptimalRandom (test_sample_n_affordable_less_than_n_samples)

gmartinon-ed (changes requested):

glide/samplers/cost_optimal.py
imerad: Traité

glide/samplers/cost_optimal.py
imerad: En fait, c'était `xi[-1]` qui était NaN dans la doctest, c'est pour montrer que le dernier élément n'a pas été sélectionné et a reçu NaN. On a mis des NaNs dans CostOptimalRandomSampler et c'est ce qui est attendu dans `simulate_annotation` aussi.

J'ai rajouté des sections pour expliquer la signification et l'usage des sorties pi et xi dans tous les samplers concernés. J'ai préféré les mettre dans la docstring de la méthode sample car la docstring de la classe contient une description très haut niveau. Ça crée une redondance avec la section "Returns" de la docstring de `sample` mais je crois que ça rend les choses plus claires

glide/samplers/cost_optimal.py
imerad: Ce serait bien de faire ça en même temps que la centralisation de ces tests de validation d'inputs

tests/unit/samplers/test_cost_optimal.py
imerad: Ajouté

tests/unit/samplers/test_cost_optimal.py
imerad: Le deuxième permet de couvrir le `else` dans `_compute_gamma()`

tests/unit/samplers/test_cost_optimal.py
imerad: Fait

tests/unit/samplers/test_cost_optimal.py
imerad: Je les ai mis après `test_sample_budget_too_small_raises`, c'est ce que tu voulais dire ?

tests/unit/samplers/test_cost_optimal.py
imerad: Fait

glide/samplers/cost_optimal.py
imerad: J'ai préféré faire comme ça car c'est beaucoup plus facile de s'adapter au budget alors que le tirage aléatoire risquait de donner des éléments dont la somme des probabilités n'y correspondait pas vraiment.

J'ai d'ores et déjà propagé à CostOptimalRandom et fait les adaptations des tests, du user guide et du tutoriel. 

Ce choix ne suppose pas vraiment que le dataset soit trié par ordre chronologique, il faut simplement l'ordre donné ne biaise pas la distribution cible. La docstring précise que les premiers éléments sont pris au nombre permis par le budget

tests/functional/samplers/test_cost_optimal.py
imerad: Oui, c'était parce que le budget était très grand donc pas de cutoff. Je viens de propager le changement au CostOptimalRandomSampler ce qui a permis de mettre un budget contraignant ici et rendre le test plus fort

imerad:
Une observation importante. Je pense que les incertitudes requises par ce sampler seront très souvent des scores d'incertitude de LLM-as-judge sur des labels proxy qu'il aura déjà donné ... Ça veut dire que tous les labels proxy seront déjà disponibles et ça n'a pas de sens de tenir compte de leur coût dans les calculs de ce sampler.

Je propose de permettre y_proxy_cost = 0 dans ce Sampler pour gérer ça. Ça correspondra au fait qu'ils ont déjà été obtenus et on veut allouer le budget restant aux annotations humaines seulement. On pourra aussi parler de ce scénario dans le tutoriel

glide/samplers/cost_optimal.py
imerad: Je regarde ça dans la PR user guide

gmartinon-ed (approved):

mraki-ed (approved):

PR #232, opened 2026-05-29
Add constrained optimization to ActiveSampler

### LINKED ISSUE #213: ActiveSampler: replace probability clipping with constrained optimisation
# ActiveSampler: replace probability clipping with constrained optimisation

## Background

`ActiveSampler` computes raw probabilities `π_i = n · u_i / Σ u_j`, then clips any value above 1 to 1. This wastes budget: the expected number of annotated items becomes `Σ min(π_i, 1)`, which is strictly less than `n` whenever any probability saturates.

The correct fix is to solve the full optimisation problem: minimise `Σ u_i² / π_i` subject to `Σ π_i = n` and `π_i ≤ 1`. This yields probabilities that sum to exactly `n` and respect the Bernoulli constraint. Notably, the unconstrained optimum of `Σ u_i² / π_i` is `π_i ∝ u_i` — exactly the current formula — so the fast path (no saturation) is unchanged. The current clipping discards the saved budget instead of redistributing it optimally.

## Design choices

`_compute_probabilities` is added as a private method on `ActiveSampler`. When `max(n · u / Σu) ≤ 1`, no saturation occurs and the method returns the current closed-form formula unchanged. When at least one probability would exceed 1, it falls back to `scipy.optimize.minimize` with `method='trust-constr'`, passing the analytic Jacobian for reliable convergence. The warm start is `n · u / Σu` clipped to 1 (the unconstrained optimum, already a feasible point). `maxiter=100` bounds runtime on large inputs.

`sample()` is simplified to a single call to `_compute_probabilities`, replacing the current `np.minimum` clip. Its docstring is updated to state that the probabilities sum to `budget` exactly.

### `_compute_probabilities` (new private method on `ActiveSampler`)

**Location:** [glide/samplers/active.py](glide/samplers/active.py)

**Purpose:** Return drawing probabilities that sum to `budget` and lie in `(0, 1]`, falling back to constrained optimisation when the closed-form result would saturate.

Add `from scipy.optimize import Bounds, LinearConstraint, minimize` to the top-level imports.

```python
def _compute_probabilities(self, uncertainties: NDArray, budget: int) -> NDArray:
    raw_pi = budget * uncertainties / uncertainties.sum()
    if np.max(raw_pi) <= 1.0:
        return raw_pi

    n = len(uncertainties)
    warm_start = np.minimum(raw_pi, 1.0)
    squared_uncertainties = np.power(uncertainties, 2)

    def objective(pi: NDArray) -> float:
        result = np.sum(squared_uncertainties / pi)
        return result

    def jacobian(pi: NDArray) -> NDArray:
        gradient = -squared_uncertainties / np.power(pi, 2)
        return gradient

    bounds = Bounds(ub=np.ones(n))
    budget_constraint = LinearConstraint(np.ones((1, n)), lb=budget, ub=budget)
    optimization_result = minimize(
        objective,
        warm_start,
        method="trust-constr",
        jac=jacobian,
        constraints=[budget_constraint],
        bounds=bounds,
        options={"maxiter": 100},
    )
    return optimization_result.x
```

### Updated `sample()` method

**Location:** [glide/samplers/active.py](glide/samplers/active.py)

Replace the two lines `pi = budget * uncertainties / uncertainties.sum()` and `pi = np.minimum(pi, 1.0)` with a single call:

```python
pi = self._compute_probabilities(uncertainties, budget)
```

Update the docstring description and the `Returns` entry to say probabilities sum to `budget` (dropping "at most"):

```
Each observation receives a drawing probability π_i proportional to
``uncertainty_i``, normalised so that the probabilities sum to ``budget``
exactly. Probabilities are constrained to ``(0, 1]``; when the closed-form
allocation would exceed 1 for any item, the constrained optimum is found
via numerical optimisation.
```

```
[0]: array of shape ``(n_samples,)``, drawing probabilities in ``(0, 1]``
     that sum to ``budget``.
```

## Tests to update

**Location:** [tests/unit/samplers/test_active.py](tests/unit/samplers/test_active.py)

`test_sample_pi_clipped_and_higher_uncertainty_gets_higher_pi` — with `[0.001, 10.0]`

## Tests to add

**Location:** [tests/unit/samplers/test_active.py](tests/unit/samplers/test_active.py)

```python
def test_compute_probabilities_no_saturation(sampler):
    pi = sampler._compute_probabilities(np.array([0.1, 0.4]), budget=1)
    np.testing.assert_allclose(pi, np.array([0.2, 0.8]), atol=1e-10)


def test_compute_probabilities_with_saturation(sampler):
    # uncertainties=[0.01, 1.0, 9.0], budget=2: raw pi_3=1.798 saturates.
    # Optimising 0.01^2/pi_1 + 1.0^2/pi_2 subject to pi_1+pi_2=1 gives pi_i proportional to u_i:
    # pi_1 = 0.01/1.01 = 1/101, pi_2 = 1.0/1.01 = 100/101.
    pi = sampler._compute_probabilities(np.array([0.01, 1.0, 9.0]), budget=2)
    np.testing.assert_allclose(pi, np.array([1 / 101, 100 / 101, 1.0]), atol=1e-6)
```

## Corner cases

- **No saturation**: `_compute_probabilities` returns the closed-form result without calling scipy.
- **All saturate** (`budget == len(uncertainties)`, uniform uncertainties): raw pi is all ones, the condition `max ≤ 1` is true, scipy is never called.
- **Cascading saturation**: the trust-constr solver handles this natively.

## Acceptance criteria

- [ ] `ActiveSampler._compute_probabilities(uncertainties, budget)` exists as a private method, returns an array summing to `budget` within `1e-6`, with all values in `(0, 1]`.
- [ ] When no raw probability exceeds 1, `_compute_probabilities` returns `budget * u / sum(u)` without calling the scipy optimiser.
- [ ] `test_compute_probabilities_with_saturation` passes with `pi ≈ [1/101, 100/101, 1.0]` for `[0.01, 1.0, 9.0]` at `budget=2`.
- [ ] `test_sample_pi_clipped_and_higher_uncertainty_gets_higher_pi` is updated to assert `pi == [1.0, 1.0]`.

imerad:
### Description

- What does this PR do?
See title
Settles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=191745084&issue=EmertonData%7Cglide%7C213) Closes #213 

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
## Code Review

Overall the approach is sound and the analytical reasoning in the issue is correct (verified the KKT conditions for the saturation test case). Five issues below, ranked by severity.

---

### 1. No convergence check — silent contract violation (High)

**File:** `glide/samplers/active.py`, line 70

`optimization_result.success` is never checked before the result is used. If `trust-constr` exhausts `maxiter=100` without converging, `optimization_result.x` is returned anyway and may not satisfy `Σπ_i = budget`. The `sample()` docstring promises probabilities that sum to `budget` exactly, but this guarantee silently fails on non-convergence.

```python
# current
return np.minimum(optimization_result.x, 1.0)

# suggested
if not optimization_result.success:
    raise RuntimeError(
        f"Constrained probability optimisation did not converge: {optimization_result.message}"
    )
return np.minimum(optimization_result.x, 1.0)
```

---

### 2. Lower bound of 0 allows pi = 0 (Confirmed)

**File:** `glide/samplers/active.py`, line 59

`Bounds(lb=np.zeros(n), ub=np.ones(n))` permits the optimizer to return `π_i = 0`. The docstring guarantees probabilities in `(0, 1]` (strictly positive). While the objective `Σu_i²/π_i` naturally repels from 0 at the optimum, a non-converged result (see finding #1) can still return zero or near-zero values.

Downstream in `ipw_ptd.py` line 167, `labeled_ipw_weights = xi / pi` is computed for **all** items before the `non_zero_pi_mask` is applied at line 170. A `pi_i = 0` produces `NaN` (`0/0`) or `inf` (`1/0`), which are then silently discarded. This masks the upstream contract violation rather than surfacing it.

Use a small positive lower bound:

```python
bounds = Bounds(lb=np.full(n, 1e-10), ub=np.ones(n))
```

---

### 3. Infeasible warm start wastes convergence budget (Medium)

**File:** `glide/samplers/active.py`, line 48

`warm_start = np.minimum(raw_pi, 1.0)` sums to strictly less than `budget` whenever saturation occurs, so the starting point violates the equality constraint. `trust-constr` handles this by running a feasibility-restoration phase that can consume 30–50 of the 100 allowed iterations before optimisation even begins — directly amplifying the non-convergence risk in finding #1.

A feasible warm start redistributes the clipped surplus to unsaturated items:

```python
clipped = np.minimum(raw_pi, 1.0)
deficit = budget - clipped.sum()
unsaturated_mask = raw_pi < 1.0
clipped[unsaturated_mask] += deficit / unsaturated_mask.sum()
warm_start = clipped
```

---

### 4. Redundant post-optimization clip (Low)

**File:** `glide/samplers/active.py`, line 70

`np.minimum(optimization_result.x, 1.0)` is dead code under normal convergence — `trust-constr` respects the `ub=np.ones(n)` bound and never returns values above 1. But if it ever did activate (e.g. due to a future scipy tolerance change), it would silently reduce `Σπ_i` below `budget`, breaking the very contract this method is meant to enforce. Better to remove the clip and add an assertion instead:

```python
result = optimization_result.x
assert np.all(result <= 1.0 + 1e-9), "Optimizer violated upper bound"
return result
```

---

### 5. Test name / invocation mismatch (Low)

**File:** `tests/unit/samplers/test_active.py`, line 56

`test_compute_probabilities_no_saturation` calls `sampler.sample(...)` rather than `sampler._compute_probabilities(...)` directly, while `test_compute_probabilities_with_saturation` (line 61) correctly calls the private method. The name implies it exercises `_compute_probabilities` in isolation. Consider renaming to `test_sample_no_saturation` to match what is actually being tested, or call `_compute_probabilities` directly for symmetry with the saturation test.

---

### Checklist items

Two unchecked items in the PR description deserve attention:

- **`CHANGELOG.md` not updated** — the behavioral change (pi now sums to `budget` exactly instead of at most `budget`) is user-facing and should have an entry under `Changed`.
- **"New public API has numpy-style docstrings"** — `_compute_probabilities` is private so no docstring is needed (correct per CLAUDE.md). The `sample()` docstring has been updated. This checklist item can be checked off.

glide/samplers/active.py
```diff
@@ -52,12 +91,11 @@ def sample(
...
         """Sample observations with probability proportional to uncertainty.
 
-        Each observation receives a drawing probability π_i proportional to
-        ``uncertainty_i``, normalised so that the raw probabilities sum to
-        ``budget`` (the expected number of selected observations). Because each
-        π_i must be a valid Bernoulli probability, values are capped at 1 before
-        the coin flip; the actual number of selected items is therefore a random
-        variable whose expectation equals at most ``budget``.
+        Each observation receives a drawing probability π_i that optimizes downstream
+        estimation variance by minimizing the sum of ``uncertainty_i / π_i`` over all
```
gmartinon-ed: c'est pas le carré des incertitudes ici qui est minimisé ? Globalement, si on ne connaît pas la formule de la variance de l'estimateur, on ne sait pas pourquoi. Préciser qu'on minimise la variance de l'estimateur propensity weighted associé, et mettre en référence le papier de Candès

gmartinon-ed (changes requested):
Pas d'impact sur les samplers cost optimal ?
Et quel impact sur les tuto et les deep dives ?

glide/samplers/active.py
imerad: Traité

imerad:
Les samplers cost optimal ne sont pas impactés car ils résolvent des problèmes d'optimisation différents.

Quelques notebooks de deep dive et tutoriels sont impactés par des changements des explications sur le fonctionnement d'`ActiveSampler` mais les résultats expérimentaux restent bons

gmartinon-ed (approved):

PR #244, opened 2026-06-03
Add `generate_clustered_binary_data` simulator

### LINKED ISSUE #234: Simulate clustered binary data
# Simulate clustered binary data

## Background

The cluster estimators `ClusterClassicalMeanEstimator`, `ClusterPTDMeanEstimator`, and `ClusterPPIMeanEstimator` all require observations grouped into clusters, where each cluster is either entirely labeled or entirely unlabeled. GLIDE currently has no simulator that produces this structure: `generate_binary_dataset` generates flat arrays with no cluster membership, and `generate_stratified_binary_dataset` groups observations by stratum but does not model the cluster sampling pattern that the cluster estimators are designed for.

A natural real-world analogy is a document evaluation pipeline: a human reviewer either reads an entire paragraph and labels every sentence in it, or reads nothing and all sentences receive only proxy labels. Each paragraph is a cluster. Paragraph lengths vary: some are short (two or three sentences) and some are long (ten or more). This irregular cluster-size structure is exactly what makes cluster-based confidence intervals necessary, and it is the structure this simulator should reproduce.

Since within-cluster observations are i.i.d. in this model (no latent per-cluster random effect), drawing all `n_total` observations at once from `generate_binary_dataset` and then randomly assigning cluster labels is statistically equivalent to drawing per-cluster. The simulator therefore delegates the data generation entirely to `generate_binary_dataset`, then produces cluster membership by placing `n_clusters - 1` random cut points among the `n_total` observation indices. This guarantees exactly `n_total` observations and exactly `n_clusters` non-empty clusters, with sizes that vary randomly around `n_total / n_clusters`. The within-cluster i.i.d. assumption is intentionally simple: the cluster structure (not intra-cluster correlation) is what the estimators need, and i.i.d. within clusters is sufficient for correctness tests and validation notebooks.

## Design choices

`generate_clustered_binary_dataset` lives in `glide/simulators/clustered_binary.py` and mirrors `generate_binary_dataset` (`binary.py`) and `generate_stratified_binary_dataset` (`stratified_binary.py`) in structure, naming, and parameter order. It delegates the data draw entirely to `generate_binary_dataset`, which keeps all joint distribution logic in one place and avoids any per-cluster loop. The only new logic is the random partition into clusters.

The function returns `(y_true_oracle, y_proxy, clusters)` with no NaN. Cluster identifiers are contiguous integers 0, 1, ..., `n_clusters - 1`, consistent with the integer group identifiers in `generate_stratified_binary_dataset`. The `clusters` array has dtype `int64`. Both `y_true_oracle` and `y_proxy` contain only 0.0 and 1.0 (no NaN), meaning they are fully oracle arrays ready to be passed to `simulate_annotation`.

Two validation constraints are enforced: `n_clusters >= 2` (fewer than two clusters is also the minimum required by the cluster estimators) and `n_total >= n_clusters` (the partition requires at least one observation per cluster, so there must be at least as many observations as clusters). The remaining parameter checks (`true_mean`, `proxy_mean`, `correlation`) reuse the same conditions as `generate_binary_dataset`.

### `generate_clustered_binary_dataset`

**Location:** `glide/simulators/clustered_binary.py`

**Purpose:** Generate a synthetic binary-label dataset with randomly sized clusters, returning fully populated oracle arrays and cluster identifiers.

```python
from typing import Optional, Tuple, Union

import numpy as np


def generate_clustered_binary_dataset(
    n_total: int,
    n_clusters: int,
    true_mean: float = 0.7,
    proxy_mean: float = 0.6,
    correlation: float = 0.8,
    random_seed: Optional[Union[int, np.random.SeedSequence]] = None,
    """Generate a synthetic clustered binary-label dataset for evaluation.

    Draws ``n_total`` i.i.d. ``(y_true_oracle, y_proxy)`` pairs from the
    joint binary distribution defined by ``true_mean``, ``proxy_mean``, and
    ``correlation``, then randomly partitions the observations into
    ``n_clusters`` non-empty groups. Returns oracle arrays with no NaN
    masking.

    Parameters
    ----------
    n_total : int
        Exact total number of observations across all clusters.
    n_clusters : int
        Exact number of clusters. Must be at least 2.
    true_mean : float
        Expected mean value of the true labels. Must be in ``(0, 1)``.
    proxy_mean : float
        Expected mean value of the proxy labels. Must be in ``(0, 1)``.
    correlation : float
        Pearson correlation between true and proxy labels.
    random_seed : int or np.random.SeedSequence, optional
        Seed for reproducibility.

    Returns
    -------
    Tuple[NDArray, NDArray, NDArray]
        [0]: ``y_true_oracle`` — shape ``(n_total,)``, all values in
             ``{0.0, 1.0}``, no NaN.
        [1]: ``y_proxy`` — shape ``(n_total,)``, all values in
             ``{0.0, 1.0}``, no NaN.
        [2]: ``clusters`` — shape ``(n_total,)``, integer cluster
             identifiers in ``{0, 1, ..., n_clusters - 1}``.

    Raises
    ------
    ValueError
        If ``true_mean`` is not in ``(0, 1)``.
    ValueError
        If ``proxy_mean`` is not in ``(0, 1)``.
    ValueError
        If the combination of ``true_mean``, ``proxy_mean``, and
        ``correlation`` leads to negative joint probabilities.
    ValueError
        If ``n_clusters < 2``.
    ValueError
        If ``n_total < n_clusters``.

    Notes
    -----
    **Step 1 — Draw observations**

    Call ``generate_binary_dataset(n_total, ...)`` to obtain ``n_total``
    i.i.d. ``(y_true_oracle, y_proxy)`` pairs from the joint binary
    distribution defined by ``true_mean``, ``proxy_mean``, and
    ``correlation``.

    **Step 2 — Random cluster partition**

    Draw ``n_clusters - 1`` cut positions uniformly without replacement from
    ``{1, 2, ..., n_total - 1}`` and sort them. Combined with ``0`` and
    ``n_total``, these define ``n_clusters`` contiguous intervals of random
    lengths that sum to ``n_total``. Assign cluster identifier ``k`` to all
    observations whose position falls in the ``k``-th interval. Every cluster contains at least 1 observation by construction.

    **Step 3 — Shuffle**

    Randomly permute the cluster identifier array so that cluster membership is not determined by position in the output.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.simulators import generate_clustered_binary_dataset
    >>> y_true_oracle, y_proxy, clusters = generate_clustered_binary_dataset(
    ...     n_total=20, n_clusters=4, random_seed=0
    ... )
    >>> len(np.unique(clusters))
    4
    >>> bool(np.all(~np.isnan(y_true_oracle)))
    True
    >>> bool(np.all(~np.isnan(y_proxy)))
    True
    """
    if n_clusters < 2:
        raise ValueError(f"'n_clusters' must be >= 2; got {n_clusters}.")
    if n_total < n_clusters:
        raise ValueError(
            f"'n_total' must be >= 'n_clusters'; got n_total={n_total} and n_clusters={n_clusters}."
        )

    seed_sequence = np.random.SeedSequence(random_seed)
    data_seed, partition_seed = seed_sequence.spawn(2)

    y_true_oracle, y_proxy = generate_binary_dataset(
        n_total=n_total,
        true_mean=true_mean,
        proxy_mean=proxy_mean,
        correlation=correlation,
        random_seed=data_seed,
    )

    rng = np.random.default_rng(partition_seed)
    cut_positions = np.sort(rng.choice(np.arange(1, n_total), size=n_clusters - 1, replace=False))
    interval_lengths = np.diff(np.hstack([[0], cut_positions, [n_total]]))
    clusters = np.repeat(np.arange(n_clusters, dtype=np.int64), interval_lengths)
    rng.shuffle(clusters)

    return y_true_oracle, y_proxy, clusters
```

### Updates to `glide/simulators/__init__.py`

Add the following import (keep alphabetical order within the group):

```python
from glide.simulators.clustered_binary import generate_clustered_binary_dataset
```

Add `"generate_clustered_binary_dataset"` to `__all__`.

## Functional tests

**Empirical means and correlation converge to the specified parameters**

With a large sample, the empirical `true_mean`, `proxy_mean`, and labeled correlation should fall within a tight tolerance of the requested parameters, just as the functional test for `generate_binary_dataset` does.

```python
def test_generate_clustered_binary_dataset_empirical_means_and_correlation():
    y_true_oracle, y_proxy, clusters = generate_clustered_binary_dataset(
        n_total=5000, n_clusters=50, true_mean=0.7, proxy_mean=0.6, correlation=0.8, random_seed=42
    )
    assert np.mean(y_true_oracle) == pytest.approx(0.7, abs=0.03)
    assert np.mean(y_proxy) == pytest.approx(0.6, abs=0.03)
    assert np.corrcoef(y_true_oracle, y_proxy)[0, 1] == pytest.approx(0.8, abs=0.05)
```

Note: these tests live in `tests/functional/simulators/`.

## Corner cases

- `n_clusters < 2` must raise `ValueError` with the message `"'n_clusters' must be >= 2; got {n_clusters}."`.
- `n_total < n_clusters` must raise `ValueError` with the message `"'n_total' must be >= 'n_clusters'; got n_total={n_total} and n_clusters={n_clusters}."`.
- `true_mean` outside `(0, 1)` and `proxy_mean` outside `(0, 1)` must raise `ValueError` with the same message format as `generate_binary_dataset`.
- An impossible correlation (leading to negative joint probabilities) must raise `ValueError` with the same message format as `generate_binary_dataset`.
- Two calls with the same `random_seed` must return identical arrays; two calls with `random_seed=None` must produce different arrays in practice.
- The actual total is always exactly `n_total` and every cluster always has at least one observation, because the partition is derived from `n_clusters - 1` cut points among `{1, ..., n_total - 1}`.

## Propagation to existing simulator tests

All four existing simulator test files (`test_binary.py`, `test_stratified_binary.py`, `test_oracle_binary.py`, `test_gaussian.py`) cover the same-seed half of reproducibility but not the `None`-seed half. Add the following test to each, adapted to the function's signature:

```python
def test_generate_binary_dataset_none_seed_is_nondeterministic():
    y_true1, y_proxy1 = generate_binary_dataset(n_total=10, seed=1)
    y_true2, y_proxy2 = generate_binary_dataset(n_total=10, seed=2)
    assert not np.array_equal(y_true1, y_true2)
    assert not np.array_equal(y_proxy1, y_proxy2)
```

The pattern is the same for every simulator: call twice with no seed, assert the outputs differ on at least one array. `simulate_annotation` is deterministic given its inputs and needs no such test.

## Acceptance criteria

- [ ] `generate_clustered_binary_dataset` is importable from `glide.simulators` and listed in `__all__`.
- [ ] The function raises `ValueError` for `n_clusters < 2`, `n_total < n_clusters`, invalid `true_mean`, invalid `proxy_mean`, and impossible `correlation` combinations, with messages matching the formats above.
- [ ] The returned arrays satisfy: `len(y_true_oracle) == len(y_proxy) == len(clusters)`, no NaN in any array, all `y_true_oracle` and `y_proxy` values in `{0.0, 1.0}`, and exactly `n_clusters` distinct values in `clusters`.
- [ ] Every cluster has at least 1 observation.
- [ ] The functional test `test_generate_clustered_binary_dataset_empirical_means_and_correlation` passes: empirical `true_mean`, `proxy_mean`, and labeled correlation fall within the specified tolerances.
- [ ] The functional test `test_generate_clustered_binary_dataset_no_partial_clusters_after_annotation` passes: no cluster is partially labeled after applying a cluster-level `xi` through `simulate_annotation`.
- [ ] `tests/unit/simulators/test_clustered_binary.py` exists and mirrors the structure of `test_binary.py`.
- [ ] A `None`-seed non-determinism test is added to each of `test_binary.py`, `test_stratified_binary.py`, `test_oracle_binary.py`, and `test_gaussian.py`, following the pattern in the Propagation section above.
- [ ] The API reference is updated to include `generate_clustered_binary_dataset`.

imerad:
### Description

- What does this PR do?
Adds `generate_clustered_binary_dataset` to `glide/simulators/`, which generates synthetic binary-label data partitioned into randomly-sized clusters. Delegates data generation to `generate_binary_dataset` and partitions observations via random cut points. Also completes a refactoring pass on existing simulator unit tests — adding different seed non-determinism tests to `test_binary.py`, `test_oracle_binary.py`, `test_gaussian.py`, and updating mock delegation assertions in `test_stratified_binary.py` to match the current call signatures.
- Which issue does it close? Closes #234
The unit test refactoring was not part of the original ticket scope but was needed to complete an incomplete refactoring in the existing simulator test files.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
## Code Review — `feat/simulate_clustered_binary_data`

Reviewed `glide/simulators/clustered_binary.py` and all changed test files. Implementation logic is correct; findings are concentrated in test coverage gaps.

---

### 🔴 Missing error-path tests (CONFIRMED)

**`tests/unit/simulators/test_clustered_binary.py` — no `pytest.raises` for `n_clusters < 2` or `n_total < n_clusters`**

The only coverage for these two validation branches is `test_generate_clustered_binary_dataset_delegates`, which mocks `_validate_bounds` to a no-op. No test ever calls the function with, e.g., `n_clusters=1` or `n_total=1, n_clusters=2` and asserts a `ValueError` is raised. Per the project convention ("each distinct error condition gets its own test function"), two tests are missing:

```python
def test_generate_clustered_binary_dataset_invalid_n_clusters_raises():
    with pytest.raises(ValueError, match="'n_clusters' must be >= 2"):
        generate_clustered_binary_dataset(n_total=4, n_clusters=1)

def test_generate_clustered_binary_dataset_n_total_less_than_n_clusters_raises():
    with pytest.raises(ValueError, match="'n_total' must be >= 'n_clusters'"):
        generate_clustered_binary_dataset(n_total=1, n_clusters=2)
```

---

### 🟠 Functional test incomplete cluster-structure assertion (CONFIRMED)

**`tests/functional/simulators/test_clustered_binary.py`, line 24**

```python
assert np.max(clusters) == n_clusters - 1
```

This only checks the maximum cluster id. A bug that drops one cluster id and duplicates another would still pass. The minimum and uniqueness are not verified. Suggested fix:

```python
assert np.min(clusters) == 0
assert len(np.unique(clusters)) == n_clusters
assert np.max(clusters) == n_clusters - 1
```

---

### 🟠 Deleted end-to-end error-message tests (PLAUSIBLE)

**`tests/unit/simulators/test_binary.py`, `test_oracle_binary.py`, `test_gaussian.py`**

The three pairs of `pytest.raises(ValueError, match=...)` tests were replaced with delegation mocks. The mock tests do protect against accidental removal of the `_validate_bounds` calls (if the call disappears, `assert_has_calls` fails). However, the original tests additionally verified the exact error message text (e.g., `"'true_mean' must be in (0, 1)"`). If `_validate_bounds` were refactored to emit a different message, or if the arguments passed to it drifted (e.g. wrong bound values), no remaining test would catch the regression at the simulator level.

---

### 🟡 No integer-type validation for `n_clusters` / `n_total` (PLAUSIBLE)

**`glide/simulators/clustered_binary.py`, line 99**

`_validate_bounds` only checks `value >= lower`; it does not verify the value is an integer. A float like `n_clusters=2.5` satisfies `2.5 >= 2` and passes, then fails inside `rng.choice(..., size=1.5)` with a NumPy `TypeError` rather than a clean `ValueError` at the public API boundary. The project error-handling convention requires raising `ValueError` for invalid inputs (wrong type). Adding `_validate_is_integer(n_clusters, "n_clusters")` and `_validate_is_integer(n_total, "n_total")` before the bounds checks would fix this.

---

### 🔵 Minor: `np.arange(1, n_total)` allocation in `rng.choice` (PLAUSIBLE)

**`glide/simulators/clustered_binary.py`, line 122**

```python
cut_positions = np.sort(rng.choice(np.arange(1, n_total), size=n_clusters - 1, replace=False))
```

`np.arange(1, n_total)` allocates an O(n_total) array solely to be consumed by `rng.choice`. The idiomatically equivalent form avoids the allocation:

```python
cut_positions = np.sort(rng.choice(n_total - 1, size=n_clusters - 1, replace=False) + 1)
```

`rng.choice(n, ...)` samples from `{0, ..., n-1}`, so adding 1 shifts to `{1, ..., n_total-1}` — same distribution, no intermediate array.

gmartinon-ed (approved):
Perfetto ! Merci d'avoir fait la propagation de refacto ! C'est Claude qui l'a vu ?

Une question sur les `test_different_seed_results_differ`: il me semblait qu'on l'avait fait dans une PR précédente, mais je ne le vois nulle part (ni dans les samplers, ni dans les estimateurs PTD). J'ai raté quelque chose ou on avait arbitré que c'était pas nécessaire ?

Bref, il faut trancher soit dans cette PR soit un ticket : supprimer ces tests ou compléter ces tests à toutes les classes ayant des random seed.

imerad:
J'avais remarqué que la refacto n'avait pas atteint le module `simulators` en y ajoutant la nouvelle feature.
Les `test_different_seed_results_differ` ont été propagés aux autres modules.

PR #245, opened 2026-06-03
Add UniformClusterSampler for cluster-level annotation selection

### LINKED ISSUE #235: Implement ClusterSampler
# Implement ClusterSampler

## Background

All existing GLIDE samplers work at the level of individual observations: they select which specific rows to annotate. Cluster sampling is different — it selects entire groups of observations at once, and every observation in a selected group is annotated. This is the right choice when annotations within a group are naturally done together: for example, when a group is a conversation thread, a document, or a batch of outputs from a single model run. Annotating every row in a selected group removes the need to set up individual annotation tasks for each one, and it ensures that correlated observations (which often appear in the same group) are handled consistently.

The key input is `clusters`, an array of cluster identifiers, one per observation. The sampler draws `n_clusters` clusters from the pool of unique cluster labels and marks every observation in those clusters for annotation. Two strategies control which clusters are drawn. Under **uniform** sampling, every cluster has exactly the same probability of being selected regardless of size. Because every observation in a selected cluster is annotated, this means every individual observation has the same marginal probability of being annotated (`n_clusters / K`, where `K` is the total number of clusters), regardless of which cluster it belongs to. Under **proportional** sampling, each cluster is drawn with probability proportional to its size. Larger clusters are therefore more likely to be selected, and their observations are more likely to be annotated — a singleton cluster competes on very unequal footing with a cluster that contains half the dataset.

There is no proxy signal involved and no within-cluster sub-sampling: the budget is expressed in number of clusters, not in number of observations. The output is a binary indicator array `xi` of the same length as `clusters`, with 1 for every observation in a selected cluster and 0 otherwise. This design matches how the downstream cluster-aware estimators will consume the result.

## Design choices

`ClusterSampler` lives in `glide/samplers/cluster.py` and is exported from `glide/samplers/__init__.py` alongside the other samplers. Its structure mirrors `StratifiedSampler`: a private `_validate` helper handles domain-specific input checks, and `sample` is the single public method. There is no `__init__` constructor — like the other samplers, `ClusterSampler` is stateless and requires no configuration at construction time.

The `sample` signature takes `clusters` (the cluster labels), `n_clusters` (how many clusters to draw), `strategy` (defaulting to `"uniform"` since it is the simpler and more familiar baseline), and `random_seed`. Unlike `StratifiedSampler`, which takes a `y_proxy` array and allocates a sample-level budget, `ClusterSampler` takes no proxy values and its budget is cluster-level. Naming the parameter `n_clusters` rather than `budget` avoids ambiguity across the public API, where `budget` consistently means a count of individual observations. The return type is `NDArray` of integers (matching `StratifiedSampler`), not a `Tuple` like `ActiveSampler`, because cluster sampling without replacement produces deterministic selection probabilities and requires no inverse-probability weighting arrays.

For the proportional strategy, selection probabilities are computed as `cluster_size / total_observations`. NumPy's `rng.choice` with the `p` parameter handles the without-replacement draw natively. Cluster sizes are computed with `np.unique(clusters, return_counts=True)` to stay fully vectorized. The final `xi` array is built with `np.isin`, which avoids an explicit loop over selected clusters.

### ClusterSampler

**Location:** `glide/samplers/cluster.py`

**Purpose:** Draws entire clusters without replacement and returns per-observation selection indicators.

```python
from typing import Literal, Optional, Union

import numpy as np
from numpy.random.bit_generator import SeedSequence


class ClusterSampler:
    """Sampler that selects entire clusters without replacement.

    Each call to ``sample`` draws a fixed number of clusters from the pool of unique cluster labels in ``clusters``, then marks every observation in a selected cluster for annotation. Two strategies control how clusters are drawn:

    - **Uniform** (default): every cluster has equal probability of being selected, so every individual observation has the same marginal probability of being annotated.
    - **Proportional**: each cluster is drawn with probability proportional to its size, so larger clusters are more likely to be selected and their observations more likely to be annotated.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.samplers import ClusterSampler
    >>> clusters = np.array(["A", "A", "B", "B"], dtype=object)
    >>> sampler = ClusterSampler()
    >>> xi = sampler.sample(clusters, n_clusters=2, random_seed=0)
    >>> xi
    array([1, 1, 1, 1])
    """

    def _validate(self, clusters: NDArray, n_clusters: int) -> None:
        unique_clusters = np.unique(clusters)
        n_total_clusters = len(unique_clusters)
        if n_clusters > n_total_clusters:
            raise ValueError(
                f"'n_clusters' must not exceed the number of unique clusters; "
                f"got n_clusters={n_clusters} but clusters contains {n_total_clusters} unique clusters."
            )

    def sample(
        self,
        clusters: NDArray,
        n_clusters: int,
        strategy: Literal["uniform", "proportional"] = "proportional",
        random_seed: Optional[Union[int, SeedSequence]] = None,
    ) -> NDArray:
        """Select entire clusters without replacement.

        Draws ``n_clusters`` clusters from the unique values of ``clusters`` and returns selection indicators: every observation whose cluster was drawn receives a 1, all others receive a 0.

        Parameters
        ----------
        clusters : NDArray
            Cluster identifiers for all observations, shape ``(n_samples,)``.
        n_clusters : int
            Number of clusters to select. Must be a strictly positive integer and must not exceed the number of unique clusters in ``clusters``.
        strategy : str, optional
            Cluster-selection strategy. ``"uniform"`` draws clusters with equal probability, giving every observation the same marginal probability of being annotated (default). ``"proportional"`` draws with probability proportional to cluster size, so larger clusters are more likely to be selected.
        random_seed : int or SeedSequence or None, optional
            Random seed passed to ``numpy.random.default_rng`` for
            reproducibility. Defaults to ``None`` (non-deterministic).

        Returns
        -------
        NDArray
            Selection indicators of shape ``(n_samples,)``: 1 if the observation belongs to a selected cluster, 0 otherwise.

        Raises
        ------
        ValueError
            If ``n_clusters`` is not a strictly positive integer, if ``n_clusters``
            exceeds the number of unique clusters in ``clusters``, or if ``strategy``
            is not recognised.
        """
        if (not isinstance(n_clusters, (int, np.integer))) or isinstance(n_clusters, bool) or n_clusters <= 0:
            raise ValueError(f"'n_clusters' must be a strictly positive integer; got {n_clusters!r}.")

        self._validate(clusters, n_clusters)

        unique_clusters, cluster_sizes = np.unique(clusters, return_counts=True)

        if strategy == "uniform":
            selected_clusters = rng.choice(unique_clusters, size=n_clusters, replace=False)
        elif strategy == "proportional":
            probabilities = cluster_sizes / cluster_sizes.sum()
            selected_clusters = rng.choice(unique_clusters, size=n_clusters, replace=False, p=probabilities)
        else:
            raise ValueError(f"Unknown strategy '{strategy}'. Expected 'uniform' or 'proportional'.")

        xi = np.isin(clusters, selected_clusters).astype(int)
        return xi
```

Also add `ClusterSampler` to `glide/samplers/__init__.py`:

```python
from glide.samplers.active import ActiveSampler
from glide.samplers.cluster import ClusterSampler
from glide.samplers.cost_optimal import CostOptimalSampler
from glide.samplers.cost_optimal_random import CostOptimalRandomSampler
from glide.samplers.stratified import StratifiedSampler
from glide.samplers.uniform import UniformSampler

__all__ = [
    "ActiveSampler",
    "ClusterSampler",
    "CostOptimalRandomSampler",
    "CostOptimalSampler",
    "StratifiedSampler",
    "UniformSampler",
]
```

## Functional tests

**Cluster integrity under both strategies**

Every observation belonging to the same cluster must receive the same xi value: a cluster is either fully selected or fully excluded, never split.

```python
@pytest.mark.parametrize("strategy", ["uniform", "proportional"])
def test_sample_preserves_cluster_integrity(strategy):
    clusters = np.array(["A", "A", "B", "B", "B", "C", "C", "C", "C"], dtype=object)
    sampler = ClusterSampler()
    xi = sampler.sample(clusters, n_clusters=2, strategy=strategy, random_seed=0)

        cluster_xi_values = xi[clusters == cluster_id]
        assert len(np.unique(cluster_xi_values)) == 1
```

**Proportional equals uniform for equal-size clusters (distribution-level)**

When all clusters have the same size, proportional reduces to uniform: selection probabilities are identical under both strategies. This can be verified by showing that both strategies produce the same result for a fixed seed.

```python
def test_sample_proportional_equals_uniform_for_equal_size_clusters():
    clusters = np.array(["A", "A", "B", "B", "C", "C"], dtype=object)
    sampler = ClusterSampler()

    xi_uniform = sampler.sample(clusters, n_clusters=2, strategy="uniform", random_seed=42)
    xi_proportional = sampler.sample(clusters, n_clusters=2, strategy="proportional", random_seed=42)

    np.testing.assert_array_equal(xi_uniform, xi_proportional)
```

**Unit-size clusters are equivalent to UniformSampler (distribution-level)**

When every cluster contains exactly one observation, `ClusterSampler` with `strategy="uniform"` and `UniformSampler` select observations from the same distribution: each observation is chosen with probability `n_clusters / n_total`. Because the two samplers call `rng.choice` on different types (strings vs integers), outputs are not path-identical for the same seed — the equivalence is distributional and must be verified with a Monte Carlo loop.

```python
def test_sample_unit_clusters_matches_uniform_sampler_distribution():
    clusters = np.array(["A", "B", "C", "D", "E", "F"], dtype=object)
    n_clusters = 3
    n_total = len(clusters)
    n_seeds = 5_000

    cluster_counts = np.zeros(n_total, dtype=int)
    uniform_counts = np.zeros(n_total, dtype=int)
    for seed in range(n_seeds):
        cluster_counts += ClusterSampler().sample(clusters, n_clusters=n_clusters, strategy="uniform", random_seed=seed)
        uniform_counts += UniformSampler().sample(n_total=n_total, n_samples=n_clusters, random_seed=seed).astype(int)

    expected_frequency = n_clusters / n_total
    np.testing.assert_allclose(cluster_counts / n_seeds, expected_frequency, atol=0.02)
    np.testing.assert_allclose(uniform_counts / n_seeds, expected_frequency, atol=0.02)
```

## Corner cases

- `n_clusters=0` → `ValueError` with message mentioning `'n_clusters'`.
- `n_clusters=-1` → `ValueError` with message mentioning `'n_clusters'`.
- `n_clusters=1.5` → `ValueError` with message mentioning `'n_clusters'`.
- `n_clusters=True` → `ValueError` (booleans must be rejected even though `bool` is a subclass of `int`).
- `n_clusters` exceeds the number of unique clusters in `clusters` → `ValueError` mentioning both the requested value and the available count.
- `strategy="unknown"` → `ValueError` with message mentioning `"Unknown strategy"`.
- `n_clusters` equals the number of unique clusters → all observations are selected, `xi` is all ones.
- Single cluster in `clusters`, `n_clusters=1` → all observations selected, `xi` is all ones.
- Clusters of very different sizes with `strategy="proportional"` → the largest cluster is drawn most often; verify cluster integrity still holds.

## Acceptance criteria

- [ ] `ClusterSampler` exists in `glide/samplers/cluster.py` and is exported from `glide/samplers/__init__.py`.
- [ ] `sample` returns `xi` of shape `(n_samples,)` with integer dtype, containing only 0s and 1s, where every observation in a selected cluster has value 1 and every observation in a non-selected cluster has value 0.
- [ ] Both `"uniform"` and `"proportional"` strategies are implemented; `"uniform"` is the default.
- [ ] Cluster integrity holds: no cluster contains a mix of 0s and 1s in the output.
- [ ] For equal-size clusters, `"proportional"` and `"uniform"` produce identical results for the same seed.
- [ ] Unit tests in `tests/unit/samplers/test_cluster.py` cover `_validate` and all branches of `sample`, following the structure of `tests/unit/samplers/test_stratified.py`.

## Open question
There is an uncertainty on the proportional sampling: does it fit into the hypotheses of the ClusterPPI and ClusterPTD estimators (developed later)? Or should we imagine a different workflow for proportional sampling, e.g. ClusterSampler -> IPWClusterPPIMeanEstimatior (not framed yet) ? For the time being, stick to this implementation, and further investigation will have to clarify this point.

imerad:
### Description

Adds `UniformClusterSampler`, a new sampler that selects entire clusters without replacement using uniform sampling and returns per-observation selection indicators (1 for every observation in a selected cluster, 0 otherwise).

Closes #235

The proportional sampling strategy was not implemented, as it is difficult to guarantee statistical validity for downstream estimation. As a result, the sampler was named `UniformClusterSampler` rather than `ClusterSampler`. The implementation uses the centralized parameter validation functions from `glide.core.validation` introduced in the recent refactor.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad (commented):
## Code Review — `feat/cluster_sampler`

Four findings survive verification (two confirmed, two plausible). Ordered by severity.

---

### 1. Fragile test: `test_sample_different_seed_results_differ` ✅ CONFIRMED

**File:** `tests/unit/samplers/test_cluster.py` (line 63)

With three unique clusters and `n_clusters=2`, there are only C(3,2)=3 possible subsets. Seeds 0 and 1 happen to differ, but seeds 2 & 3, 3 & 6, and 2 & 6 all produce the identical subset `[1, 1, 1, 1, 0, 0]`. The test gives false confidence because it asserts one coincidentally-different pair. A future NumPy release changing the PRNG sequence or a change to the fixture could make it fail non-deterministically.

**Fix:** drop the test and instead assert that the *number* of selected clusters is correct, or seed it deterministically with a pair that is guaranteed to differ (verified by inspection, documented).

---

### 2. `np.unique(clusters)` computed twice per `sample()` call ✅ CONFIRMED

**File:** `glide/samplers/cluster.py` (lines 35 and 79)

`_validate` computes `np.unique(clusters)` to check the upper bound, then `sample` recomputes it immediately. Both calls run on every `sample()` invocation, wasting a full sort of `clusters` each time. There is also a latent inconsistency risk if `_validate` is ever refactored to use the precomputed value without updating `sample`.

**Fix:** compute `unique_clusters = np.unique(clusters)` once in `sample`, pass it (or just `n_total_clusters`) into `_validate`, or inline the check.

---

### 3. Silent failure when `clusters` contains NaN float labels 🔶 PLAUSIBLE

**File:** `glide/samplers/cluster.py` (lines 32–35, 83)

`np.unique` treats each NaN as a distinct cluster. `rng.choice` may then select a NaN cluster, but `np.isin(clusters, [nan])` always returns `False` (NaN ≠ NaN), so all observations nominally belonging to that cluster receive `xi=0`. The class docstring promises "every individual observation has the same marginal probability of being annotated" — this is silently violated for NaN-labeled observations without any error.

The practical risk is low because cluster IDs are typically strings or integers (as in all doctests and tests), but there is no type or NaN guard on `clusters`.

**Fix:** add `_validate_has_no_nan(clusters, "clusters")` (or an equivalent check) in `_validate`.

---

### 4. No shape validation on `clusters`; 2-D input returns wrong-shape `xi` 🔶 PLAUSIBLE

**File:** `glide/samplers/cluster.py` (line 79, 83)

`np.unique` flattens multi-dimensional arrays silently. If `clusters` is 2-D (e.g., shape `(n, 1)` after a reshape), `np.unique` collapses it to 1-D for counting, but `np.isin(clusters, selected_clusters)` broadcasts element-wise on the original shape, returning a `(n, 1)` xi instead of `(n,)`. No error is raised; downstream estimators receive a silently wrong shape.

**Fix:** add `if clusters.ndim != 1: raise ValueError(...)` in `_validate`, consistent with how other samplers constrain array inputs.

---

*Below the threshold (not actionable):*
- Doctest output is correct (verified by running the code).
- `SeedSequence` private import path is consistent with every other sampler in the repo.
- `dtype=int` return is consistent with `StratifiedSampler`; NumPy arithmetic handles int/float xi transparently.
- Empty `clusters` with `n_clusters=1` does raise `ValueError`; the message ("0 unique clusters") is clear enough.

glide/samplers/cluster.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        _validate_is_integer(n_clusters, "n_clusters")
+        _validate_strictly_positive(n_clusters, "n_clusters")
+        unique_clusters = np.unique(clusters)
+        n_total_clusters = len(unique_clusters)
+        if n_clusters > n_total_clusters:
+            raise ValueError(
+                f"'n_clusters' must not exceed the number of unique clusters; "
+                f"got n_clusters={n_clusters} but there are only {n_total_clusters} unique clusters."
+            )
```
gmartinon-ed: C'est pas un validate bounds ça ?

tests/unit/samplers/test_cluster.py
```diff
@@ -0,0 +1,64 @@
...
+    assert isinstance(xi, np.ndarray)
+    assert xi.shape == (len(clusters),)
+    assert np.isin(xi, [0, 1]).all()
+
+
+def test_sample_known_output(sampler, clusters):
+    xi = sampler.sample(clusters, n_clusters=2, random_seed=0)
+
+    expected_xi = np.array([0, 0, 1, 1, 1, 1])
+    np.testing.assert_array_equal(xi, expected_xi)
```
gmartinon-ed: A fusionner ?

gmartinon-ed (changes requested):

tests/unit/samplers/test_cluster.py
imerad: Fait

glide/samplers/cluster.py
imerad: Certes, j'étais un peu réticent à le faire car ça fait plus de lignes de code. C'est remplacé maintenant

gmartinon-ed (approved):

PR #254, opened 2026-06-04
Add ClusterClassicalMeanEstimator for cluster-sampling-aware mean estimation

### LINKED ISSUE #203: Implement ClusterClassicalMeanEstimator
# ClusterClassicalMeanEstimator: classical mean estimation for cluster-sampled data

## Background

When observations are grouped into clusters (e.g., all questions from the same conversation thread, or all items from the same document), standard statistical methods underestimate uncertainty. The standard classical estimator assumes every observation is independent, so it treats 100 observations from 2 clusters of 50 the same as 100 independent observations. In reality, items from the same cluster are correlated: they share context, style, and model behavior, which inflates confidence. The cluster classical estimator corrects for this by treating each cluster as a single sampling unit rather than individual observations.

The key idea is to replace individual observations with cluster-level summaries before computing the mean and its standard error. For each cluster k, compute the cluster mean μ_k and the cluster size n_k. The population mean estimate is the size-weighted average of these cluster means: in math notation, θ̂ = Σ_k (n_k * μ_k) / N, where N = Σ_k n_k is the total number of observations. A larger cluster contributes proportionally more to the estimate, just as larger strata do in `StratifiedClassicalMeanEstimator`.

For the standard error, define the cluster sum u_k = n_k * μ_k. The variance of θ̂ is estimated as Var(θ̂) = K * Var(u_k, ddof=1) / N², where K is the number of clusters. This treats the u_k as the i.i.d. sampling units and applies the standard CLT to their mean. When all clusters have the same size s, this simplifies to Var(μ_k, ddof=1) / K (the variance of the cluster mean average), which is the natural cluster-aware analogue of the single-observation formula Var(y, ddof=1) / n. This variance formula is also identical to the one used for the unlabeled part in `ClusterPTDMeanEstimator`, so both estimators share a consistent treatment of cluster-level uncertainty.

## Design choices

`ClusterClassicalMeanEstimator` lives in `glide/estimators/cluster_classical.py` and mirrors `StratifiedClassicalMeanEstimator` (`glide/estimators/stratified_classical.py`) in structure, API, and naming. The `estimate` method takes `clusters` instead of `groups`, but the signature is otherwise identical. Like `StratifiedClassicalMeanEstimator`, this estimator uses `CLTConfidenceInterval` (not bootstrap) and returns `ClassicalMeanInferenceResult`. All observations are assumed to be labeled: there is no labeled/unlabeled distinction, and no `y_proxy` argument. NaN values in `y` are silently dropped at the per-cluster level, exactly as `StratifiedClassicalMeanEstimator` does per stratum.

The variance is computed across cluster sums (not within clusters), so only 1 non-NaN observation per cluster is required. The minimum constraint is at least 2 clusters with at least 1 valid observation each, so that `np.var(cluster_sums, ddof=1)` is defined. This differs from `StratifiedClassicalMeanEstimator`, which requires at least 2 observations per stratum because it computes within-stratum variance. Clusters that become empty after NaN removal are skipped silently, just as `StratifiedClassicalMeanEstimator` would raise only when the remaining count falls below its threshold.

The preprocessing and aggregation logic is inlined in the `estimate` method, following the same pattern as `StratifiedClassicalMeanEstimator`. There is no separate helper module. The `n` field in the returned `ClassicalMeanInferenceResult` is the total number of non-NaN individual observations (Σ_k n_k), consistent with how both `ClassicalMeanEstimator` and `StratifiedClassicalMeanEstimator` set `n`.

### `ClusterClassicalMeanEstimator`

**Location:** `glide/estimators/cluster_classical.py`

**Purpose:** Public estimator class that computes a size-weighted CLT-based confidence interval for the population mean under cluster sampling.

```python
from typing import List

import numpy as np

from glide.core.mean_inference_result import ClassicalMeanInferenceResult


class ClusterClassicalMeanEstimator:
    """Cluster classical estimator for population mean.

    Extends mean estimation as in ``ClassicalMeanEstimator`` to datasets where
    observations are grouped into clusters. Each cluster's size-weighted
    contribution is treated as the sampling unit, which accounts for
    within-cluster correlation and produces valid confidence intervals under
    cluster sampling designs.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.estimators import ClusterClassicalMeanEstimator
    >>> y = np.array([5.0, 5.0, 7.0, 7.0])
    >>> clusters = np.array(["A", "A", "B", "B"])
    >>> estimator = ClusterClassicalMeanEstimator()
    >>> result = estimator.estimate(y, clusters)
    >>> print(result)
    Metric: Metric
    Point Estimate: 6.000
    Confidence Interval (95%): [4.040, 7.960]
    Estimator : ClusterClassicalMeanEstimator
    n: 4
    """

    def estimate(
        self,
        y: NDArray,
        clusters: NDArray,
        metric_name: str = "Metric",
    ) -> ClassicalMeanInferenceResult:
        """Estimate the population mean using the cluster classical estimator.

        Aggregates observations into cluster-level means weighted by cluster
        size, then applies the CLT to the cluster sums as sampling units:

            theta = sum_k  (n_k * mu_k) / N
            sigma2 = K * Var(u_k, ddof=1) / N^2

        where ``u_k = n_k * mu_k`` are the cluster sums, ``K`` is the number
        of clusters, and ``N = sum_k n_k`` is the total number of observations.
        NaN values in ``y`` are dropped per cluster before computing cluster
        means. Clusters that contain only NaN are skipped.

        Parameters
        ----------
        y : NDArray
            Array of observations, shape ``(n_samples,)``. NaN values are
            treated as missing and dropped per cluster.
        clusters : NDArray
        metric_name : str, optional

        Returns
        -------
        ClassicalMeanInferenceResult
            Contains the CLT-based confidence interval, the metric name,
            the estimator name (``"ClusterClassicalMeanEstimator"``), and ``n``
            (total number of non-NaN observations across all clusters).

        Raises
        ------
        ValueError
            - If ``y`` and ``clusters`` do not have the same length.
            - If fewer than 2 clusters have at least one non-NaN observation.
        """
        if len(y) != len(clusters):
            raise ValueError(
                f"y and clusters must have the same length, got {len(y)} and {len(clusters)}."
            )

        cluster_means: List[float] = []
        cluster_sizes: List[int] = []

            cluster_mask = clusters == cluster_id
            cluster_y = y[cluster_mask]
            cluster_y_valid = cluster_y[~np.isnan(cluster_y)]
            if len(cluster_y_valid) == 0:
                continue
            cluster_means.append(np.mean(cluster_y_valid))
            cluster_sizes.append(len(cluster_y_valid))

        n_clusters = len(cluster_means)
        if n_clusters < 2:
            raise ValueError(
                f"Need at least 2 clusters with non-NaN observations; got {n_clusters}."
            )

        means = np.array(cluster_means)
        sizes = np.array(cluster_sizes)
        cluster_sums = sizes * means
        total_size = np.sum(sizes)

        weighted_mean = np.sum(cluster_sums) / total_size
        var = n_clusters * np.var(cluster_sums, ddof=1) / total_size**2
        std = np.sqrt(var)

        ci = CLTConfidenceInterval(
            mean=weighted_mean,
            std=std,
        )
        result = ClassicalMeanInferenceResult(
            confidence_interval=ci,
            metric_name=metric_name,
            n=total_size,
        )
        return result
```

Add the following import and `__all__` entry (keep alphabetical order within each group):

```python
from glide.estimators.cluster_classical import ClusterClassicalMeanEstimator
```

Add `"ClusterClassicalMeanEstimator"` to `__all__`.

## Functional tests

**Single-observation clusters produce identical results to `ClassicalMeanEstimator`**

When every cluster contains exactly one observation, `means` equals `y`, `sizes` equals `np.ones(K)`, cluster sums equal `y`, and `Var(θ̂) = K * Var(y, ddof=1) / K² = Var(y, ddof=1) / K`, which is exactly what `ClassicalMeanEstimator` computes. The full confidence interval (including bounds and variance) must be identical.

```python
def test_single_observation_clusters_equals_classical():
    y = np.array([5.0, 7.0, 4.0, 8.0])  # IMPORTANT: Increase the number of data points for the functional test
    clusters = np.array(["A", "B", "C", "D"])   # IMPORTANT: Increase the number of data points for the functional test

    cluster_result = ClusterClassicalMeanEstimator().estimate(y, clusters)
    classical_result = ClassicalMeanEstimator().estimate(y)

        classical_result.confidence_interval.lower_bound, abs=1e-10
    )
        classical_result.confidence_interval.upper_bound, abs=1e-10
    )
```

**Equal-size clusters produce identical results to `ClassicalMeanEstimator`**

TBC

## Corner cases

- `y` and `clusters` of different lengths must raise `ValueError`.
- Fewer than 2 clusters with non-NaN observations must raise `ValueError`.
- A cluster where all observations are NaN is silently skipped and does not count toward the cluster total; the error is raised only if this causes the total to fall below 2.

## Acceptance criteria

- [ ] `ClusterClassicalMeanEstimator` is importable from `glide.estimators` and listed in `__all__`.
- [ ] The analytical doctest passes: for 2 clusters of size 2 with means 5.0 and 7.0, `confidence_interval.mean == 6.0` and `std == 1.0` (derivable by hand: cluster sums are 10 and 14, `Var([10,14], ddof=1) = 8`, `Var(θ̂) = 2 * 8 / 16 = 1`).
- [ ] `estimate` raises `ValueError` when fewer than 2 clusters have non-NaN observations.
- [ ] `result.n` equals the total number of non-NaN individual observations: for 2 clusters of size 3 each, `result.n == 6`.
- [ ] `tests/unit/estimators/test_cluster_classical.py` exists and mirrors the structure of `test_stratified_classical.py`.
- [ ] The functional test `test_single_observation_clusters_equals_classical` passes in `tests/functional/`.

imerad:
### Description

- Implements `ClusterClassicalMeanEstimator`, a CLT-based classical mean estimator that treats each cluster as the sampling unit (via cluster sums) rather than individual observations, producing valid confidence intervals under cluster sampling designs.
- Closes #203
- Per-cluster computations use sums (`u_k = n_k * μ_k`) rather than means, so the variance formula `K * Var(u_k, ddof=1) / N²` is consistent with the unlabeled-part treatment in `ClusterPTDMeanEstimator`. The `_preprocess` method centralises all validation checks and unique cluster identification, following the same factorisation pattern as `StratifiedClassicalMeanEstimator`.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/core/validation.py
+++ b/glide/core/validation.py
@@ -11,8 +11,12 @@ def _validate_non_constant(array: NDArray, error_message: str) -> None:
 
 
 def _validate_has_no_nan(array: NDArray, name: str) -> None:
-    if np.isnan(array).any():
-        raise ValueError(f"'{name}' contains NaN values.")
+    if np.issubdtype(array.dtype, np.number):
+        if np.isnan(array).any():
+            raise ValueError(f"'{name}' contains NaN values.")
+    else:
+        if any(v is None for v in array):
+            raise ValueError(f"'{name}' contains None values.")
 
 
 def _get_non_zero_mask(values: NDArray, warning_message: Optional[str] = None) -> NDArray:
--- a/glide/core/validation.py
+++ b/glide/core/validation.py
@@ -15,7 +15,7 @@ def _validate_has_no_nan(array: NDArray, name: str) -> None:
         if np.isnan(array).any():
             raise ValueError(f"'{name}' contains NaN values.")
     else:
-        if any(v is None for v in array):
+        if None in array:
             raise ValueError(f"'{name}' contains None values.")
 
 
--- a/glide/estimators/cluster_classical.py
+++ b/glide/estimators/cluster_classical.py
@@ -1,0 +1,120 @@
+from typing import List
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.confidence_intervals import CLTConfidenceInterval
+from glide.core.validation import _validate_equal_lengths
+from glide.mean_inference_results import ClassicalMeanInferenceResult
+
+
+class ClusterClassicalMeanEstimator:
+    """Cluster classical estimator for population mean.
+
+    Extends mean estimation as in ``ClassicalMeanEstimator`` to datasets where
+    observations are grouped into clusters. Each cluster's size-weighted
+    contribution is treated as the sampling unit, which accounts for
+    within-cluster correlation and produces valid confidence intervals under
+    cluster sampling designs.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from glide.estimators import ClusterClassicalMeanEstimator
+    >>> y = np.array([5.0, 5.0, 7.0, 7.0])
+    >>> clusters = np.array(["A", "A", "B", "B"])
+    >>> estimator = ClusterClassicalMeanEstimator()
+    >>> result = estimator.estimate(y, clusters)
+    >>> print(result)
+    Metric: Metric
+    Point Estimate: 6.000
+    Confidence Interval (95%): [4.040, 7.960]
+    Estimator : ClusterClassicalMeanEstimator
+    n: 4
+    """
+
+    def estimate(
+        self,
+        y: NDArray,
+        clusters: NDArray,
+        metric_name: str = "Metric",
+        confidence_level: float = 0.95,
+    ) -> ClassicalMeanInferenceResult:
+        """Estimate the population mean using the cluster classical estimator.
+
+        Aggregates observations into cluster-level means weighted by cluster
+        size, then applies the CLT to the cluster sums as sampling units:
+
+            theta = sum_k  (n_k * mu_k) / N
+            sigma2 = K * Var(u_k, ddof=1) / N^2
+
+        where ``u_k = n_k * mu_k`` are the cluster sums, ``K`` is the number
+        of clusters, and ``N = sum_k n_k`` is the total number of observations.
+        NaN values in ``y`` are dropped per cluster before computing cluster
+        means. Clusters that contain only NaN are skipped.
+
+        Parameters
+        ----------
+        y : NDArray
+            Array of observations, shape ``(n_samples,)``. NaN values are
+            treated as missing and dropped per cluster.
+        clusters : NDArray
+            Array of cluster identifiers, shape ``(n_samples,)``.
+            Unique values define the clusters.
+        metric_name : str, optional
+            Human-readable label for the metric. Defaults to ``"Metric"``.
+        confidence_level : float, optional
+            Target coverage for the confidence interval. Defaults to ``0.95``.
+
+        Returns
+        -------
+        ClassicalMeanInferenceResult
+            Contains the CLT-based confidence interval, the metric name,
+            the estimator name (``"ClusterClassicalMeanEstimator"``), and ``n``
+            (total number of non-NaN observations across all clusters).
+
+        Raises
+        ------
+        ValueError
+            - If ``y`` and ``clusters`` do not have the same length.
+            - If fewer than 2 clusters have at least one non-NaN observation.
+        """
+        _validate_equal_lengths(y, clusters, names=["y", "clusters"])
+
+        cluster_means: List[float] = []
+        cluster_sizes: List[int] = []
+
+        for cluster_id in np.unique(clusters):
+            cluster_mask = clusters == cluster_id
+            cluster_y = y[cluster_mask]
+            cluster_y_valid = cluster_y[~np.isnan(cluster_y)]
+            if len(cluster_y_valid) == 0:
+                continue
+            cluster_means.append(np.mean(cluster_y_valid))
+            cluster_sizes.append(len(cluster_y_valid))
+
+        n_clusters = len(cluster_means)
+        if n_clusters < 2:
+            raise ValueError(f"Need at least 2 clusters with non-NaN observations; got {n_clusters}.")
+
+        means = np.array(cluster_means)
+        sizes = np.array(cluster_sizes)
+        cluster_sums = sizes * means
+        total_size = np.sum(sizes)
+
+        weighted_mean = np.sum(cluster_sums) / total_size
+        var = n_clusters * np.var(cluster_sums, ddof=1) / total_size**2
+        std = np.sqrt(var)
+
+        ci = CLTConfidenceInterval(
+            mean=weighted_mean,
+            std=std,
+            confidence_level=confidence_level,
+        )
+        result = ClassicalMeanInferenceResult(
+            confidence_interval=ci,
+            metric_name=metric_name,
+            estimator_name=self.__class__.__name__,
+            n=total_size,
+        )
+        return result
--- a/glide/estimators/cluster_classical.py
+++ b/glide/estimators/cluster_classical.py
@@ -39,8 +39,7 @@ class ClusterClassicalMeanEstimator:
         clusters: NDArray,
     ) -> Tuple[NDArray, NDArray, int]:
         _validate_equal_lengths(y, clusters, names=["y", "clusters"])
-        if np.issubdtype(clusters.dtype, np.number):
-            _validate_has_no_nan(clusters, "clusters")
+        _validate_has_no_nan(clusters, "clusters")
         not_nan_mask = ~np.isnan(y)
         y_valid = y[not_nan_mask]
         clusters_valid = clusters[not_nan_mask]
@@ -64,22 +63,22 @@ class ClusterClassicalMeanEstimator:
     ) -> ClassicalMeanInferenceResult:
         """Estimate the population mean using the cluster classical estimator.
 
-        Aggregates observations into cluster-level means weighted by cluster
-        size, then applies the CLT to the cluster sums as sampling units:
+        Computes within-cluster sums and uses them as sampling units to apply
+        the CLT:
 
-            theta = sum_l  (n_l * mu_l) / N
+            theta = (1 / N) * sum_l u_l
             sigma2 = L * Var(u_l, ddof=1) / N^2
 
-        where ``u_l = n_l * mu_l`` are the cluster sums, ``L`` is the number
-        of clusters, and ``N = sum_l n_l`` is the total number of observations.
-        NaN values in ``y`` are dropped before making the computations. Clusters
-        that contain only NaN are not used.
+        where ``u_l = sum_{i in l} y_i`` are the cluster sums, ``L`` is the
+        number of clusters, and ``N = sum_l n_l`` is the total number of
+        observations. NaN values in ``y`` are dropped before making the
+        computations. Clusters that contain only NaN are not used.
 
         Parameters
         ----------
         y : NDArray
             Array of observations, shape ``(n_samples,)``. NaN values are
-            treated as missing and dropped per cluster.
+            treated as missing and dropped.
         clusters : NDArray
             Array of cluster identifiers, shape ``(n_samples,)``.
             Unique values define the clusters.
@@ -99,7 +98,7 @@ class ClusterClassicalMeanEstimator:
         ------
         ValueError
             - If ``y`` and ``clusters`` do not have the same length.
-            - If ``clusters`` has a numeric dtype and contains NaN values.
+            - If ``clusters`` contains NaN values (numeric dtype) or None values (non-numeric dtype).
             - If fewer than 2 clusters have at least one non-NaN observation.
         """
         y_valid, cluster_indices, n_valid_clusters = self._preprocess(y, clusters)
--- a/glide/estimators/cluster_classical.py
+++ b/glide/estimators/cluster_classical.py
@@ -1,5 +1,3 @@
-from typing import List
-
 import numpy as np
 from numpy.typing import NDArray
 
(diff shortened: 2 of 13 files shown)


imerad:
## Code Review

### 🔴 High — Silent wrong results when `clusters` contains NaN

**File:** `glide/estimators/cluster_classical.py`, line 42

`not_nan_mask` is derived only from `y`, so NaN values in a numeric `clusters` array survive into `clusters_valid`. `np.unique` then treats the NaN as a legitimate third cluster, but the loop condition `clusters_valid == nan` is always `False` (IEEE 754), so that cluster accumulates a sum of zero while the corresponding `y` values are never added to any cluster. No error is raised.

**Concrete example:**
```python
y        = [1.0, 2.0, 3.0, 4.0]
clusters = [1.0, float('nan'), 2.0, 2.0]
# mean → 2.0 (wrong, correct is 2.5)
# CI is completely wrong, no error raised
```

**Fix:** add `_validate_has_no_nan(clusters, "clusters")` in `_preprocess` (the helper already exists in `glide/core/validation.py`), or at minimum filter out rows where the cluster label itself is NaN.

---

### 🟡 Medium — Python `for` loop violates the project's vectorization rule

**File:** `glide/estimators/cluster_classical.py`, lines 107–110

CLAUDE.md states _"Use NumPy vectorization instead of Python for loops whenever possible."_ The loop runs `L` full-array comparisons of length `N` (O(L×N)). A vectorized O(N) replacement:

```python
_, inverse = np.unique(clusters_valid, return_inverse=True)
sums = np.bincount(inverse, weights=y_valid)
```

`np.bincount` supports float `weights`, so this is a direct drop-in.

---

### 🟡 Medium — No `pytest.raises` test for the length-mismatch `ValueError`

**File:** `tests/unit/estimators/test_cluster_classical.py`

CLAUDE.md requires _"each distinct error condition gets its own test function."_ The mock-based delegation test verifies the call is made in the success path, but there is no test asserting that mismatched array lengths actually raise. If the `_validate_equal_lengths` call were accidentally removed or given wrong arguments, nothing in this file would catch it.

```python
def test_preprocess_mismatched_lengths_raises(estimator):
    with pytest.raises(ValueError):
        estimator.estimate(np.array([1.0, 2.0]), np.array(["A"]))
```

---

### 🟡 Medium — No `pytest.raises` test for the fewer-than-2-clusters `ValueError`

**File:** `tests/unit/estimators/test_cluster_classical.py`

Same gap as above. A minimal failing case:

```python
def test_preprocess_single_cluster_raises(estimator):
    y = np.array([5.0, np.nan, np.nan])
    clusters = np.array(["A", "A", "B"])
    with pytest.raises(ValueError, match="Need at least 2 clusters"):
        estimator.estimate(y, clusters)
```

---

### 🔵 Low — `ClusterClassicalMeanEstimator` is out of order in `__all__`

**File:** `glide/estimators/__init__.py`, line 14

It is currently placed second (right after `ClassicalMeanEstimator`), but both `docs/api/estimators.md` and `docs/api/index.md` place it last among Classical variants, after `StratifiedClassicalMeanEstimator` and `IPWClassicalMeanEstimator`. The PTD family in `__all__` follows the Base→Stratified→IPW order correctly. `ClusterClassicalMeanEstimator` should move to after `IPWClassicalMeanEstimator` to stay consistent.

---

### 🔵 Low — `n_valid_clusters` is computed twice

**File:** `glide/estimators/cluster_classical.py`, lines 47 and 103

`_preprocess` computes `n_valid_clusters = len(unique_valid_clusters)` for the bounds check and then discards it (only `unique_valid_clusters` is returned). `estimate` recomputes the same value on line 103. Either return it from `_preprocess` or compute it once in `estimate` before calling the preprocessor.

glide/core/validation.py
```diff
@@ -11,8 +11,12 @@ def _validate_non_constant(array: NDArray, error_message: str) -> None:
 
 
 def _validate_has_no_nan(array: NDArray, name: str) -> None:
-    if np.isnan(array).any():
-        raise ValueError(f"'{name}' contains NaN values.")
+    if np.issubdtype(array.dtype, np.number):
+        if np.isnan(array).any():
+            raise ValueError(f"'{name}' contains NaN values.")
+    else:
+        if any(v is None for v in array):
```
gmartinon-ed: Why ?

glide/estimators/cluster_classical.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _preprocess(
+        self,
+        y: NDArray,
+        clusters: NDArray,
+    ) -> Tuple[NDArray, NDArray, int]:
+        _validate_equal_lengths(y, clusters, names=["y", "clusters"])
+        _validate_has_no_nan(clusters, "clusters")
+        not_nan_mask = ~np.isnan(y)
+        y_valid = y[not_nan_mask]
```
gmartinon-ed, line 44: GOod, petite propagation à ClassicalMeanEstimator ?

glide/estimators/cluster_classical.py
```diff
@@ -0,0 +1,124 @@
...
+        clusters: NDArray,
+        metric_name: str = "Metric",
+        confidence_level: float = 0.95,
+    ) -> ClassicalMeanInferenceResult:
+        """Estimate the population mean using the cluster classical estimator.
+
+        Computes within-cluster sums and uses them as sampling units to apply
+        the CLT:
+
+            theta = (1 / N) * sum_l u_l
```
gmartinon-ed, line 69: Question bête : c'est pas juste np.mean(y) du coup ?

glide/estimators/cluster_classical.py
```diff
@@ -0,0 +1,124 @@
...
+        ------
+        ValueError
+            - If ``y`` and ``clusters`` do not have the same length.
+            - If ``clusters`` contains NaN values (numeric dtype) or None values (non-numeric dtype).
+            - If fewer than 2 clusters have at least one non-NaN observation.
+        """
+        y_valid, cluster_indices, n_valid_clusters = self._preprocess(y, clusters)
+        total_size = len(y_valid)
+
+        cluster_sums = np.bincount(cluster_indices, weights=y_valid)
```
gmartinon-ed, line 107: Le weights ici, c'est pour les cas non-binaires ?

glide/estimators/cluster_classical.py
```diff
@@ -0,0 +1,124 @@
...
+
+        mean = np.sum(cluster_sums) / total_size
```
gmartinon-ed, line 109: c'est pas juste np.mean(y_valid) ?

glide/estimators/cluster_classical.py
```diff
@@ -0,0 +1,124 @@
...
+        var = n_valid_clusters * np.var(cluster_sums, ddof=1) / total_size**2
```
gmartinon-ed, line 110: Question bête : ça semble intuitif comme résultat, mais est-ce que ça se démontre facilement ?

gmartinon-ed (changes requested):
Review incomplète : je n'ai pas encore regardé les tests

glide/estimators/cluster_classical.py
imerad: Oui, il faut écrire l'estimateur, c'est juste la moyenne empirique. Puis en cherchant sa variance, on factorise le 1/N^2 puis on passe de la variance de la somme à la somme des variances. Mais comme les échantillons intracluster ne sont pas indépendants, on se retrouve avec la somme des variances des sommes intraclusters. On va donc estimer la variance d'une somme intracluster et il y avait L termes dans la somme des variances.

glide/estimators/cluster_classical.py
imerad: Si, mais c'est plus efficace de faire comme ça vu qu'on a déjà calculé les sommes intraclusters

glide/estimators/cluster_classical.py
imerad: Non ça marche dans tous les cas. C'est assez malin, la fonction bincount va compter les occurrences des entiers dans `cluster_indices` mais chaque élément comptera pour l'élément à la même position dans `y_valid`. Ça permet de calculer les sommes intraclusters de façon vectorisée et éviter la boucle for avec les masquages, merci Claude.
Je me rends compte qu'on pourrait utiliser ça dans le cas stratifié aussi pour optimiser

glide/estimators/cluster_classical.py
imerad: Si, mais la docstring montre comment le calcul est implémenté

glide/core/validation.py
imerad: On risquait d'avoir des NaNs dans `clusters` et cette valeur serait considérée comme un cluster à part entière. En plus, `clusters` pouvait contenir des objets comme des strings et dans ce cas, l'appel à `np.unique` provoque directement une erreur si l'un des éléments est None. Je me suis dit que ça devrait être géré par cette fonction donc je l'ai étendue, en évitant de donner un array non numérique à `np.isnan` car ça provoque une erreur ...

glide/estimators/cluster_classical.py
imerad: Fait

glide/core/validation.py
gmartinon-ed: Pourquoi pas juste un "if None in array" ?

glide/estimators/cluster_classical.py
gmartinon-ed: Go ahead

tests/unit/estimators/test_cluster_classical.py
```diff
@@ -0,0 +1,117 @@
...
+
+@pytest.fixture
+def estimator() -> ClusterClassicalMeanEstimator:
+    return ClusterClassicalMeanEstimator()
+
+
+# --- _preprocess ---
+
+
+def test_preprocess_delegates_to_validation(estimator, y, clusters):
```
gmartinon-ed, line 30: J'ai l'impression que dans tests/test_classical.py, on n'a pas fait ce test de délégation, et qu'on teste directement que ça raise (test_preprocess_raises_when_fewer_than_two_non_nan_values)

tests/unit/estimators/test_cluster_classical.py
```diff
@@ -0,0 +1,117 @@
...
+        patch.object(cluster_classical_module, "_validate_equal_lengths") as mock_validate_equal_lengths,
+        patch.object(cluster_classical_module, "_validate_has_no_nan") as mock_validate_has_no_nan,
+        patch.object(cluster_classical_module, "_validate_bounds") as mock_validate_bounds,
+    ):
+        estimator._preprocess(y, clusters)
+
+        mock_validate_equal_lengths.assert_called_once()
+        np.testing.assert_array_equal(mock_validate_equal_lengths.call_args[0][0], y)
+        np.testing.assert_array_equal(mock_validate_equal_lengths.call_args[0][1], clusters)
+        assert mock_validate_equal_lengths.call_args[1] == {"names": ["y", "clusters"]}
```
gmartinon-ed, line 42: C'est marrant ce [1], si le mock est appelé une seule fois, ça devrait pas être [0] comme au-dessus ? Instinctivement j'aurais dit : 

```
assert mock_validate_equal_lengths.call_args[0][2] == {"names": ["y", "clusters"]}
```

gmartinon-ed (changes requested):
Review - last part

glide/core/validation.py
imerad: J'ai demandé une explication à Claude, la voici ci-dessous. En résumé, `None in array` fonctionne mais pourrait échouer dans des cas très spéciaux. Je remplace par `None in array`

None in array would use NumPy's __contains__, which calls np.equal(array, None) elementwise. For object arrays containing None, that comparison falls back to Python equality (None == None is True), so it would actually work here.

However, is None is semantically safer: np.equal uses ==, and a custom object could override __eq__ to return True when compared to None. The is check tests identity, not equality, which is the correct way to detect None in Python.

That said, for a NumPy array of plain Python objects (strings, categories), overriding __eq__ is extremely unlikely in practice. You could change it to None in array and it would work correctly for all realistic inputs. The current form is just the more principled choice.

glide/estimators/cluster_classical.py
imerad: J'ai adapté le StratifiedClassical en conséquence, mais c'était trop compliqué pour PTD et PPI stratifiés à cause des calculs de covariance. Je me demande si on n'a pas trop perdu en lisibilité dans StratifiedClassical, je te laisse en juger

tests/unit/estimators/test_cluster_classical.py
imerad: Traité

tests/unit/estimators/test_cluster_classical.py
imerad: Apparemment, `call_args` contient les arguments sous forme d'objet `call`. La première partie `call_args[0]` contient les arguments sous forme de tuple. On vérifie leurs égalités avec les arrays attendus un à un. La deuxième partie `call_args[1]` contient le dictionnaire des keyword arguments et est utilisée ici pour vérifier quel array a été passé en tant que paramètre de quel nom.

glide/estimators/cluster_classical.py
gmartinon-ed: Hum oui, en effet, j'ai l'impression qu'on est tombé sur quelque chose de plus difficile à lire et à comprendre. Sorry c'était une mauvaise intuition de ma part. On peut revert ?

imerad: C'est fait

gmartinon-ed (approved):

PR #256, opened 2026-06-05
Add ClusterPPIMeanEstimator for cluster-sampled PPI

### LINKED ISSUE #236: Implement ClusterPPI
# ClusterPPIMeanEstimator: Prediction-Powered Inference for cluster-sampled data

## Background

Prediction-Powered Inference (PPI++) estimates a population mean by combining a small labeled set with a large set of proxy-predicted labels. The standard `PPIMeanEstimator` treats every observation as independent and applies a bias correction: `θ̂ = mean(y_true) + λ * (mean(y_proxy_unlabeled) - mean(y_proxy_labeled))`. The tuning weight λ is chosen to minimize the width of the confidence interval. The confidence interval is then constructed with the Central Limit Theorem (CLT), treating sample means of individual observations as approximately Gaussian.

When observations are grouped into clusters (for example, all questions from the same conversation thread), the independence assumption breaks down: items from the same cluster share context, phrasing style, and model behavior, so their labels are correlated. Using individual observations as the sampling unit underestimates the true uncertainty. The fix is to replace individual observations with cluster-level summaries. For each cluster k, compute the cluster sum `u_k = n_k * μ_k^true` (for labeled clusters) or `v_k = n_k * μ_k^proxy` (for unlabeled clusters), where n_k is the cluster size. These sums are then the i.i.d. sampling units to which the CLT is applied. In math notation, the point estimate is `θ̂ = Σ_k u_k / N_L + λ * (Σ_k v_k / N_U − Σ_k s_k / N_L)`, where `s_k = n_k * μ_k^proxy` for labeled clusters, `N_L = Σ_k n_k` over labeled clusters, and `N_U = Σ_k n_k` over unlabeled clusters. The CLT variance is `Var(θ̂) = K_L * Var(u_k − λ*s_k, ddof=1) / N_L² + λ² * K_U * Var(v_k, ddof=1) / N_U²`, where K_L and K_U are the numbers of labeled and unlabeled clusters. When all clusters have size 1 (each cluster is a single observation), these formulas reduce exactly to the standard PPI++ formulas, and the two estimators agree when `power_tuning=False`.

The optimal single λ minimises the CLT variance over all labeled and unlabeled cluster sums jointly. Taking the derivative with respect to λ and setting it to zero gives `λ = (K_L * Cov(u_k, s_k, ddof=1) / N_L²) / (K_L * Var(s_k, ddof=1) / N_L² + K_U * Var(v_k, ddof=1) / N_U²)`. This is a single scalar, shared across all clusters, which mirrors the design of `ClusterPTDMeanEstimator` (one tuning parameter from the joint bootstrap) and differs from `StratifiedPPIMeanEstimator` (which computes a separate λ per stratum).

## Design choices

`ClusterPPIMeanEstimator` lives in `glide/estimators/cluster_ppi.py` and mirrors `PPIMeanEstimator` (`glide/estimators/ppi.py`) in structure and API. The `estimate` method takes `clusters` instead of separating labeled from unlabeled, but the remaining signature is identical (metric_name, confidence_level, power_tuning). Like `PPIMeanEstimator`, it uses `CLTConfidenceInterval` (not bootstrap) and returns `PredictionPoweredMeanInferenceResult`. The key difference is that the underlying i.i.d. units are cluster sums, not individual observations.

Three new private helpers — `_compute_cluster_tuning_parameter`, `_compute_cluster_mean_estimate`, and `_compute_cluster_std_estimate` — are added to `glide/estimators/cluster_core.py`. They are the cluster-level analogues of `ppi_core.py`'s three helpers. The functions in `ppi_core.py` cannot be reused directly because their variance formulas divide by the number of individual observations rather than using the `K * Var(cluster_sum, ddof=1) / N²` formula required for cluster-level CLT. Adding the new helpers to `cluster_core.py` keeps all cluster-specific math in one place, alongside `_preprocess` and `_compute_bootstrap_labeled_cluster_means` already established there.

Preprocessing is handled entirely by `_preprocess` from `cluster_core.py` (established in the ClusterPTD ticket). The same strict requirement applies: every cluster must be either fully labeled (all `y_true` entries finite) or fully unlabeled (all `y_true` entries are `np.nan`). Partially labeled clusters are rejected with a `ValueError`. A single λ is computed from all labeled and unlabeled cluster sums together; there is no per-cluster λ. The `n_true` and `n_proxy` fields in the result count individual observations (not clusters), consistent with `ClusterPTDMeanEstimator` and `PPIMeanEstimator`.

### New helpers in `glide/estimators/cluster_core.py`

**Purpose:** Cluster-level analogues of `ppi_core.py`'s `_compute_tuning_parameter`, `_compute_mean_estimate`, and `_compute_std_estimate`, working on cluster sums instead of individual observations.

```python
def _compute_cluster_tuning_parameter(
    labeled_true_sums: NDArray,
    labeled_proxy_sums: NDArray,
    unlabeled_proxy_sums: NDArray,
    labeled_total_size: int,
    unlabeled_total_size: int,
    power_tuning: bool,
) -> float:
    if not power_tuning:
        return 1.0
    n_labeled_clusters = len(labeled_true_sums)
    n_unlabeled_clusters = len(unlabeled_proxy_sums)
    cov_tu = np.cov(labeled_true_sums, labeled_proxy_sums, ddof=1)[0, 1]
    var_labeled_proxy = np.var(labeled_proxy_sums, ddof=1)
    var_unlabeled_proxy = np.var(unlabeled_proxy_sums, ddof=1)
    numerator = n_labeled_clusters * cov_tu / labeled_total_size**2
    denominator = (
        n_labeled_clusters * var_labeled_proxy / labeled_total_size**2
        + n_unlabeled_clusters * var_unlabeled_proxy / unlabeled_total_size**2
    )
    if denominator == 0.0:
        raise ValueError(
            "Proxy cluster sums have zero variance across both labeled and unlabeled clusters; "
            "cannot estimate the tuning parameter."
        )
    lambda_ = numerator / denominator
    return lambda_


def _compute_cluster_mean_estimate(
    labeled_true_sums: NDArray,
    labeled_proxy_sums: NDArray,
    unlabeled_proxy_sums: NDArray,
    labeled_total_size: int,
    unlabeled_total_size: int,
    lambda_: float,
) -> float:
    labeled_true_mean = np.sum(labeled_true_sums) / labeled_total_size
    labeled_proxy_mean = np.sum(labeled_proxy_sums) / labeled_total_size
    unlabeled_proxy_mean = np.sum(unlabeled_proxy_sums) / unlabeled_total_size
    rectifier = labeled_true_mean - lambda_ * labeled_proxy_mean
    mean_estimate = lambda_ * unlabeled_proxy_mean + rectifier 
    return mean_estimate


def _compute_cluster_std_estimate(
    labeled_true_sums: NDArray,
    labeled_proxy_sums: NDArray,
    unlabeled_proxy_sums: NDArray,
    labeled_total_size: int,
    unlabeled_total_size: int,
    lambda_: float,
) -> float:
    n_labeled_clusters = len(labeled_true_sums)
    n_unlabeled_clusters = len(unlabeled_proxy_sums)
    rectifier_sums = labeled_true_sums - lambda_ * labeled_proxy_sums
    labeled_var = n_labeled_clusters * np.var(rectifier_sums, ddof=1) / labeled_total_size**2
    unlabeled_var = lambda_**2 * n_unlabeled_clusters * np.var(unlabeled_proxy_sums, ddof=1) / unlabeled_total_size**2
    total_var = labeled_var + unlabeled_var
    std_estimate = np.sqrt(total_var)
    return std_estimate
```

### `ClusterPPIMeanEstimator`

**Location:** `glide/estimators/cluster_ppi.py`

**Purpose:** Public estimator class that wires cluster preprocessing, the single tuning parameter, and CLT-based confidence interval assembly.

```python
from typing import Optional

import numpy as np

from glide.estimators.cluster_core import (
    _compute_cluster_mean_estimate,
    _compute_cluster_std_estimate,
    _compute_cluster_tuning_parameter,
    _preprocess,
)
from glide.mean_inference_results import PredictionPoweredMeanInferenceResult
from glide.utils import compute_effective_sample_size


class ClusterPPIMeanEstimator:
    """Cluster PPI++ estimator for population mean.

    Extends Prediction-Powered Inference to datasets where observations are
    grouped into clusters and each cluster is either entirely labeled or
    entirely unlabeled. The CLT is applied to cluster sums as the sampling
    units, which accounts for within-cluster correlation and produces valid
    confidence intervals under cluster sampling designs.

    A single power-tuning parameter λ is estimated from the joint covariance
    of labeled true sums and labeled proxy sums. The final confidence interval
    is CLT-based and centred on:

        θ̂ = Σ_k u_k / N_L + λ * (Σ_k v_k / N_U − Σ_k s_k / N_L)

    where u_k = n_k * μ_k^true and s_k = n_k * μ_k^proxy are cluster sums
    for labeled clusters, v_k = n_k * μ_k^proxy are cluster sums for
    unlabeled clusters, and N_L, N_U are the total labeled and unlabeled
    observation counts.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.estimators import ClusterPPIMeanEstimator
    >>> y_true = np.array([1.0, 2.0, 3.0, 4.0, np.nan, np.nan, np.nan, np.nan])
    >>> y_proxy = np.array([1.1, 2.2, 3.1, 3.9, 1.5, 1.8, 4.5, 4.8])
    >>> clusters = np.array(["A", "A", "B", "B", "C", "C", "D", "D"])
    >>> estimator = ClusterPPIMeanEstimator()
    >>> result = estimator.estimate(y_true, y_proxy, clusters)
    >>> print(result)  # fill in exact output after running the implementation
    Metric: Metric
    Point Estimate: X.XXX
    Confidence Interval (95%): [X.XXX, X.XXX]
    Estimator : ClusterPPIMeanEstimator
    n_true: 4
    n_proxy: 8
    Effective Sample Size: X
    """

    def estimate(
        self,
        y_true: NDArray,
        y_proxy: NDArray,
        clusters: NDArray,
        metric_name: str = "Metric",
        power_tuning: bool = True,
    ) -> PredictionPoweredMeanInferenceResult:
        """Estimate the population mean using the Cluster PPI++ estimator.

        Aggregates observations into cluster sums, computes a single
        power-tuning parameter λ from the joint covariance of labeled
        cluster sums, and assembles a CLT-based confidence interval.

        The point estimate is:

            θ̂ = Σ_k u_k / N_L + λ * (Σ_k v_k / N_U − Σ_k s_k / N_L)

        and the variance is:

            Var(θ̂) = K_L * Var(u_k − λ*s_k, ddof=1) / N_L²
                    + λ² * K_U * Var(v_k, ddof=1) / N_U²

        where K_L and K_U are the numbers of labeled and unlabeled clusters.

        Labeled and unlabeled clusters are distinguished by the NaN pattern in
        ``y_true``: a cluster is labeled if every one of its ``y_true`` entries
        is finite, and unlabeled if every entry is ``np.nan``. Partially labeled
        clusters are not supported.

        Parameters
        ----------
        y_true : NDArray
            Array of observations, shape ``(n_samples,)``.
            All observations in the same cluster must share the same label status.
        y_proxy : NDArray
        clusters : NDArray
        metric_name : str, optional
        power_tuning : bool, optional
            If ``True`` (default), estimate the optimal power-tuning parameter
            λ from the cluster-level covariances. If ``False``, use λ = 1.0.

        Returns
        -------
        PredictionPoweredMeanInferenceResult
            Contains the CLT-based confidence interval, the metric name,
            the estimator name (``"ClusterPPIMeanEstimator"``), and the counts
            ``n_true`` (total labeled observations) and ``n_proxy`` (total
            dataset size).

        Raises
        ------
        ValueError
            - If ``y_true``, ``y_proxy``, and ``clusters`` do not all have the
              same length.
            - If any proxy value is NaN.
            - If any cluster contains both labeled and unlabeled observations.
            - If fewer than 2 clusters are fully labeled.
            - If fewer than 2 clusters are fully unlabeled.
            - If proxy cluster sums have zero variance across both labeled and
              unlabeled clusters and ``power_tuning=True``.
        """
        (
            labeled_y_true_means,
            labeled_y_proxy_means,
            unlabeled_y_proxy_means,
            labeled_sizes,
            unlabeled_sizes,
        ) = _preprocess(y_true, y_proxy, clusters)

        labeled_true_sums = labeled_sizes * labeled_y_true_means
        labeled_proxy_sums = labeled_sizes * labeled_y_proxy_means
        unlabeled_proxy_sums = unlabeled_sizes * unlabeled_y_proxy_means
        labeled_total_size = np.sum(labeled_sizes)
        unlabeled_total_size = np.sum(unlabeled_sizes)

        lambda_ = _compute_cluster_tuning_parameter(
            labeled_true_sums,
            labeled_proxy_sums,
            unlabeled_proxy_sums,
            labeled_total_size,
            unlabeled_total_size,
            power_tuning,
        )
        mean = _compute_cluster_mean_estimate(
            labeled_true_sums,
            labeled_proxy_sums,
            unlabeled_proxy_sums,
            labeled_total_size,
            unlabeled_total_size,
            lambda_,
        )
        std = _compute_cluster_std_estimate(
            labeled_true_sums,
            labeled_proxy_sums,
            unlabeled_proxy_sums,
            labeled_total_size,
            unlabeled_total_size,
            lambda_,
        )
        confidence_interval = CLTConfidenceInterval(
            mean=mean,
            std=std,
        )
        effective_sample_size = compute_effective_sample_size(y_true_labeled, confidence_interval.var)
        result = PredictionPoweredMeanInferenceResult(
            confidence_interval=confidence_interval,
            metric_name=metric_name,
            n_true=len(y_true_labeled),
            n_proxy=len(y_true),
            effective_sample_size=effective_sample_size,
        )
        return result
```

Add the following import (keep alphabetical order within each group):

```python
from glide.estimators.cluster_ppi import ClusterPPIMeanEstimator
```

Add `"ClusterPPIMeanEstimator"` to `__all__`.

### Doctest values

After implementing, run the class-level doctest example and replace the `X.XXX` placeholders with the actual output. For the method-level doctest, `result.n_true == 4` and `result.n_proxy == 8` are exact and can be verified analytically from the input arrays.

## Prerequisite: fix `ppi_core._compute_tuning_parameter` to use the exact optimal λ

`ppi_core._compute_tuning_parameter` currently estimates the variance-minimising λ using the pooled proxy variance:

```python
y_proxy_all = np.hstack([y_proxy_labeled, y_proxy_unlabeled])
var = np.var(y_proxy_all, ddof=1)
lambda_ = cov / ((1 + n_labeled / n_unlabeled) * var)
```

The exact optimal λ is obtained by taking the derivative of `Var(θ̂)` with respect to λ and setting it to zero:

```
λ* = Cov(y_true, y_proxy_labeled) / (Var(y_proxy_labeled) + n_L/n_U * Var(y_proxy_unlabeled))
```

The pooled form `(1 + n_L/n_U) * Var(y_proxy_all)` equals the exact denominator only when `Var(y_proxy_labeled) = Var(y_proxy_unlabeled) = Var(y_proxy_all)`, which holds in the population but not in finite samples. As a result, `_compute_cluster_tuning_parameter` (which implements the exact formula at the cluster level) cannot reduce to `ppi_core._compute_tuning_parameter` for single-observation clusters, breaking the expected equivalence between the two estimators.

Replacing the pooled form with the exact denominator restores this equivalence. The validation in `PPIMeanEstimator._preprocess` (`len(np.unique(y_proxy_all)) == 1` raises `ValueError`) already covers the only case where the exact denominator can be zero (all proxy values identical implies both variances are zero), so no new error handling is needed.

**IMPORTANT : We have two options:**
1. We modify the PPI power tuning formula implemented to match the current proposition
2. We modify the ClusterPPI power tuning formula to align with the existing PPI one

In order to identify which option is best, we need to first understand how the two formulas have been obtained, in particular, which minimization problem lambda solves : is it a minimizer of the asymptotic variance of the estimator, or is it a minimizer of the empirical variance of the estimator. Guts feeling: Claude, when designing the ticket, forgot to consider that both proxy samples (labelde and unlabeled) come from exactly the same population, and thus their variance should be pooled.

**Updated `ppi_core._compute_tuning_parameter`:**

```python
def _compute_tuning_parameter(
    y_true: NDArray,
    y_proxy_labeled: NDArray,
    y_proxy_unlabeled: NDArray,
    power_tuning: bool,
) -> float:
    if not power_tuning:
        return 1.0
    n_labeled = len(y_true)
    n_unlabeled = len(y_proxy_unlabeled)
    cov = np.cov(y_true, y_proxy_labeled, ddof=1)[0, 1]
    var_labeled = np.var(y_proxy_labeled, ddof=1)
    var_unlabeled = np.var(y_proxy_unlabeled, ddof=1)
    denominator = var_labeled + (n_labeled / n_unlabeled) * var_unlabeled
    lambda_ = cov / denominator
    return lambda_
```

**Impact sites — hardcoded expected values that must be updated:**

`glide/estimators/ppi.py` doctest (fixed arrays `y_true=[5,6,nan,nan]`, `y_proxy=[4.9,6.1,5.2,6.1]`):

| Field | Before | After |
|---|---|---|
| Point Estimate | 5.618 | 5.580 |
| CI (95%) | [4.923, 6.312] | [4.992, 6.168] |
| Effective Sample Size | 3 | 5 |

`glide/estimators/stratified_ppi.py` doctest (fixed arrays with `groups=[0,0,0,0,1,1,1,1]`):

| Field | Before | After |
|---|---|---|
| Point Estimate | 3.086 | 3.059 |
| CI (95%) | [2.720, 3.452] | [2.884, 3.234] |
| Effective Sample Size | 95 | 416 |

`tests/unit/estimators/test_stratified_ppi.py` (fixed arrays with `groups=["A","A","A","A","B","B","B","B"]`):

| Test | Field | Before | After |
|---|---|---|---|
| `test_estimate_metadata` | `effective_sample_size` | 5 | 7 |
| `test_estimate_custom_confidence_level` | `expected_mean` | 5.618 | 5.580 |
| `test_estimate_custom_confidence_level` | `expected_std` | 0.250 | 0.212 |
| `test_estimate_custom_confidence_level` | `expected_lower` | 5.257 | 5.275 |
| `test_estimate_custom_confidence_level` | `expected_upper` | 5.978 | 5.885 |
| `test_str_format` | Point Estimate | 5.618 | 5.580 |
| `test_str_format` | CI (95%) | [5.127, 6.108] | [5.164, 5.996] |
| `test_str_format` | Effective Sample Size | 5 | 7 |

`tests/unit/estimators/test_ppi.py`: uses randomly generated data (seed=42) for `test_str_format` and `test_estimate_custom_confidence_level`. Re-run after the change and update the hardcoded values in those tests.

## Functional tests

**Single-observation clusters produce identical results to PPIMeanEstimator**

When every cluster contains exactly one observation, `u_k = y_true_k`, `s_k = y_proxy_labeled_k`, `v_k = y_proxy_unlabeled_k`, and `K_L = N_L`, `K_U = N_U`. The cluster variance formula `K_L * Var(u_k − λ*s_k, ddof=1) / N_L²` reduces to `Var(y_true − λ*y_proxy_labeled, ddof=1) / n_L`, and `λ² * K_U * Var(v_k, ddof=1) / N_U²` reduces to `λ² * Var(y_proxy_unlabeled, ddof=1) / n_U` — identical to `ppi_core._compute_std_estimate`. Once the prerequisite fix to `ppi_core._compute_tuning_parameter` is in place, `_compute_cluster_tuning_parameter` reduces to the same formula for single-observation clusters, so the equivalence holds for both `power_tuning=False` and `power_tuning=True`.

```python
@pytest.mark.parametrize("power_tuning", [False, True])
def test_single_observation_clusters_equals_ppi(power_tuning):
    y_true = np.array([4.0, np.nan, 6.0, np.nan])
    y_proxy = np.array([2.0, 4.0, 6.0, 6.0])
    clusters = np.array(["A", "C", "B", "D"])

    cluster_result = ClusterPPIMeanEstimator().estimate(
        y_true, y_proxy, clusters, power_tuning=power_tuning
    )
    ppi_result = PPIMeanEstimator().estimate(
        y_true, y_proxy, power_tuning=power_tuning
    )
        ppi_result.confidence_interval.lower_bound, abs=1e-10
    )
        ppi_result.confidence_interval.upper_bound, abs=1e-10
    )
```

## Corner cases

- A cluster with both labeled and unlabeled observations (e.g., `y_true = [5.0, np.nan]` for cluster "A") must raise `ValueError` with the cluster ID in the message.
- Exactly 1 labeled cluster must raise `ValueError` because `np.var(labeled_true_sums, ddof=1)` is undefined with a single value.
- Exactly 1 unlabeled cluster must raise `ValueError` for the same reason.
- `power_tuning=True` when all proxy cluster sums are identical across both labeled and unlabeled clusters (denominator = 0) must raise `ValueError`; `power_tuning=False` must succeed because λ is hardcoded to 1.0 and no denominator is computed.
- A cluster where all observations have `np.nan` in `y_proxy` is caught by the global NaN-in-proxy check inside `_preprocess` before cluster iteration.

## Analytical verification

The following values are derivable by hand and should be used in unit tests. Given:

```python
y_true = np.array([4.0, np.nan, 6.0, np.nan])
y_proxy = np.array([2.0, 4.0, 6.0, 6.0])
clusters = np.array(["A", "C", "B", "D"])
```

After preprocessing: labeled clusters A (u=4, s=2) and B (u=6, s=6), unlabeled clusters C (v=4) and D (v=6). K_L=K_U=2, N_L=N_U=2.

**With power_tuning=False (λ=1):**
- `labeled_true_mean = (4+6)/2 = 5`, `labeled_proxy_mean = (2+6)/2 = 4`, `unlabeled_proxy_mean = (4+6)/2 = 5`
- `mean = 5 + 1*(5-4) = 6`
- `Var(rectifier_k=[2,0], ddof=1) = 2` → `labeled_var = 2*2/4 = 1`
- `Var(v_k=[4,6], ddof=1) = 2` → `unlabeled_var = 1²*2*2/4 = 1`
- `std = sqrt(2)`

**With power_tuning=True:**
- `Cov([4,6],[2,6], ddof=1) = 4`, `Var([2,6], ddof=1) = 8`, `Var([4,6], ddof=1) = 2`
- `numerator = 2*4/4 = 2`, `denominator = 2*8/4 + 2*2/4 = 4+1 = 5`
- `λ = 2/5 = 0.4`
- `mean = 5 + 0.4*(5-4) = 5.4`
- `rectifier_k = [4-0.8, 6-2.4] = [3.2, 3.6]` → `Var = 0.08`, `labeled_var = 2*0.08/4 = 0.04`
- `unlabeled_var = 0.4²*2*2/4 = 0.16`
- `std = sqrt(0.2)`

These values must appear verbatim as `pytest.approx` assertions in `tests/unit/estimators/test_cluster_core.py` (for the helpers) and `tests/unit/estimators/test_cluster_ppi.py` (for the estimator).

## References

- Broska, David. Cluster-robust PPI reference implementation. https://github.com/davidbroska/ppi_py/blob/main/ClusterPPI/mean.py

## Acceptance criteria

- [ ] `ppi_core._compute_tuning_parameter` uses the exact denominator `Var(y_proxy_labeled) + n_L/n_U * Var(y_proxy_unlabeled)` instead of the pooled form; all affected hardcoded expected values in `test_ppi.py`, `test_stratified_ppi.py`, and the `ppi.py` / `stratified_ppi.py` doctests are updated to match.
- [ ] `ClusterPPIMeanEstimator` is importable from `glide.estimators` and listed in `__all__`.
- [ ] `_preprocess` raises `ValueError` with the cluster ID when a cluster contains both labeled and unlabeled observations (reused from ClusterPTD — no new code needed, but must be covered by tests in `test_cluster_ppi.py`).
- [ ] `_compute_cluster_tuning_parameter` raises `ValueError` when proxy cluster sums have zero variance in both labeled and unlabeled clusters and `power_tuning=True`.
- [ ] With `power_tuning=False`, `confidence_interval.mean == 6.0` and `std == sqrt(2)` to within `1e-10` for the analytical example above.
- [ ] With `power_tuning=True`, `lambda_ == 0.4`, `confidence_interval.mean == 5.4`, and `std == sqrt(0.2)` to within `1e-10` for the analytical example above.
- [ ] `result.n_true` equals the total number of labeled individual observations (not clusters); `result.n_proxy` equals the total number of observations.
- [ ] All class-level and method-level doctests pass (exact values filled in after running the implementation).
- [ ] `cluster_core.py` is extended with exactly three new helpers: `_compute_cluster_tuning_parameter`, `_compute_cluster_mean_estimate`, and `_compute_cluster_std_estimate`.
- [ ] `tests/unit/estimators/test_cluster_core.py` is extended with unit tests for the three new helpers, and `tests/unit/estimators/test_cluster_ppi.py` exists and mirrors the structure of `test_stratified_ppi.py`.
- [ ] The estimator table in `README.md` is updated to include `ClusterPPIMeanEstimator`.

imerad:
### Description

- Adds `ClusterPPIMeanEstimator`, a first (unrevised) implementation of PPI++ for cluster-sampled data. Cluster sums replace individual observations as the CLT sampling units, accounting for within-cluster correlation.
- Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=171055645&issue=EmertonData%7Cglide%7C236) Closes #236

No change has been brought to the original method for computing power tuning in PPI

The suggested implementations in the ticket has been modified to align with the existing implementation of `ClusterClassicalMeanEstimator`

Some updates (nan checks) have been added to stratified_core.py code on the way

Clustered PPI has been added to the list of implemented papers on the ReadME, but it actually has no associated paper. SHould we change how we list our implemented algorithms ?

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
## Code Review — `ClusterPPIMeanEstimator`

Overall structure is clean and consistent with the rest of the codebase. Four issues below need to be addressed before merge; the remaining items are lower-priority observations.

---

### 🔴 Critical — incorrect tuning parameter formula

**[`glide/estimators/cluster_core.py:90`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_core.py#L90)**

```python
# current (wrong)
factor = 1 + (n_unlabeled_clusters / n_labeled_clusters) / (labeled_total_size / unlabeled_total_size) ** 2

# correct
factor = 1 + (n_unlabeled_clusters / n_labeled_clusters) * (labeled_total_size / unlabeled_total_size) ** 2
```

The optimal λ minimises total variance `K_L·Var(u_k−λ·s_k)/N_L² + λ²·K_U·Var(v_k)/N_U²`. Setting the derivative to zero (under the pooled-variance approximation) gives:

```
λ* = Cov(u_k, s_k) / [var_proxy · (1 + (K_U/K_L)·(N_L/N_U)²)]
```

The code computes `1 + (K_U/K_L)·(N_U/N_L)²` instead — the ratio is inverted. When the unlabeled set is larger than the labeled set (the typical case), the denominator becomes orders of magnitude too large, driving λ toward zero and nearly zeroing out the proxy correction. With K_L=2, K_U=20, N_L=20, N_U=100 the code gives factor≈251 (λ≈0) versus the correct factor≈1.4 (λ≈0.76).

**The existing tests do not catch this bug.** Every fixture uses `labeled_total_size = unlabeled_total_size`, which makes `(N_L/N_U)² = (N_U/N_L)² = 1`, rendering both formulas identical. A test with unequal population sizes is needed in `test_compute_cluster_tuning_parameter_known_value`.

---

### 🔴 High — `ZeroDivisionError` / `ValueError` crash in ESS computation

**[`glide/estimators/cluster_ppi.py:176`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_ppi.py#L176)**

```python
effective_sample_size = floor(labeled_total_size * classical_confidence_interval.var / confidence_interval.var)
```

When `power_tuning=False` and all labeled cluster sums happen to be equal (e.g. every labeled cluster has the same true sum), `_compute_cluster_std_estimate` returns `0.0`, so `confidence_interval.var == 0`. The expression then evaluates to `float('nan')`, and `floor(nan)` raises `ValueError: cannot convert float NaN to integer`. The `power_tuning=True` path is guarded by `_validate_non_constant` on the proxy sums, but there is no equivalent guard for the labeled true sums.

---

### 🟠 High — circular import fragility

**[`glide/estimators/cluster_ppi.py:7`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_ppi.py#L7)**

```python
from glide.estimators import ClusterClassicalMeanEstimator   # imports the package __init__
```

`glide/estimators/__init__.py` imports `cluster_ppi` on line 4. When Python processes that import it starts executing `cluster_ppi.py`, which then tries to import from `glide.estimators` — the same partially-initialised package. This works today only because `ClusterClassicalMeanEstimator` is registered on line 3 of `__init__.py`, before `cluster_ppi` is reached on line 4. Reordering or inserting an import ahead of line 3 will produce `ImportError: cannot import name 'ClusterClassicalMeanEstimator'`.

Every other estimator that depends on a sibling class uses the direct submodule import (e.g. `ppi.py` line 14: `from glide.estimators.classical import ClassicalMeanEstimator`). Use the same pattern here:

```python
from glide.estimators.cluster_classical import ClusterClassicalMeanEstimator
```

---

### 🟡 Medium — `_validate_non_constant` guards the wrong scope

**[`glide/estimators/cluster_core.py:83–88`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_core.py#L83)**

```python
all_proxy_sums = np.hstack([labeled_proxy_sums, unlabeled_proxy_sums])
_validate_non_constant(all_proxy_sums, ...)          # guards the pool
cov = np.cov(labeled_true_sums, labeled_proxy_sums, ddof=1)[0, 1]  # uses only labeled
```

The guard fires only when the entire pool is constant. If `labeled_proxy_sums` is constant (e.g. all labeled cluster proxy sums equal 5.0) but the unlabeled sums vary, `all_proxy_sums` passes the check while `np.cov(labeled_true_sums, labeled_proxy_sums)` returns `NaN`. The result is λ = `NaN/denominator = NaN`, propagating silently into the mean and std estimates with no diagnostic. A separate `_validate_non_constant(labeled_proxy_sums, ...)` call before computing `cov` would close this gap.

---

### 🟡 Medium — redundant full re-estimation for ESS denominator

**[`glide/estimators/cluster_ppi.py:175`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_ppi.py#L175)**

```python
classical_confidence_interval = ClusterClassicalMeanEstimator().estimate(y_true, clusters).confidence_interval
```

This re-runs the full classical pipeline — NaN scan, `np.unique`, five `np.bincount` calls, validation — purely to read `.var`. Everything needed is already in scope from the earlier `_preprocess` call:

```python
classical_var = n_labeled_clusters * np.var(labeled_true_sums, ddof=1) / labeled_total_size**2
```

(`n_labeled_clusters = len(labeled_true_sums)` from the sums returned by `_preprocess`.)

This also removes a hidden coupling: if `ClusterClassicalMeanEstimator._preprocess` ever diverges in its NaN-handling or cluster-counting rules, the ESS denominator would silently use a different set of clusters than the PPI numerator, producing an inconsistent result.

---

### ⚪ Low — docstring formula notation inconsistency

**[`glide/estimators/cluster_ppi.py:31`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_ppi.py#L31)**

The class docstring writes the PPI estimator as:

```
θ̂ = Σ_k u_k / N_L + λ * (Σ_k v_k / N_U - Σ_k s_k / N_L)
```

The `estimate` method docstring repeats the same formula. Both are correct but note they describe the **mean of cluster sums** (divided by total observation count), not the mean of per-observation values — worth making explicit that `N_L` and `N_U` are observation counts, not cluster counts, since the distinction matters when cluster sizes are unequal.

---

### ⚪ Low — `n_labeled_clusters` / `n_unlabeled_clusters` computed twice

**[`glide/estimators/cluster_core.py:45–46`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_core.py#L45)**

`n_labeled_clusters = len(labeled_true_sums)` and `n_unlabeled_clusters = len(unlabeled_proxy_sums)` are derived quantities that are already available as `len(unique_labeled_clusters)` and `len(unique_unlabeled_clusters)` on lines 27–28. Minor but introduces two redundant names for the same values.

---

### ⚪ Low — `assume_unique=True` is a future maintenance trap

**[`glide/estimators/cluster_core.py:30`](https://github.com/EmertonData/glide/blob/feat/cluster_ppi/glide/estimators/cluster_core.py#L30)**

```python
intersection = np.intersect1d(unique_labeled_clusters, unique_unlabeled_clusters, assume_unique=True)
```

The inputs are outputs of `np.unique` so `assume_unique=True` is currently correct. However it provides no performance benefit on sorted arrays and creates a silent-correctness trap: if the `np.unique` calls above are ever removed or moved, `np.intersect1d` with `assume_unique=True` will silently return a wrong (incomplete) intersection, allowing mixed clusters to pass validation undetected. Drop the flag.

README.md
```diff
@@ -83,6 +83,7 @@ If you use GLIDE in your work, please cite us using the "Cite this repository" b
 | 2024 | [Active Statistical Inference](https://dl.acm.org/doi/10.5555/3692070.3694680) |ICML'24|[Link](https://github.com/tijana-zrnic/active-inference)| samplers.ActiveSampler, estimators.ASIMeanEstimator |
 | 2025 | [Can Unconfident LLM Annotations Be Used for Confident Conclusions?](https://aclanthology.org/2025.naacl-long.179/) |NAACL'25|[Link](https://github.com/kristinagligoric/confidence-driven-inference)| samplers.ActiveSampler, estimators.ASIMeanEstimator |
 | 2025 | [Prediction-Powered Inference with Imputed Covariates and Nonuniform Sampling](https://arxiv.org/abs/2501.18577) |Preprint|[Link](https://github.com/DanKluger/PTDBoot)| estimators.PTDMeanEstimator, estimators.StratifiedPTDMeanEstimator, estimators.IPWPTDMeanEstimator |
+| 2025 | [Cluster-Robust Prediction-Powered Inference](https://github.com/davidbroska/ppi_py/blob/main/ClusterPPI/mean.py) |Preprint|[Link](https://github.com/davidbroska/ppi_py)| estimators.ClusterPPIMeanEstimator |
```
gmartinon-ed: Il n'y a pas d'articles publiés sur ce sujet à ma connaissance

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+
+from glide.core.validation import (
+    _validate_bounds,
+    _validate_equal_lengths,
+    _validate_has_no_nan,
+    _validate_non_constant,
+)
+
+
+def _preprocess(
```
gmartinon-ed: Tiens pourquoi ne pas avoir laissé `_preprocess` comme méthode de la classe ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    unique_unlabeled_clusters, unlabeled_cluster_indices = np.unique(unlabeled_clusters, return_inverse=True)
+
+    intersection = np.intersect1d(unique_labeled_clusters, unique_unlabeled_clusters, assume_unique=True)
+    bad_cluster = intersection[0] if len(intersection) else None
+    _validate_bounds(
+        len(intersection),
+        "clusters_intersection",
+        upper=0,
+        error_message=f"Cluster '{bad_cluster}' contains both labeled and unlabeled observations.",
+    )
```
gmartinon-ed: Est-ce que tout ça n'est pas juste une fonction de validation `_validate_clusters` ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    if not power_tuning:
+        return 1.0
+    n_labeled_clusters = len(labeled_true_sums)
+    n_unlabeled_clusters = len(unlabeled_proxy_sums)
+    all_proxy_sums = np.hstack([labeled_proxy_sums, unlabeled_proxy_sums])
+    _validate_non_constant(
+        all_proxy_sums,
+        "Proxy cluster sums have zero variance across both labeled and unlabeled clusters; "
+        "cannot estimate the tuning parameter.",
+    )
```
gmartinon-ed, line 24: Je ne vois pas d'équivalent dans PPI. Est-ce que ce check ne devrait pas être fait dans `_preprocess` ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    cov = np.cov(labeled_true_sums, labeled_proxy_sums, ddof=1)[0, 1]
+    var_proxy = np.var(all_proxy_sums, ddof=1)
+    factor = 1 + (n_unlabeled_clusters / n_labeled_clusters) * (labeled_total_size / unlabeled_total_size) ** 2
+    lambda_ = cov / (var_proxy * factor)
```
gmartinon-ed, line 28: J'aime bien cette variable intermédiaire `factor` -> possible d'harmoniser dans ppi_core ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    return lambda_
```
gmartinon-ed, line 29: On avait eu un débat pendant le refinement sur cette formule : 

"IMPORTANT : We have two options:

We modify the PPI power tuning formula implemented to match the current proposition
We modify the ClusterPPI power tuning formula to align with the existing PPI one
In order to identify which option is best, we need to first understand how the two formulas have been obtained, in particular, which minimization problem lambda solves : is it a minimizer of the asymptotic variance of the estimator, or is it a minimizer of the empirical variance of the estimator. Guts feeling: Claude, when designing the ticket, forgot to consider that both proxy samples (labelde and unlabeled) come from exactly the same population, and thus their variance should be pooled.
"

-> Qu'en est-il et quel est le rationnel derrière cette proposition d'implémentation ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    labeled_total_size: int,
+    unlabeled_total_size: int,
+    lambda_: float,
+) -> float:
+    labeled_true_mean = np.sum(labeled_true_sums) / labeled_total_size
+    labeled_proxy_mean = np.sum(labeled_proxy_sums) / labeled_total_size
+    unlabeled_proxy_mean = np.sum(unlabeled_proxy_sums) / unlabeled_total_size
+    rectifier = labeled_true_mean - lambda_ * labeled_proxy_mean
+    mean_estimate = lambda_ * unlabeled_proxy_mean + rectifier
+    return mean_estimate
```
gmartinon-ed, line 45: J'ai l'impression que ce truc est complètement équivalent à `_compute_mean_estimate` de ppi_core. Veut-on le garder du coup ?

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    labeled_proxy_sums: NDArray,
+    unlabeled_proxy_sums: NDArray,
+    labeled_total_size: int,
+    unlabeled_total_size: int,
+    lambda_: float,
+) -> float:
+    n_labeled_clusters = len(labeled_true_sums)
+    n_unlabeled_clusters = len(unlabeled_proxy_sums)
+    rectifier_sums = labeled_true_sums - lambda_ * labeled_proxy_sums
+    labeled_var = n_labeled_clusters * np.var(rectifier_sums, ddof=1) / labeled_total_size**2
```
gmartinon-ed: rectifier_var ? 
(cf notation dans ppi_core)

glide/estimators/cluster_core.py
```diff
@@ -0,0 +1,126 @@
...
+    unlabeled_var = lambda_**2 * n_unlabeled_clusters * np.var(unlabeled_proxy_sums, ddof=1) / unlabeled_total_size**2
```
gmartinon-ed: proxy_var ? (ppi_core)

tests/unit/estimators/test_cluster_core.py
```diff
@@ -0,0 +1,154 @@
...
+                ),
+                call(2, "n_labeled_clusters", lower=2, error_message="Need at least 2 fully labeled clusters; got 2."),
+                call(
+                    2, "n_unlabeled_clusters", lower=2, error_message="Need at least 2 fully unlabeled clusters; got 2."
+                ),
+            ]
+        )
+
+
+def test_preprocess_valid_output(
```
gmartinon-ed: Oui, cet ordre de test est bien : d'abord on vérifie la délégation, ensuite on vérifie les outputs -> à propager pour PPI

tests/unit/estimators/test_stratified_core.py
```diff
@@ -1,38 +1,53 @@
...
     strata = _preprocess(y_true, y_proxy, groups)
+
     assert len(strata) == 2
     for y_true_filtered, y_proxy_labeled, y_proxy_unlabeled in strata:
-        assert len(y_true_filtered) == 2
-        assert len(y_proxy_labeled) == 2
-        assert len(y_proxy_unlabeled) == 2
+        np.testing.assert_array_equal(y_true_filtered, np.array([5.0, 6.0]))
+        np.testing.assert_array_equal(y_proxy_labeled, np.array([4.9, 6.1]))
+        np.testing.assert_array_equal(y_proxy_unlabeled, np.array([5.2, 6.1]))
```
gmartinon-ed: Bien. C'est Claude qui l'a suggéré ? Est-ce qu'on a déjà loggé dans CLAIUDE.md qu'on préférait tester, quand c'est possible, des valeurs en durs plutôt que des agrégats, avec de genre d'exemple ?

gmartinon-ed (changes requested):

README.md
imerad: J'enlève cette ligne, mais c'est dommage de ne pas inclure ClusterPPI ici. On pourrait faire un tableau "implemented algorithms" et avoir une colonne "Reference papers" ?

glide/estimators/cluster_core.py
imerad: J'ai créé une fonction `_validate_labeled_unlabeled_clusters` dans validation.py pour factoriser ces lignes (30-37).

glide/estimators/cluster_core.py
imerad: Rajouté dans ppi_core.py. Ce test n'a pas besoin de se faire si `power_tuning=False` donc je pense qu'il peut rester là

glide/estimators/cluster_core.py
imerad: Fait

glide/estimators/cluster_core.py
imerad: Traité

glide/estimators/cluster_core.py
imerad: Traité

tests/unit/estimators/test_cluster_core.py
imerad: Fait

tests/unit/estimators/test_stratified_core.py
imerad: Pour le coup ce n'était pas Claude, mais plutôt un souci d'homogénéisation entre le traitement des clusters et des strates (autant que possible). Je demande à Claude de rajouter une instruction à ce propos au claude.md

glide/estimators/cluster_core.py
imerad: Pas tout à fait, cette fonction a ces deux paramètres en plus `labeled_total_size` et `unlabeled_total_size`. On pourrait néanmoins la réécrire avec ces paramètres en options. Ils auraient pour valeurs par défaut les tailles de `labeled_true_sums` et `labeled_proxy_sums`. Mais dans ce cas, je pense qu'il serait plus cohérent de le faire pour toutes les fonctions `_compute_mean_estimate`, `_compute_std_estimate` et `_compute_tuning_parameter` de ppi_core.py et supprimer les fonctions analogues dans cluster_core.py. Je te laisse voir si c'est une bonne idée

glide/estimators/cluster_core.py
imerad: Les proxys labellisés et non labellisés suivent la même distribution et la formule du power tuning s'écrit en termes de la variance de ces proxys. Dans ppi standard, on aggrégeait tous les proxys pour estimer leur variance et rien ne nous empêche de faire de même dans le cas clusterisé (la seule différence étant qu'on calcule la variance des sommes par cluster). Donc je pense que Claude s'était trompé sur le fait que la formule de power turning de PPI ne serait pas simplement un cas particulier de la formule dans le cas clusterisé.

glide/estimators/cluster_core.py
imerad: Je l'ai déplacée en méthode de `ClusterPPIMeanEstimator`

README.md
gmartinon-ed: Oui bonne idée, on peut renommer la section avec "algorithms" au lieu de "papers", et changer l'ordre des colonnes : 

Name | Class | Reference Paper | 

Et préciser dans Reference paper l'année entre parenthèses. Pas sûr que les "Venue" et "Original Implementation" soient nécessaires tout compte fait.

Et dans Reference Paper, quand il n'y en a pas dire un truc du genre "original implementation in GLIDE" ou un truc dans le genre, pour dire qu'on fait aussi nos propres algorithmes

glide/core/validation.py
```diff
@@ -159,3 +159,14 @@ def _validate_min_samples(values: NDArray, name: str, stratum_id: Optional[Hasha
                 f"got {len(values)} in stratum '{stratum_id}'."
             )
         raise ValueError(f"'{name}' must have at least 2 non-NaN values; got {len(values)}.")
+
+
+def _validate_labeled_unlabeled_clusters(unique_labeled_clusters: NDArray, unique_unlabeled_clusters: NDArray) -> None:
```
gmartinon-ed: je propose `_validate_unique_clusters` ?

imerad: Appliqué

README.md
imerad: C'est fait, j'ai préféré garder une colonne pour les implémentations originales. J'ai mis des citations dans la colonne des références pour éviter de surcharger le tableau

gmartinon-ed (approved):

PR #278, opened 2026-06-12
Add `generate_multi_binary_dataset` simulator

### LINKED ISSUE #271: `generate_multi_binary_dataset`: binary oracle dataset with M proxy models
## Background

`generate_binary_dataset` generates a synthetic dataset where each sample has a ground-truth binary label and a single proxy prediction, both drawn from a bivariate binary distribution with a specified correlation. Internally it draws all `n_total` pairs in one categorical call over the four possible outcomes, then decodes the two labels via integer arithmetic (`// 2` and `% 2`). This is enough when there is one proxy model, but `MultiPPIMeanEstimator` takes M proxy models simultaneously, each with its own quality (mean and correlation with the true label). There is currently no simulator that produces the M-proxy analogue.

`generate_multi_binary_dataset` fills this gap. It generates a single oracle array `y_true` and a 2D proxy array `y_proxies` of shape `(n_total, M)`, where each column m is a binary proxy with marginal mean `proxy_means[m]` and Pearson correlation `correlations[m]` with `y_true`. The proxies are conditionally independent given `y_true`, which is the natural model: two proxy models make independent errors, but their outputs are correlated with each other only because both track the same ground truth.

The function follows the same oracle-only philosophy as `generate_binary_dataset`: it returns a fully observed dataset with no NaN. Callers that need a labeled/unlabeled split apply `simulate_annotation` separately.

## Design choices

`generate_multi_binary_dataset` lives in a new file `glide/simulators/multi_binary.py`, mirroring the one-function-per-file convention of `binary.py`, `stratified_binary.py`, and `gaussian.py`. It is exported from `glide/simulators/__init__.py` alongside the other simulators.

The `true_mean` parameter is placed in second position, matching the signature of `generate_binary_dataset`. Unlike its single-proxy counterpart, it has no default value: the function already requires the caller to supply `proxy_means` and `correlations`, so there is no convenience gain in defaulting `true_mean`. The per-proxy parameters `proxy_means` and `correlations` are `List[float]` with no default values, since there is no sensible scalar default for a multi-proxy function.

The function cannot reuse `generate_binary_dataset` directly, because that function generates (y_true, y_proxy) pairs jointly via a single categorical draw over four outcomes. This approach does not extend cleanly to M proxies: the outcome space would grow as 2^(M+1), and there is no straightforward way to enforce conditional independence across proxies. Instead, `generate_multi_binary_dataset` uses a two-stage algorithm: generate `y_true` from a Bernoulli distribution, then generate each proxy column conditional on `y_true` using probabilities derived from the same bivariate binary joint distribution that underlies `generate_binary_dataset`. This yields the correct marginal distribution for each (y_true, y_proxy_m) pair while keeping M generation steps instead of 2^(M+1).

This same two-stage algorithm applies cleanly to the single-proxy case, so `generate_binary_dataset` is refactored in this PR to use it as well. The outputs of the refactored function are not sample-for-sample identical to the original (the RNG is consumed differently), but the joint distribution of `(y_true, y_proxy)` is unchanged.

Validation follows the same pattern as `generate_binary_dataset`: each `proxy_means[m]` must be in (0, 1), and each `(true_mean, proxy_means[m], correlations[m])` triple must not lead to negative joint probabilities. Errors for per-proxy failures include the proxy index (e.g., `"proxy 1: impossible combination..."`).

### `generate_multi_binary_dataset`

**Location:** `glide/simulators/multi_binary.py`

**Purpose:** Generate a synthetic binary oracle dataset with M proxy models, each with its own mean and correlation with the true label.

```python
from typing import List, Optional, Tuple, Union

import numpy as np

from glide.core.validation import _validate_bounds, _validate_equal_lengths


def generate_multi_binary_dataset(
    n_total: int,
    true_mean: float,
    proxy_means: List[float],
    correlations: List[float],
    random_seed: Optional[Union[int, np.random.SeedSequence]] = None,
) -> Tuple[NDArray, NDArray]:
    """Generate a synthetic binary oracle dataset with M proxy models.

    Parameters
    ----------
    n_total : int
        Total number of samples.
    true_mean : float
        Expected mean value of the true labels. Must be in (0, 1).
    proxy_means : List[float]
        Expected mean value of each proxy label. Length M determines the number of proxies.
        Each value must be in (0, 1).
    correlations : List[float]
        Pearson correlation between the true label and each proxy label. Length must equal
        ``len(proxy_means)``. Each value must yield a valid joint binary distribution
        given ``true_mean`` and the corresponding ``proxy_means`` entry.
    random_seed : int or np.random.SeedSequence, optional
        Seed for reproducibility.

    Returns
    -------
    Tuple[NDArray, NDArray]
        [0]: array of shape ``(n_total,)``, y_true containing ground-truth labels.
        [1]: array of shape ``(n_total, M)``, y_proxies where column m contains
             proxy labels with mean ``proxy_means[m]`` and correlation ``correlations[m]``
             with y_true.

    Raises
    ------
    ValueError
        If ``proxy_means`` and ``correlations`` have different lengths.
    ValueError
        If ``true_mean`` is not in (0, 1).
    ValueError
        If any ``proxy_means[m]`` is not in (0, 1).
    ValueError
        If the combination of ``true_mean``, ``proxy_means[m]``, and ``correlations[m]``
        is impossible (leads to negative joint probabilities).

    Notes
    -----
    **Step 1 — Generate y_true**

    Each sample's ground-truth label is drawn independently from a Bernoulli distribution:

    ```
    y_true_i ~ Bernoulli(true_mean)
    ```

    **Step 2 — Conditional probabilities for each proxy**

    For proxy m with marginal ``p_p = proxy_means[m]`` and correlation ``rho_m``, the
    bivariate binary joint distribution (see ``generate_binary_dataset``) gives:

    ```
    D_m  = sqrt(p_t * p_p * (1 - p_t) * (1 - p_p))
    p11_m = rho_m * D_m + p_t * p_p
    p01_m = p_p - p11_m
    ```

    where ``p_t = true_mean``. The conditional probabilities follow:

    ```
    P(proxy_m = 1 | y_true = 1) = p11_m / p_t
    P(proxy_m = 1 | y_true = 0) = p01_m / (1 - p_t)
    ```

    **Step 3 — Vectorized conditional generation**

    For each sample i and proxy m, draw independently:

    ```
    proxy_{m,i} | y_true_i ~ Bernoulli(P(proxy_m = 1 | y_true_i))
    ```

    All M proxies are generated in a single vectorised call over the ``(n_total, M)``
    matrix of conditional probabilities. The proxies are conditionally independent given
    y_true, so they have positive marginal correlation with each other only through the
    shared y_true.

    **Validation bounds**

    The same feasibility constraint as in ``generate_binary_dataset`` applies per proxy:

    ```
    max(-p_t * p_p_m, p_p_m + p_t - 1 - p_t * p_p_m) / D_m
        <= correlations[m]
        <= min(p_t * (1 - p_p_m), p_p_m * (1 - p_t)) / D_m
    ```

    Examples
    --------
    >>> import numpy as np
    >>> from glide.simulators import generate_multi_binary_dataset
    >>> y_true, y_proxies = generate_multi_binary_dataset(8, 0.7, [0.6, 0.5], [0.7, 0.6], random_seed=42)
    >>> len(y_true)
    8
    >>> y_proxies.shape
    (8, 2)
    >>> bool(np.all(np.isin(y_true, [0.0, 1.0])))
    True
    >>> bool(np.all(np.isin(y_proxies, [0.0, 1.0])))
    True
    """
    _validate_equal_lengths(proxy_means, correlations, names=["proxy_means", "correlations"])
    _validate_bounds(true_mean, "true_mean", lower=0, upper=1, left_inclusive=False, right_inclusive=False)

    proxy_means_arr = np.array(proxy_means, dtype=float)
    correlations_arr = np.array(correlations, dtype=float)

    for m, p_p in enumerate(proxy_means_arr):
        _validate_bounds(p_p, f"proxy_means[{m}]", lower=0, upper=1, left_inclusive=False, right_inclusive=False)

    p_t = true_mean
    D = np.sqrt(p_t * proxy_means_arr * (1 - p_t) * (1 - proxy_means_arr))
    min_correlations = np.maximum(-p_t * proxy_means_arr, proxy_means_arr + p_t - 1 - p_t * proxy_means_arr) / D
    max_correlations = np.minimum(p_t * (1 - proxy_means_arr), proxy_means_arr * (1 - p_t)) / D

    for m, (rho, lo, hi, p_p) in enumerate(zip(correlations_arr, min_correlations, max_correlations, proxy_means_arr)):
        if rho < lo or rho > hi:
            raise ValueError(
                f"proxy {m}: impossible combination of 'true_mean'={true_mean!r}, "
                f"'proxy_means[{m}]'={p_p!r}, and 'correlations[{m}]'={rho!r}: "
                f"leads to negative joint probabilities; "
                f"possible 'correlations[{m}]' values are in the range ({lo:.3f}, {hi:.3f})."
            )

    rng = np.random.default_rng(seed=random_seed)
    y_true = rng.binomial(1, p_t, size=n_total).astype(float)

    p11 = correlations_arr * D + p_t * proxy_means_arr
    p01 = proxy_means_arr - p11
    cond_prob_given_1 = p11 / p_t
    cond_prob_given_0 = p01 / (1 - p_t)

    is_true_one = y_true[:, np.newaxis].astype(bool)
    cond_probs = np.where(is_true_one, cond_prob_given_1, cond_prob_given_0)
    y_proxies = rng.binomial(1, cond_probs).astype(float)

    return y_true, y_proxies
```

### `glide/simulators/__init__.py` update

Add the import and `__all__` entry after `generate_binary_dataset`:

```python
from glide.simulators.multi_binary import generate_multi_binary_dataset
```

and `"generate_multi_binary_dataset"` to `__all__`.

### `generate_binary_dataset` refactoring

**Location:** `glide/simulators/binary.py`

**Purpose:** Replace the categorical draw + integer-decode approach with the same two-stage conditional Bernoulli algorithm used by `generate_multi_binary_dataset`. The public interface, validation logic, and output distribution are unchanged.

The Steps 2–3 block in the docstring Notes is replaced:

```
**Step 2 — Conditional probabilities**

p11 = correlation * D + p_t * p_p
p01 = p_p - p11

P(y_proxy = 1 | y_true = 1) = p11 / p_t
P(y_proxy = 1 | y_true = 0) = p01 / (1 - p_t)

**Step 3 — Two-stage generation**

y_true_i ~ Bernoulli(p_t)
y_proxy_i | y_true_i ~ Bernoulli(P(y_proxy = 1 | y_true_i))
```

The new generation block (replacing the `rng.choice` + `//` / `%` lines):

```python
rng = np.random.default_rng(seed=random_seed)
y_true = rng.binomial(1, p_t, size=n_total).astype(float)

p11 = correlation * D + p_t * p_p
p01 = p_p - p11
cond_prob_given_1 = p11 / p_t
cond_prob_given_0 = p01 / (1 - p_t)

cond_probs = np.where(y_true.astype(bool), cond_prob_given_1, cond_prob_given_0)
y_proxy = rng.binomial(1, cond_probs).astype(float)
```

The existing functional tests for `generate_binary_dataset` (empirical mean and correlation convergence) must still pass with the refactored implementation.

## Functional tests

**Empirical means and correlations converge to specified parameters**

For a large dataset, the empirical mean of y_true must approach `true_mean`, the empirical mean of each proxy column must approach `proxy_means[m]`, and the empirical Pearson correlation between y_true and each column must approach `correlations[m]`.

```python
def test_generate_multi_binary_dataset_empirical_statistics():
    y_true_mean = 0.6
    y_proxy_means = [0.5, 0.7]
    correlations = [0.5, 0.6]
    y_true, y_proxies = generate_multi_binary_dataset(10000, y_true_mean, y_proxy_means, correlations, random_seed=0)

    assert np.mean(y_true) == pytest.approx(y_true_mean, abs=0.02)
    assert np.mean(y_proxies[:, 0]) == pytest.approx(y_proxy_means[0], abs=0.02)
    assert np.mean(y_proxies[:, 1]) == pytest.approx(y_proxy_means[1], abs=0.02)
    assert np.corrcoef(y_true, y_proxies[:, 0])[0, 1] == pytest.approx(correlations[0], abs=0.05)
    assert np.corrcoef(y_true, y_proxies[:, 1])[0, 1] == pytest.approx(correlations[1], abs=0.05)
```

## Corner cases

- **`proxy_means` and `correlations` have different lengths** — raise `ValueError` before any other validation.
- **`proxy_means[m]` is 0 or 1** — raise `ValueError` naming the index (e.g., `"proxy_means[1] must be in (0, 1)"`).
- **Impossible correlation for proxy m** — raise `ValueError` with the proxy index, the offending values, and the valid range. Stop at the first failing proxy.
- **M=1** — fully supported; returns `y_true` of shape `(n_total,)` and `y_proxies` of shape `(n_total, 1)`. The marginal joint distribution of `(y_true, y_proxies[:, 0])` matches what `generate_binary_dataset` would produce with the same parameters, but the outputs are not sample-for-sample identical: the proxy draw uses a 2D conditional probability array of shape `(n_total, 1)` rather than the 1D array used by `generate_binary_dataset`, so the two RNG call sequences differ.

## Acceptance criteria

- [ ] `generate_multi_binary_dataset` is importable from `glide.simulators` and appears in `glide.simulators.__all__`.
- [ ] Returns `y_true` of shape `(n_total,)` and `y_proxies` of shape `(n_total, M)`, both containing only 0.0 and 1.0.
- [ ] Passing lists of different lengths for `proxy_means` and `correlations` raises `ValueError`.
- [ ] An impossible correlation for proxy m raises `ValueError` that includes the index m and the valid range.
- [ ] Functional test confirms empirical means and correlations converge to specified parameters for `n_total=10000`.
- [ ] `generate_binary_dataset` is refactored to use the two-stage conditional Bernoulli approach; its existing functional tests still pass.

imerad:
### Description

Add `generate_multi_binary_dataset` in `glide/simulators/multi_binary.py`, which generates a single ground-truth array and an `(n_total, M)` proxy array where each column has a specified marginal mean and Pearson correlation with the true label (proxies are conditionally independent given the truth). Also refactors `generate_binary_dataset` to use the same two-stage conditional Bernoulli algorithm; the output distribution is unchanged but RNG consumption differs, so fixed-seed results will change.

Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/7?pane=issue&itemId=195385204&issue=EmertonData%7Cglide%7C271) Closes #271

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
## Code Review

Reviewed at high effort (7 finder angles × parallel verifiers). Findings ranked by severity.

---

### 🔴 Bug — Floating-point overflow crashes `rng.binomial` at boundary correlation

**Files:** `glide/simulators/binary.py:142`, `glide/simulators/multi_binary.py:147`

When `correlation` equals exactly `max_possible_correlation` (a value a caller can compute from the public formula), the sequence `max_possible_correlation / D * D` introduces ~1 ULP of rounding error. This makes `p11` marginally greater than `p_t`, so `cond_prob_given_1 = p11 / p_t` becomes e.g. `1.0000000000000002`. NumPy's `rng.binomial` then raises a raw `ValueError: p > 1` with no user-friendly message — bypassing all of GLIDE's validation.

**Fix:** clip `cond_probs` before the binomial draw:
```python
cond_probs = np.clip(np.where(...), 0.0, 1.0)
```

---

### 🟠 Bug — `n_total` is never validated in `generate_binary_dataset` or `generate_multi_binary_dataset`

**Files:** `glide/simulators/binary.py:116`, `glide/simulators/multi_binary.py:118`

- `n_total=0` silently returns empty arrays.
- `n_total=-1` raises an opaque NumPy `ValueError: negative dimensions are not allowed` from inside `rng.binomial`.

`_validate_strictly_positive` is available in `glide/core/validation.py` but is never called for `n_total`. The clustered simulator validates `n_total >= n_clusters` as a side-effect; the binary simulators have no guard at all. Neither docstring documents the positivity constraint or lists it in `Raises`.

---

### 🟠 Test — `test_generate_multi_binary_dataset_different_seeds_differ` uses `or`, masking proxy-seeding bugs

**File:** `tests/unit/simulators/test_multi_binary.py:53`

```python
assert (not np.array_equal(y_true1, y_true2)) or (not np.array_equal(y_proxies1, y_proxies2))
```

Because `y_true` almost always differs between any two seeds, the `or` passes unconditionally. A bug that generates identical `y_proxies` regardless of seed would never be caught. The same issue exists in the analogous test in `tests/unit/simulators/test_binary.py:48`.

**Fix:** Use two separate assertions:
```python
assert not np.array_equal(y_true1, y_true2)
assert not np.array_equal(y_proxies1, y_proxies2)
```

---

### 🟡 Test — `assert_has_calls` without call-count check in delegation test

**File:** `tests/unit/simulators/test_multi_binary.py:35`

`mock_validate_bounds.assert_has_calls([...])` verifies the listed calls appear as a *subsequence* — it does not verify the total count. The number of `_validate_bounds` calls scales dynamically with the length of `proxy_means` (1 + M calls), so the test should also assert `mock_validate_bounds.call_count == 3`. Without it, an extra spurious call (or a dropped loop iteration) goes undetected. This is the pattern used elsewhere in the codebase for loop-dependent call counts (e.g., `tests/unit/estimators/test_stratified_core.py:43`).

---

### 🟡 Convention — Two Python `for` loops over already-vectorized arrays

**File:** `glide/simulators/multi_binary.py:124–139`

`D`, `min_correlations`, and `max_correlations` are already computed as NumPy arrays, yet the subsequent validation uses two Python loops. CLAUDE.md requires "NumPy vectorization instead of Python for loops whenever possible." `_validate_bounds` accepts arrays natively (uses `np.all` internally), so the `proxy_means` loop (lines 124–125) can be replaced with a single vectorized call. The correlation feasibility loop (lines 132–139) can be replaced with a vectorized mask + `np.argmax` to find the first failing index and produce an equally precise error message.

---

### 🟡 Convention — Error message starts with lowercase in `multi_binary.py`

**File:** `glide/simulators/multi_binary.py:135`

```python
f"proxy {m}: impossible combination of ..."   # lowercase 'i'
```

Every other `ValueError` in the codebase starts with a capital letter (including the analogous message in `binary.py`: `"Impossible combination of ..."`). Fix: `f"Proxy {m}: Impossible combination of ..."`.

---

### 🔵 Maintainability — Core joint-probability math duplicated between `binary.py` and `multi_binary.py`

**Files:** `glide/simulators/binary.py:137–143`, `glide/simulators/multi_binary.py:144–151`

The `p11 / p_t`, `p01 / (1 - p_t)` conditional-probability formulas and the `rng.binomial` generation pattern are copy-pasted verbatim (with the only difference being scalar vs. array). The docstring cross-references `generate_binary_dataset` in prose, but there is no code relationship. A future fix to these formulas (e.g., the floating-point clipping above) must be applied in both places independently.

---

### 🔵 Maintainability — `random_seed` type inconsistency across sibling simulators

**Files:** `glide/simulators/binary.py:13`, `glide/simulators/multi_binary.py:14`, `glide/simulators/stratified_binary.py:15`

`binary.py` and `multi_binary.py` declare `random_seed: Optional[Union[int, np.random.SeedSequence]]`, while `stratified_binary.py` still declares `Optional[int]`. This is inconsistent: `stratified_binary.py` internally calls `np.random.SeedSequence(random_seed)`, so passing a `SeedSequence` to it would silently nest sequences rather than erroring. All three siblings should share the same type signature.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ## [Next release]
 
 ### ✨ Added
+- `generate_multi_binary_dataset` simulator for generating binary oracle datasets with multiple proxies.
```
gmartinon-ed, line 12: A mettre dans la prochaine release (résolution de conflit au merge)

glide/simulators/stratified_binary.py
```diff
@@ -8,10 +8,10 @@
 
 
 def generate_stratified_binary_dataset(
-    n_total: List[int],
-    true_mean: List[float],
-    proxy_mean: List[float],
-    correlation: List[float],
+    n_total: Union[List[int], NDArray],
```
gmartinon-ed: Investiguer si le type ArrayLike ne ferait pas l'affaire ici (dans np.typing)

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.core.validation import _validate_bounds, _validate_equal_lengths
+
+
+def generate_multi_binary_dataset(
+    n_total: int,
+    true_mean: float,
+    proxy_means: Union[List[float], NDArray],
```
gmartinon-ed: ArrayLike ?

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
...
+) -> Tuple[NDArray, NDArray]:
+    """Generate a synthetic binary oracle dataset with multiple proxy models.
+
+    Parameters
+    ----------
+    n_total : int
+        Total number of samples.
+    true_mean : float
+        Expected mean value of the true labels. Must be in (0, 1).
+    proxy_means : list of float or NDArray of shape (M,)
```
gmartinon-ed: Privilégier "of shape (n_proxies,)"

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.core.validation import _validate_bounds, _validate_equal_lengths
+
+
+def generate_multi_binary_dataset(
+    n_total: int,
```
gmartinon-ed, line 10: Désolé je tique là-dessus : est-ce ce serait pas plus logique `n_samples` ici ?

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
...
+           [0., 0.],
+           [1., 0.]])
+    """
+    proxy_means_arr = np.asarray(proxy_means, dtype=float)
+    correlations_arr = np.asarray(correlations, dtype=float)
+
+    _validate_equal_lengths(proxy_means_arr, correlations_arr, names=["proxy_means", "correlations"])
+    _validate_bounds(true_mean, "true_mean", lower=0, upper=1, left_inclusive=False, right_inclusive=False)
+
+    for m, p_p in enumerate(proxy_means_arr):
```
gmartinon-ed:
```suggestion
    for m, proxy_mean in enumerate(proxy_means_arr):
```

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
...
+    >>> y_true, y_proxies = generate_multi_binary_dataset(4, 0.7, [0.6, 0.5], [0.7, 0.6], random_seed=42)
+    >>> y_true
+    array([0., 1., 0., 1.])
+    >>> y_proxies
+    array([[0., 1.],
+           [1., 0.],
+           [0., 0.],
+           [1., 0.]])
+    """
+    proxy_means_arr = np.asarray(proxy_means, dtype=float)
```
gmartinon-ed: p_p ? par symétrie avec le cas simple ?

glide/simulators/multi_binary.py
```diff
@@ -0,0 +1,154 @@
...
+    y_true = rng.binomial(1, p_t, size=n_total).astype(float)
+
+    p11 = correlations_arr * D + p_t * proxy_means_arr
+    p01 = proxy_means_arr - p11
+    cond_prob_given_1 = p11 / p_t
+    cond_prob_given_0 = p01 / (1 - p_t)
+
+    is_true_one = y_true[:, np.newaxis].astype(bool)
+    cond_probs = np.where(is_true_one, cond_prob_given_1, cond_prob_given_0)
+    cond_probs = np.clip(cond_probs, 0.0, 1.0)
```
gmartinon-ed, line 151: Pourquoi ça ? C'est pas déjà garanti par les bornes sur la corrélation ?

gmartinon-ed (changes requested):

glide/simulators/multi_binary.py
imerad: C'était une remarque de Claude, comme quoi certaines valeurs pouvaient provoquer des imprécisions numériques et donner des probabilités légèrement > 1 provoquant une erreur.

glide/simulators/multi_binary.py
imerad: Mais on a récemment fait une refacto pour mettre `n_total ` dans tous les modules simulators

CHANGELOG.md
imerad: Traité

glide/simulators/stratified_binary.py
imerad: Appliqué

glide/simulators/multi_binary.py
imerad: Appliqué

glide/simulators/multi_binary.py
imerad: Traité

glide/simulators/multi_binary.py
imerad: Appliqué

glide/simulators/multi_binary.py
imerad: Appliqué

glide/simulators/multi_binary.py
gmartinon-ed: Avant on avait `n_labeled` et `n_unlabeled`, et une logique à changer. Là on serait juste sur du renaming. Qu'en penses-tu ? on peut faire un ticket séparé

glide/simulators/multi_binary.py
gmartinon-ed: Du coup on propage à `generate_binary` ?

imerad: Fait

glide/simulators/multi_binary.py
imerad: Je viens de créer un ticket pour ça ici https://github.com/orgs/EmertonData/projects/26/views/1?pane=issue&itemId=199484843

gmartinon-ed (approved):

PR #280, opened 2026-06-15
Add Spider 1.0 text-to-SQL case study

imerad:
### Description

Adds `docs/deep_dive/spider_case_study.ipynb`, a deep-dive notebook demonstrating stratified PPI++ on the Spider 1.0 text-to-SQL benchmark. The notebook loads the labelled dataset from HuggingFace and produces a forest plot comparing LLM judge, human-only classical, and GLIDE (stratified PPI) estimates.

Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=184698592&issue=EmertonData%7Cglide%7C270).

The `docs/spider/` directory contains the dataset preprocessing and label generation scripts included here for review on GitHub only. This directory will reside exclusively in the HuggingFace dataset repository and will not be merged into the GitHub repo.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes
- [ ] `make coverage` reports 100% coverage

### LLM usage

- [ ] No LLM used


imerad:
## Code Review

Reviewed at high effort across 7 angles (correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude). 9 findings below, ranked most-severe first.

---

### 🔴 CRITICAL — Notebook: debug override replaces real Spider data with synthetic data

**File:** `docs/deep_dive/spider_case_study.ipynb`, cell 2 (last line)

```python
y_true_oracle, y_proxy = generate_binary_dataset(len(groups), random_seed=RANDOM_SEED)  #######
```

This line overwrites the real `y_true_oracle` and `y_proxy` loaded from the parquet file with synthetic binary data. Every figure in the notebook (forest plot, coverage curves, interval-width plot) comes from `generate_binary_dataset`, not the Spider benchmark. The narrative — the Human Annotator section, the Annex, the HuggingFace link — all describe a real dataset that is never actually used. The `#######` comment is a debug artifact that was left in.

---

### 🔴 CRITICAL — `generate_human_labels.py:13`: `executescript()` silently invalidates all human labels

```python
# cursor.execute(sql)
cursor.executescript(sql)
rows = cursor.fetchall()   # always returns []
```

Empirically verified: `executescript` does not populate the cursor's result set. `fetchall()` always returns `[]`. As a result, `gold_rows` and `predicted_rows` are both `[]` for every syntactically valid query, so `predicted_rows == gold_rows` is always `True` → label `1`. Only queries that throw an exception score `0`. Every syntactically correct but semantically wrong prediction is falsely labeled as correct.

Secondary consequence: `executescript` accepts multiple semicolon-separated statements. An LLM-generated prediction like `DROP TABLE t; SELECT 1;` would silently corrupt the shared SQLite file for that `db_id`, invalidating all subsequent examples from that database.

The commented-out `cursor.execute(sql)` on line 12 is the correct call.

---

### 🟠 HIGH — Notebook: hardcoded local path; HuggingFace load commented out

**File:** `docs/deep_dive/spider_case_study.ipynb`, cell 2

```python
# dataset = load_dataset(HF_REPO, split="train")
ds_path = "../spider/data/spider_dataset.parquet"
dataset = pd.read_parquet(ds_path)
```

The path only exists on the author's machine. Anyone who clones the repo gets `FileNotFoundError` before any computation starts. The `load_dataset` call that would fetch from the published HuggingFace repo needs to be uncommented and the local path removed.

---

### 🟠 HIGH — `build_dataset.py:44`: default `--hf-repo` points to a personal account

```python
parser.add_argument(
    "--hf-repo",
    default="imerad-kv/spider-text-to-sql",          # personal account
    help="... (default: glide-py/spider-text-to-sql)",  # org account
)
```

The default value and the help text disagree. The README (step 5) also instructs `--hf-repo glide-py/spider-text-to-sql`. Running the script without `--hf-repo` silently pushes the dataset to the personal account. The Annex in the notebook links to `glide-py/spider-text-to-sql`, which would be empty or stale.

---

### 🟡 MEDIUM — `generate_human_labels.py:86` and `generate_llm_judge_labels.py:182`: `--input` defaults to `None` but `open()` is called unconditionally

Both scripts declare `--input` with `default=None` and no `required=True`, then call `open(args.input)` directly. Omitting `--input` raises `TypeError: expected str, bytes or os.PathLike object, not NoneType` instead of a helpful argparse error message. The same applies to `--predictions`, `--judge-labels`, and `--human-labels` in `build_dataset.py`. All should use `required=True` or an explicit guard.

---

### 🟡 MEDIUM — `generate_human_labels.py:36`: gold SQL execution failure silently discarded

```python
gold_rows, _ = _execute_query(db_path, gold_sql)   # error discarded
```

If the `.sqlite` file is missing, `sqlite3.connect` creates an empty database and the SELECT fails, returning `(None, error)`. If the predicted SQL then succeeds and returns `[]`, the comparison `[] == None` is `False` → label `0`, with no error recorded. A misconfigured `--db-root` silently zeros out every label for the affected databases. The error from `_execute_query` on the gold side should be checked and surfaced.

---

### 🟡 MEDIUM — `_utils.py:70–74`: 4xx API errors swallowed silently

```python
except anthropic.APIStatusError as exc:
    if exc.status_code >= 500:
        delay = base_delay * (2**attempt)
        time.sleep(delay)
    else:
        return None   # no message printed
```

Non-5xx errors (400 Bad Request, 401 Unauthorized) return `None` immediately with no output. In `generate_predictions.py`, `if predicted_sql is None: continue` silently skips the example. With a wrong `ANTHROPIC_API_KEY`, every call returns `None` and the output file is empty, with `Done. 0 records written` as the only indication that something went wrong.

---

### 🔵 Cleanup — `generate_human_labels.py:46–58`: `_load_checkpoint` copy-pasted from `_utils.py`

`generate_human_labels.py` does not import from `_utils` at all and defines an identical `_load_checkpoint`. The two implementations are already diverging: the scripts in `generate_predictions.py` and `generate_llm_judge_labels.py` both have a `--sleep` / `time.sleep(args.sleep)` call in their main loops; the human labels script does not. Fix: `from _utils import _load_checkpoint` and delete the local copy.

---

### 🔵 Cleanup — `generate_predictions.py:32,50`: generator functions named `*_judge`

```python
def anthropic_judge(model, base_delay, max_retries): ...
def openai_judge(model, base_delay, max_retries): ...
```

These closures generate SQL predictions (returning `Optional[str]`), not correctness labels. The same names in `generate_llm_judge_labels.py` return `Optional[int]`. The naming collision makes the two files harder to read side-by-side and would cause silent data corruption (integers written as `predicted_sql`) if either function were imported into the other module. Rename to `anthropic_predictor` / `openai_predictor` in `generate_predictions.py`.

---

**tl;dr — two blockers before merge:** (1) remove the `generate_binary_dataset` override line in the notebook and uncomment the HuggingFace load; (2) replace `cursor.executescript(sql)` with `cursor.execute(sql)` in `generate_human_labels.py`.

PR #291, opened 2026-06-22
Implement ClusteredPTDMeanEstimator

### LINKED ISSUE #202: Implement ClusterPTD
# TODO

- Update the ticket based on the most recent evolution of the ClusterPPI estimator
- Keep in mind the recent refactor to use "Clustered" rather than "Cluster" as prefix, so ClusteredPTDMeanEstimator below

## Background

In many GenAI evaluation pipelines, annotations are collected in clusters rather than independently. For example, all questions from the same conversation thread, the same document, or the same topic domain may be evaluated together: either a human reviews every question in the cluster, or none are reviewed (proxy labels are used for all). When observations share a cluster, they tend to be correlated — questions from the same conversation use similar vocabulary, reference the same context, and may be answered similarly by the model. Standard statistical methods assume independence across observations, so they underestimate uncertainty when the data is actually clustered.

The Cluster Predict-Then-Debias (PTD) bootstrap corrects for this. It extends the PTD bootstrap (Algorithm 2 in Kluger et al., 2025) to cluster-sampled data (Algorithm 5). Instead of resampling individual observations, it resamples entire clusters with replacement. This preserves the within-cluster correlation structure in each bootstrap replicate and produces confidence intervals that are valid even when observations within the same cluster are highly correlated.

In math notation: let K be the total number of clusters, with n_l labeled clusters (those where every observation has a human label) and n_u unlabeled clusters (those where every observation has only a proxy label). For each bootstrap replicate b, resample n_l cluster indices from the labeled set with replacement. Compute the mean of the resampled labeled cluster means (bootstrap estimate of the true mean θ^(b)) and the mean of the resampled labeled cluster proxy means (γ^{*(b)}). The unlabeled mean γ^{u(b)} is not bootstrapped: by the central limit theorem, the mean of n_u cluster-level proxy means is asymptotically Gaussian with mean equal to the observed cluster-mean average and variance equal to the sample variance of those cluster means divided by n_u. Each replicate draws γ^{u(b)} from this Gaussian. The PTD estimate for replicate b is `θ^(b) + λ * (γ^{u(b)} - γ^{*(b)})` and the percentile interval over the B replicates is the confidence interval. This Gaussian approximation for the unlabeled part is identical to the convolution trick used in `PTDMeanEstimator` and `StratifiedPTDMeanEstimator`, with the only difference being that the variance is computed from cluster-level means (not individual observations).

No assumption is made about labeling probability being uniform across clusters. Clusters are weighted by their size (number of observations) in all mean computations, so a larger cluster contributes proportionally more to the estimate, regardless of how likely it was to be labeled. This mirrors how `StratifiedPTDMeanEstimator` uses stratum size weights `w_k = n_k / N` to combine per-stratum estimates into a population-level estimate. The structural requirement is strict: a cluster must be either fully labeled (every observation has a non-NaN y_true) or fully unlabeled (every observation has NaN in y_true). Partially labeled clusters are not supported.

## Design choices

`ClusterPTDMeanEstimator` lives in `glide/estimators/cluster_ptd.py` and mirrors `StratifiedPTDMeanEstimator` (`glide/estimators/stratified_ptd.py`) in structure, API, and naming. The `estimate` method takes `clusters` instead of `groups`, but the signature is otherwise identical. The key algorithmic difference is the resampling unit: stratified PTD resamples individual observations within each stratum, while cluster PTD resamples entire clusters. Both use the Gaussian convolution trick for the unlabeled part.

Preprocessing and cluster-specific bootstrap helpers live in `glide/estimators/cluster_core.py`, mirroring `glide/estimators/stratified_core.py`. The preprocessing function `_preprocess` aggregates individual observations into cluster-level means and records cluster sizes before any bootstrap takes place. A new helper `_compute_bootstrap_labeled_cluster_means` computes size-weighted bootstrap means by weighting each resampled cluster mean by the cluster size. This is analogous to the stratum size weight `w_k = n_k / N` in `StratifiedPTDMeanEstimator`, which ensures larger groups contribute proportionally more. `_compute_tuning_parameter` and `_compute_bootstrap_mean_estimates` from `ptd_core.py` are reused unchanged.

The only two adaptations to `ptd_core.py`'s interface are: (1) the labeled bootstrap uses size-weighted means instead of simple means; (2) `mean_proxy_unlabeled` and `var_proxy_unlabeled` are derived from cluster sums (defined as `u_k = n_k * μ_proxy_k`) rather than individual observations. Concretely: `mean_proxy_unlabeled = np.sum(u_k) / np.sum(n_k)` and `var_proxy_unlabeled = n_u * np.var(u_k, ddof=1) / np.sum(n_k)**2`. When all clusters have equal size, these reduce exactly to `np.mean(cluster_means)` and `np.var(cluster_means, ddof=1) / n_u`, which are in turn identical to the standard PTD formulas when every cluster contains a single observation.

### `_preprocess`

**Location:** `glide/estimators/cluster_core.py`

**Purpose:** Validate inputs and aggregate individual observations into cluster-level means.

```python

import numpy as np


def _preprocess(
    y_true: NDArray,
    y_proxy: NDArray,
    clusters: NDArray,
) -> Tuple[NDArray, NDArray, NDArray, NDArray, NDArray]:
    if len(y_true) != len(y_proxy) or len(y_true) != len(clusters):
        raise ValueError(
            f"y_true, y_proxy, and clusters must have the same length, "
            f"got {len(y_true)}, {len(y_proxy)}, and {len(clusters)}"
        )
    if np.isnan(y_proxy).any():
        raise ValueError("Input proxy values contain NaN.")

    labeled_y_true_cluster_means: List[float] = []
    labeled_y_proxy_cluster_means: List[float] = []
    labeled_cluster_sizes: List[int] = []
    unlabeled_y_proxy_cluster_means: List[float] = []
    unlabeled_cluster_sizes: List[int] = []

        cluster_mask = clusters == cluster_id
        cluster_y_true = y_true[cluster_mask]
        cluster_y_proxy = y_proxy[cluster_mask]
        labeled_mask = ~np.isnan(cluster_y_true)
        n_labeled_in_cluster = labeled_mask.sum()
        n_total_in_cluster = cluster_mask.sum()

        if 0 < n_labeled_in_cluster < n_total_in_cluster:
            raise ValueError(
                f"Cluster '{cluster_id}' has both labeled and unlabeled observations. "
                f"Each cluster must be entirely labeled or entirely unlabeled."
            )

        if n_labeled_in_cluster == n_total_in_cluster:
            labeled_y_true_cluster_means.append(np.mean(cluster_y_true))
            labeled_y_proxy_cluster_means.append(np.mean(cluster_y_proxy))
            labeled_cluster_sizes.append(n_total_in_cluster)
        else:
            unlabeled_y_proxy_cluster_means.append(np.mean(cluster_y_proxy))
            unlabeled_cluster_sizes.append(n_total_in_cluster)

    n_labeled_clusters = len(labeled_y_true_cluster_means)
    n_unlabeled_clusters = len(unlabeled_y_proxy_cluster_means)
    if n_labeled_clusters < 2:
        raise ValueError(
            f"Need at least 2 labeled clusters; got {n_labeled_clusters}."
        )
    if n_unlabeled_clusters < 2:
        raise ValueError(
            f"Need at least 2 unlabeled clusters; got {n_unlabeled_clusters}."
        )

    labeled_y_true_means = np.array(labeled_y_true_cluster_means)
    labeled_y_proxy_means = np.array(labeled_y_proxy_cluster_means)
    unlabeled_y_proxy_means = np.array(unlabeled_y_proxy_cluster_means)
    labeled_sizes = np.array(labeled_cluster_sizes)
    unlabeled_sizes = np.array(unlabeled_cluster_sizes)
    return labeled_y_true_means, labeled_y_proxy_means, unlabeled_y_proxy_means, labeled_sizes, unlabeled_sizes
```

### `_compute_bootstrap_labeled_cluster_means`

**Location:** `glide/estimators/cluster_core.py`

**Purpose:** Bootstrap size-weighted means of labeled cluster means by resampling cluster indices with replacement.

```python
from typing import Tuple

import numpy as np


def _compute_bootstrap_labeled_cluster_means(
    labeled_y_true_cluster_means: NDArray,
    labeled_y_proxy_cluster_means: NDArray,
    labeled_cluster_sizes: NDArray,
    n_bootstrap: int,
    rng: np.random.Generator,
) -> Tuple[NDArray, NDArray]:
    n_l = len(labeled_y_true_cluster_means)
    idx = rng.choice(n_l, size=(n_bootstrap, n_l), replace=True)
    resampled_sizes = labeled_cluster_sizes[idx]  # shape (n_bootstrap, n_l)
    total_resampled_weights = np.sum(resampled_sizes, axis=1)  # shape (n_bootstrap,)
    bootstrap_y_true_means = (
        np.sum(labeled_y_true_cluster_means[idx] * resampled_sizes, axis=1) / total_resampled_weights
    )
    bootstrap_y_proxy_labeled_means = (
        np.sum(labeled_y_proxy_cluster_means[idx] * resampled_sizes, axis=1) / total_resampled_weights
    )
    return bootstrap_y_true_means, bootstrap_y_proxy_labeled_means
```

### `ClusterPTDMeanEstimator`

**Location:** `glide/estimators/cluster_ptd.py`

**Purpose:** Public estimator class that wires preprocessing, bootstrap, tuning, and result assembly.

```python
from typing import Optional

import numpy as np

from glide.confidence_intervals import BootstrapConfidenceInterval
from glide.core.mean_inference_result import PredictionPoweredMeanInferenceResult
from glide.core.utils import compute_effective_sample_size
from glide.estimators.cluster_core import (
    _compute_bootstrap_labeled_cluster_means,
    _preprocess,
)
from glide.estimators.ptd_core import (
    _compute_bootstrap_mean_estimates,
    _compute_tuning_parameter,
)


class ClusterPTDMeanEstimator:
    """Cluster Predict-Then-Debias estimator for population mean.

    Extends PTD to datasets where observations are grouped into clusters and
    each cluster is either entirely labeled or entirely unlabeled. The bootstrap
    resamples whole clusters (rather than individual observations), which accounts
    for within-cluster correlation and produces valid confidence intervals under
    cluster sampling designs.

    A per-cluster power-tuning parameter λ is estimated from the joint bootstrap
    covariance of the labeled means and proxy means. The final confidence interval
    is a percentile interval over B bootstrap replicates of:

        θ_b = mean(y_true_b) + λ * (mean(y_proxy_unlabeled_b) - mean(y_proxy_labeled_b))

    where each _b subscript denotes a resample of cluster-level means.

    References
    ----------
    Kluger, Dan M., Kerri Lu, Tijana Zrnic, Sherrie Wang, and Stephen Bates. "Prediction-powered inference with imputed covariates and nonuniform sampling." arXiv preprint arXiv:2501.18577 (2025).

    Examples
    --------
    >>> import numpy as np
    >>> from glide.estimators import ClusterPTDMeanEstimator
    >>> y_true = np.array([5.0, 6.0, np.nan, np.nan, 5.0, 6.0, np.nan, np.nan])
    >>> y_proxy = np.array([4.9, 6.1, 5.2, 6.1, 4.9, 6.1, 5.2, 6.1])
    >>> clusters = np.array(["A", "A", "B", "B", "C", "C", "D", "D"])
    >>> estimator = ClusterPTDMeanEstimator()
    >>> result = estimator.estimate(y_true, y_proxy, clusters, n_bootstrap=5, random_seed=0)
    >>> print(result)  # fill in exact output after running the implementation
    Metric: Metric
    Point Estimate: X.XXX
    Confidence Interval (95%): [X.XXX, X.XXX]
    Estimator : ClusterPTDMeanEstimator
    n_true: 4
    n_proxy: 8
    Effective Sample Size: X
    """

    def estimate(
        self,
        y_true: NDArray,
        y_proxy: NDArray,
        clusters: NDArray,
        metric_name: str = "Metric",
        n_bootstrap: int = 2000,
        power_tuning: bool = True,
        random_seed: Optional[int] = None,
    ) -> PredictionPoweredMeanInferenceResult:
        """Estimate the population mean using the Cluster Predict-Then-Debias bootstrap.

        Aggregates observations into cluster-level means, bootstraps labeled and
        unlabeled cluster means independently, and assembles per-replicate PTD
        estimates to form a percentile confidence interval.

        Labeled and unlabeled clusters are distinguished by the NaN pattern in
        ``y_true``: a cluster is labeled if every one of its ``y_true`` entries is
        finite, and unlabeled if every entry is ``np.nan``. Partially labeled clusters
        (some NaN, some finite within the same cluster) are not supported.

        Parameters
        ----------
        y_true : NDArray
            Array of observations, shape ``(n_samples,)``.
            All observations in the same cluster must share the same label status.
        y_proxy : NDArray
        clusters : NDArray
        metric_name : str, optional
        n_bootstrap : int, optional
            Number of bootstrap resamples. Defaults to ``2000``.
        power_tuning : bool, optional
            If ``True`` (default), estimate the optimal power-tuning parameter
            ``λ`` from the bootstrap covariances. If ``False``, use ``λ = 1.0``.
        random_seed : int, optional
            Seed for the random number generator. Defaults to ``None``.

        Returns
        -------
        PredictionPoweredMeanInferenceResult
            Contains the bootstrap-based confidence interval, the metric name,
            the estimator name (``"ClusterPTDMeanEstimator"``), and the counts
            ``n_true`` (total labeled observations) and ``n_proxy`` (total dataset size).

        Raises
        ------
        ValueError
            - If ``y_true``, ``y_proxy``, and ``clusters`` do not all have the same length.
            - If any proxy value is NaN.
            - If any cluster contains both labeled and unlabeled observations.
            - If fewer than 2 clusters are fully labeled.
            - If fewer than 2 clusters are fully unlabeled.
        """
        (
            labeled_y_true_cluster_means,
            labeled_y_proxy_cluster_means,
            unlabeled_y_proxy_cluster_means,
            labeled_cluster_sizes,
            unlabeled_cluster_sizes,
        ) = _preprocess(y_true, y_proxy, clusters)

        n_unlabeled_clusters = len(unlabeled_y_proxy_cluster_means)
        # Cluster sums: each cluster's contribution to the population total
        unlabeled_cluster_sums = unlabeled_cluster_sizes * unlabeled_y_proxy_cluster_means
        total_unlabeled_size = np.sum(unlabeled_cluster_sizes)
        mean_proxy_unlabeled = np.sum(unlabeled_cluster_sums) / total_unlabeled_size
        var_proxy_unlabeled = n_unlabeled_clusters * np.var(unlabeled_cluster_sums, ddof=1) / total_unlabeled_size**2

        bootstrap_y_true_means, bootstrap_y_proxy_labeled_means = _compute_bootstrap_labeled_cluster_means(
            labeled_y_true_cluster_means, labeled_y_proxy_cluster_means, labeled_cluster_sizes, n_bootstrap, rng
        )
        lambda_ = _compute_tuning_parameter(
            bootstrap_y_true_means, bootstrap_y_proxy_labeled_means, var_proxy_unlabeled, power_tuning
        )
        bootstrap_estimates = _compute_bootstrap_mean_estimates(
            bootstrap_y_true_means,
            bootstrap_y_proxy_labeled_means,
            mean_proxy_unlabeled,
            var_proxy_unlabeled,
            lambda_,
            rng,
        )

        confidence_interval = BootstrapConfidenceInterval(
            bootstrap_estimates=bootstrap_estimates,
        )
        effective_sample_size = compute_effective_sample_size(y_true_labeled, confidence_interval.var)
        result = PredictionPoweredMeanInferenceResult(
            confidence_interval=confidence_interval,
            metric_name=metric_name,
            n_true=len(y_true_labeled),
            n_proxy=len(y_true),
            effective_sample_size=effective_sample_size,
        )
        return result
```

Add the following import and `__all__` entry (keep alphabetical order within each group):

```python
from glide.estimators.cluster_ptd import ClusterPTDMeanEstimator
```

Add `"ClusterPTDMeanEstimator"` to `__all__`.

### Doctest values

After implementing, run the class-level doctest example and replace the `X.XXX` placeholders with the actual output. For the method-level doctest, `result.n_true == 4` and `result.n_proxy == 8` are exact and can be verified analytically from the input arrays.

## Functional tests

**Analytical point estimate with power_tuning=False**

When `power_tuning=False` (λ = 1), the PTD estimate equals `mean(y_true_cluster_means) + mean(y_proxy_unlabeled_cluster_means) - mean(y_proxy_labeled_cluster_means)`. With clusters where all observations in each cluster are identical, the bootstrap always resamples the same cluster means and every bootstrap estimate is identical, so `confidence_interval.mean` is exact. This test verifies the formula without relying on stochastic convergence.

```python
def test_point_estimate_lambda_one_with_identical_clusters():
    # 2 labeled clusters: A and B, each with constant y_true=5.0, y_proxy=4.8
    # 2 unlabeled clusters: C and D, each with constant y_proxy=5.3
    # PTD formula with λ=1: 5.0 + (5.3 - 4.8) = 5.5 (exact, since all resamples are identical)
    y_true = np.array([5.0, 5.0, 5.0, 5.0, np.nan, np.nan, np.nan, np.nan])
    y_proxy = np.array([4.8, 4.8, 4.8, 4.8, 5.3, 5.3, 5.3, 5.3])
    clusters = np.array(["A", "A", "B", "B", "C", "C", "D", "D"])

    result = ClusterPTDMeanEstimator().estimate(
        y_true, y_proxy, clusters, power_tuning=False, n_bootstrap=50, random_seed=0
    )
    assert result.confidence_interval.mean == pytest.approx(5.5, abs=1e-10)
```

**Single-observation clusters produce identical results to PTDMeanEstimator**

When every cluster contains exactly one observation, `labeled_y_true_cluster_means` equals `y_true_labeled`, `labeled_y_proxy_cluster_means` equals `y_proxy_labeled`, and `var_proxy_unlabeled = np.var(y_proxy_unlabeled, ddof=1) / n_unlabeled` — identical to what `PTDMeanEstimator` computes. Given the same `random_seed`, the full bootstrap distribution (and therefore the full confidence interval) must be identical.

```python
def test_single_observation_clusters_equals_ptd():
    y_true = np.array([5.0, 6.0, np.nan, np.nan])
    y_proxy = np.array([4.9, 6.1, 5.2, 6.1])
    clusters = np.array(["A", "B", "C", "D"])

    cluster_result = ClusterPTDMeanEstimator().estimate(
        y_true, y_proxy, clusters, power_tuning=False, n_bootstrap=50, random_seed=0
    )
    ptd_result = PTDMeanEstimator().estimate(
        y_true, y_proxy, power_tuning=False, n_bootstrap=50, random_seed=0
    )
        ptd_result.confidence_interval.lower_bound, abs=1e-10
    )
        ptd_result.confidence_interval.upper_bound, abs=1e-10
    )
```

## Corner cases

- A cluster with both labeled and unlabeled observations (e.g., `y_true = [5.0, np.nan]` for cluster "A") must raise `ValueError` with the cluster ID in the message.
- Exactly 1 labeled cluster (n_l = 1) must raise `ValueError` because a bootstrap of one cluster has zero variance.
- Exactly 1 unlabeled cluster (n_u = 1) must raise `ValueError` for the same reason.
- A cluster where all observations have `np.nan` in `y_proxy` is caught by the global NaN-in-proxy check before cluster iteration.
- `random_seed=None` (default) must produce non-deterministic results across two calls; `random_seed=0` must produce identical results across two calls.

## References

- Kluger, Daniel M., et al. "Prediction-Powered Inference with Imputed Covariates and Nonuniform Sampling." arXiv:2501.18577 (2025). https://arxiv.org/abs/2501.18577
- Kluger, Daniel M. PTDBoot Tutorial. https://dankluger.github.io/PTDBootTutorial/Tutorial.html
- Kluger, Daniel M. PTDBoot R package — `PTD_cluster_bootstrap.R` (reference implementation of Algorithm 5). https://github.com/DanKluger/PTDBoot/blob/master/R/PTD_cluster_bootstrap.R

## Acceptance criteria

- [ ] `ClusterPTDMeanEstimator` is importable from `glide.estimators` and listed in `__all__`.
- [ ] `_preprocess` raises `ValueError` with the cluster ID when a cluster contains both labeled and unlabeled observations.
- [ ] `_preprocess` raises `ValueError` when there are fewer than 2 labeled clusters or fewer than 2 unlabeled clusters.
- [ ] The analytical test `test_point_estimate_lambda_one_with_identical_clusters` passes: `confidence_interval.mean == 5.5` to within `1e-10`.
- [ ] `result.n_true` equals the total number of labeled individual observations (not clusters); `result.n_proxy` equals the total number of observations.
- [ ] All class-level and method-level doctests pass (exact values filled in after running the implementation).
- [ ] `cluster_core.py` contains exactly `_preprocess` and `_compute_bootstrap_labeled_cluster_means`; `_compute_tuning_parameter` and `_compute_bootstrap_mean_estimates` from `ptd_core.py` are reused unchanged.
- [ ] `tests/unit/estimators/test_cluster_core.py` and `tests/unit/estimators/test_cluster_ptd.py` exist and mirror the structure of `test_stratified_core.py` and `test_stratified_ptd.py`.
- [ ] The estimator table in `README.md` is updated to include `ClusterPTDMeanEstimator`.

# Open question 
Can we mutualize/factorize parts together with existing PTD or ClusterPPI ? Have we already done some of the work ? Can existing methods be extended and generalized ? -> Do not hesitate to report and argue on this point during implementation without going too far from the ticket. Maybe a topic of discussion during PR review, but this needs alignment before implementation.

imerad:
### Description

Adds `ClusteredPTDMeanEstimator`, a cluster-robust Predict-Then-Debias estimator that resamples whole clusters to account for within-cluster correlation. The clustered preprocessing logic (`_preprocess`, `_compute_bootstrap_labeled_cluster_means`) has been factored into a new shared module `clustered_core.py`, and `ClusteredPPIMeanEstimator` has been updated to delegate to it.

Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=171055706&issue=EmertonData%7Cglide%7C202) Closes #202

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,12 +79,14 @@ If you use GLIDE in your work, please cite us using the "Cite this repository" b
 | Prediction-Powered Inference | `estimators.PPIMeanEstimator` (with `power_tuning=False`) | [[1]](#ref-1) | [Link](https://github.com/aangelopoulos/ppi_py/) |
 | PPI++ | `estimators.PPIMeanEstimator` | [[2]](#ref-2) | [Link](https://github.com/aangelopoulos/ppi_py/tree/ppi++) |
 | Stratified Prediction-Powered Inference | `estimators.StratifiedPPIMeanEstimator` | [[3]](#ref-3) | — |
+| Clustered Prediction-Powered Inference | `estimators.ClusteredPPIMeanEstimator` | — | [Link](https://github.com/davidbroska/ppi_py) |
 | Stratified Sampling | `samplers.StratifiedSampler` | [[4]](#ref-4) | [Link](https://github.com/amazon-science/ssepy) |
 | Active Statistical Inference | `estimators.ASIMeanEstimator` | [[5]](#ref-5), [[6]](#ref-6) | [Link](https://github.com/tijana-zrnic/active-inference) |
 | Active Sampling | `samplers.ActiveSampler` | [[5]](#ref-5), [[6]](#ref-6) | [Link](https://github.com/kristinagligoric/confidence-driven-inference) |
-| Predict-Then-Debias | `estimators.PTDMeanEstimator`, `estimators.StratifiedPTDMeanEstimator`, `estimators.IPWPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
-| Cluster Prediction-Powered Inference | `estimators.ClusteredPPIMeanEstimator` | — | [Link](https://github.com/davidbroska/ppi_py) |
+| Predict-Then-Debias | `estimators.PTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
+| Stratified Predict-Then-Debias | `estimators.StratifiedPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
 | Clustered Predict-Then-Debias | `estimators.ClusteredPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
+| IPW Predict-Then-Debias | `estimators.IPWPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
 
 ### References
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ If you use GLIDE in your work, please cite us using the "Cite this repository" b
 | Active Sampling | `samplers.ActiveSampler` | [[5]](#ref-5), [[6]](#ref-6) | [Link](https://github.com/kristinagligoric/confidence-driven-inference) |
 | Predict-Then-Debias | `estimators.PTDMeanEstimator`, `estimators.StratifiedPTDMeanEstimator`, `estimators.IPWPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
 | Cluster Prediction-Powered Inference | `estimators.ClusteredPPIMeanEstimator` | — | [Link](https://github.com/davidbroska/ppi_py) |
+| Clustered Predict-Then-Debias | `estimators.ClusteredPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
 
 ### References
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -45,6 +45,7 @@
 | [`PTDMeanEstimator`](estimators.md#glide.estimators.ptd.PTDMeanEstimator) | Predict-then-debias with bootstrap confidence intervals |
 | [`StratifiedPTDMeanEstimator`](estimators.md#glide.estimators.stratified_ptd.StratifiedPTDMeanEstimator) | PTD with per-stratum optimal weighting |
 | [`IPWPTDMeanEstimator`](estimators.md#glide.estimators.ipw_ptd.IPWPTDMeanEstimator) | PTD with inverse probability weighting |
+| [`ClusteredPTDMeanEstimator`](estimators.md#glide.estimators.clusterd_ptd.ClusterdPTDMeanEstimator) | PTD for clustered data |
 
 ## Confidence Intervals
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -40,12 +40,12 @@
 |-------|-------------|
 | [`PPIMeanEstimator`](estimators.md#glide.estimators.ppi.PPIMeanEstimator) | Combines labeled data with proxy predictions |
 | [`StratifiedPPIMeanEstimator`](estimators.md#glide.estimators.stratified_ppi.StratifiedPPIMeanEstimator) | PPI with per-stratum optimal weighting |
-| [`ASIMeanEstimator`](estimators.md#glide.estimators.asi.ASIMeanEstimator) | Active statistical inference with non-uniform sampling |
 | [`ClusteredPPIMeanEstimator`](estimators.md#glide.estimators.clustered_ppi.ClusteredPPIMeanEstimator) | PPI for clustered data |
+| [`ASIMeanEstimator`](estimators.md#glide.estimators.asi.ASIMeanEstimator) | Active statistical inference with non-uniform sampling |
 | [`PTDMeanEstimator`](estimators.md#glide.estimators.ptd.PTDMeanEstimator) | Predict-then-debias with bootstrap confidence intervals |
 | [`StratifiedPTDMeanEstimator`](estimators.md#glide.estimators.stratified_ptd.StratifiedPTDMeanEstimator) | PTD with per-stratum optimal weighting |
-| [`IPWPTDMeanEstimator`](estimators.md#glide.estimators.ipw_ptd.IPWPTDMeanEstimator) | PTD with inverse probability weighting |
 | [`ClusteredPTDMeanEstimator`](estimators.md#glide.estimators.clustered_ptd.ClusteredPTDMeanEstimator) | PTD for clustered data |
+| [`IPWPTDMeanEstimator`](estimators.md#glide.estimators.ipw_ptd.IPWPTDMeanEstimator) | PTD with inverse probability weighting |
 
 ## Confidence Intervals
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -45,7 +45,7 @@
 | [`PTDMeanEstimator`](estimators.md#glide.estimators.ptd.PTDMeanEstimator) | Predict-then-debias with bootstrap confidence intervals |
 | [`StratifiedPTDMeanEstimator`](estimators.md#glide.estimators.stratified_ptd.StratifiedPTDMeanEstimator) | PTD with per-stratum optimal weighting |
 | [`IPWPTDMeanEstimator`](estimators.md#glide.estimators.ipw_ptd.IPWPTDMeanEstimator) | PTD with inverse probability weighting |
-| [`ClusteredPTDMeanEstimator`](estimators.md#glide.estimators.clusterd_ptd.ClusterdPTDMeanEstimator) | PTD for clustered data |
+| [`ClusteredPTDMeanEstimator`](estimators.md#glide.estimators.clustered_ptd.ClusteredPTDMeanEstimator) | PTD for clustered data |
 
 ## Confidence Intervals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ## [Next release]
 
 ### ✨ Added
-- `ClusteredPTDMeanEstimator`: cluster-robust Predict-Then-Debias estimator for clsutered data.
+- `ClusteredPTDMeanEstimator`: cluster-robust Predict-Then-Debias estimator for clustered data.
 - `MultiPPIMeanEstimator` for prediction-powered inference with multiple proxy models.
 - `generate_multi_binary_dataset` simulator for generating binary oracle datasets with multiple proxies.
 - Spider Text-to-SQL case study
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ## [Next release]
 
 ### ✨ Added
-- `ClusteredPTDMeanEstimator`: cluster-robust Predict-Then-Debias estimator that resamples whole clusters to account for within-cluster correlation.
+- `ClusteredPTDMeanEstimator`: cluster-robust Predict-Then-Debias estimator for clsutered data.
 - `generate_multi_binary_dataset` simulator for generating binary oracle datasets with multiple proxies.
 - Spider Text-to-SQL case study
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ## [Next release]
 
 ### ✨ Added
+- `ClusteredPTDMeanEstimator`: cluster-robust Predict-Then-Debias estimator that resamples whole clusters to account for within-cluster correlation.
 - `generate_multi_binary_dataset` simulator for generating binary oracle datasets with multiple proxies.
 - Spider Text-to-SQL case study
 
--- a/docs/api/estimators.md
+++ b/docs/api/estimators.md
@@ -41,3 +41,7 @@
 ---
 
 ::: glide.estimators.stratified_ptd.StratifiedPTDMeanEstimator
+
+---
+
+::: glide.estimators.clustered_ptd.ClusteredPTDMeanEstimator
--- a/glide/estimators/__init__.py
+++ b/glide/estimators/__init__.py
@@ -2,6 +2,7 @@ from glide.estimators.asi import ASIMeanEstimator
 from glide.estimators.classical import ClassicalMeanEstimator
 from glide.estimators.clustered_classical import ClusteredClassicalMeanEstimator
 from glide.estimators.clustered_ppi import ClusteredPPIMeanEstimator
+from glide.estimators.clustered_ptd import ClusteredPTDMeanEstimator
 from glide.estimators.ipw_classical import IPWClassicalMeanEstimator
 from glide.estimators.ipw_ptd import IPWPTDMeanEstimator
 from glide.estimators.ppi import PPIMeanEstimator
@@ -22,4 +23,5 @@ __all__ = [
     "PTDMeanEstimator",
     "StratifiedPTDMeanEstimator",
     "IPWPTDMeanEstimator",
+    "ClusteredPTDMeanEstimator",
 ]
--- a/glide/estimators/clustered_core.py
+++ b/glide/estimators/clustered_core.py
@@ -1,0 +1,58 @@
+from typing import Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.core.validation import (
+    _validate_bounds,
+    _validate_equal_lengths,
+    _validate_has_no_nan,
+    _validate_unique_clusters,
+)
+
+
+def _preprocess(
+    y_true: NDArray,
+    y_proxy: NDArray,
+    clusters: NDArray,
+) -> Tuple[NDArray, NDArray, NDArray]:
+    _validate_equal_lengths(y_true, y_proxy, clusters, names=["y_true", "y_proxy", "clusters"])
+    _validate_has_no_nan(y_proxy, "y_proxy")
+    _validate_has_no_nan(clusters, "clusters")
+
+    labeled_mask = ~np.isnan(y_true)
+    labeled_clusters = clusters[labeled_mask]
+    unlabeled_clusters = clusters[~labeled_mask]
+
+    unique_labeled_clusters, labeled_cluster_indices = np.unique(labeled_clusters, return_inverse=True)
+    unique_unlabeled_clusters, unlabeled_cluster_indices = np.unique(unlabeled_clusters, return_inverse=True)
+
+    _validate_unique_clusters(unique_labeled_clusters, unique_unlabeled_clusters)
+
+    n_labeled_clusters = len(unique_labeled_clusters)
+    n_unlabeled_clusters = len(unique_unlabeled_clusters)
+
+    _validate_bounds(
+        n_labeled_clusters,
+        "n_labeled_clusters",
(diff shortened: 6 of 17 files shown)


imerad:
## Code Review

5 findings, ranked by severity.

---

### 1. `clustered_core._preprocess` skips `_validate_y_true`

**File:** `glide/estimators/clustered_core.py`, line 19

`_preprocess` validates `y_proxy` and `clusters` for NaN but never calls `_validate_y_true` on `y_true`. Every other core preprocess function (`ppi_core`, `ptd_core`, `stratified_core`) calls it. As a result, finite but invalid labeled values (e.g. `np.inf`) pass all validation and silently produce infinite confidence interval bounds.

**Reproduction:**
```python
y_true = np.array([np.inf, 2.0, np.nan, np.nan])
clusters = np.array(["A", "B", "C", "D"])
y_proxy = np.array([1.0, 2.0, 3.0, 4.0])
ClusteredPTDMeanEstimator().estimate(y_true, y_proxy, clusters)
# → CI with lower_bound = inf, no error raised
```

---

### 2. `_preprocess` runs twice per `estimate` call

**File:** `glide/estimators/clustered_ptd.py`, line 157

`ClusteredClassicalMeanEstimator().estimate(y_true, clusters)` calls its own internal `_preprocess` (with `np.unique` + 4× `np.bincount`) on data already preprocessed at line 125. The same pattern exists in `clustered_ppi.py` line 135 — this PR carries it forward rather than fixing it. Worth addressing in both files at the same time.

---

### 3. `n_proxy=len(y_true)` instead of `len(y_proxy)`

**File:** `glide/estimators/clustered_ptd.py`, line 164

`ClusteredPPIMeanEstimator` (line 142) correctly writes `n_proxy=len(y_proxy)`; `ClusteredPTDMeanEstimator` writes `n_proxy=len(y_true)`. The numeric value is identical today since both arrays are validated to have equal lengths, but it's semantically wrong and will silently produce an incorrect `n_proxy` if the two arrays are ever allowed to differ.

---

### 4. Needless `int()` cast on `n_labeled`

**File:** `glide/estimators/clustered_ptd.py`, line 156

```python
n_labeled = int(np.sum(~np.isnan(y_true)))  # int() is redundant
```

`np.sum` on a boolean array already returns an integer-typed NumPy scalar. `clustered_ppi.py` line 134 omits the cast. CLAUDE.md: *"Do not use needless type conversions like `float()` or `int()` unless required by the caller."*

---

### 5. Duplicate `estimator_name` assertion across two test functions

**File:** `tests/unit/estimators/test_clustered_ptd.py`, lines 40 and 46

`result.estimator_name == "ClusteredPTDMeanEstimator"` is asserted in both `test_estimate_returns_valid_inference_result` and `test_estimate_metadata` with identical inputs. CLAUDE.md: *"do not write two test functions that exercise the exact same code path with equivalent inputs."* Remove the assertion from `test_estimate_returns_valid_inference_result` where it doesn't belong thematically.

docs/api/index.md
```diff
@@ -45,6 +45,7 @@
 | [`PTDMeanEstimator`](estimators.md#glide.estimators.ptd.PTDMeanEstimator) | Predict-then-debias with bootstrap confidence intervals |
 | [`StratifiedPTDMeanEstimator`](estimators.md#glide.estimators.stratified_ptd.StratifiedPTDMeanEstimator) | PTD with per-stratum optimal weighting |
 | [`IPWPTDMeanEstimator`](estimators.md#glide.estimators.ipw_ptd.IPWPTDMeanEstimator) | PTD with inverse probability weighting |
+| [`ClusteredPTDMeanEstimator`](estimators.md#glide.estimators.clustered_ptd.ClusteredPTDMeanEstimator) | PTD for clustered data |
```
gmartinon-ed, line 48: A rapprocher de Stratified

README.md
```diff
@@ -84,6 +84,7 @@ If you use GLIDE in your work, please cite us using the "Cite this repository" b
 | Active Sampling | `samplers.ActiveSampler` | [[5]](#ref-5), [[6]](#ref-6) | [Link](https://github.com/kristinagligoric/confidence-driven-inference) |
 | Predict-Then-Debias | `estimators.PTDMeanEstimator`, `estimators.StratifiedPTDMeanEstimator`, `estimators.IPWPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
 | Cluster Prediction-Powered Inference | `estimators.ClusteredPPIMeanEstimator` | — | [Link](https://github.com/davidbroska/ppi_py) |
+| Clustered Predict-Then-Debias | `estimators.ClusteredPTDMeanEstimator` | [[7]](#ref-7) | [Link](https://github.com/DanKluger/PTDBoot) |
```
gmartinon-ed: Deux remarques : 
- "Clustered" avec le -ed à la fin
- Je pense qu'on peut splitter les lignes PTD en 4 (PTD, Stratified, Clustered, IPW)

gmartinon-ed (changes requested):

docs/api/index.md
imerad: Fait, je rapproche également ClusteredPPI de StratifiedPPI

README.md
imerad: Traité, j'ai réorganisé les lignes comme dans l'API index

gmartinon-ed (approved):

PR #330, opened 2026-07-02
Extract labeled/unlabeled NaN split into method-agnostic core.py

### LINKED ISSUE #312: Extract the labeled/unlabeled NaN split into a method-agnostic `estimators/core.py`
# TODO

- See how this ticket can encompass the module `cluster_core.py`. It may or may not have an impact on clustered algorithms.

## Background

The prediction-powered estimators all partition aligned `(y_true, y_proxy)` arrays into the triple `(y_true_labeled, y_proxy_labeled, y_proxy_unlabeled)` using the NaN convention (a sample is labeled where `y_true` is finite). This exact partition is currently duplicated verbatim in three places: `PPIMeanEstimator._preprocess` ([glide/estimators/ppi.py:62-67](glide/estimators/ppi.py#L62-L67)), `PTDMeanEstimator._preprocess` ([glide/estimators/ptd.py:66-71](glide/estimators/ptd.py#L66-L71)), and the per-stratum loop in `stratified_core._preprocess` ([glide/estimators/stratified_core.py:30-35](glide/estimators/stratified_core.py#L30-L35)).

The goal is to extract this split once into a shared helper `_split_labeled_unlabeled` and have all three call sites use it, removing the duplication. This is a behaviour-preserving refactor fully covered by the existing estimator tests.

## Design choices

The helper goes in a new method-agnostic module `glide/estimators/core.py`, **not** in `ppi_core.py`. The split is a general data-partitioning primitive with nothing PPI-specific about it (it only applies the NaN convention), so housing it in `ppi_core.py` would force `ptd.py` and `stratified_core.py` to import from a *PPI* module they otherwise have no relationship with — PTD is not PPI. The existing `ppi_core.py` / `ptd_core.py` / `stratified_core.py` are all method-specific cores; the unprefixed `core.py` is their natural method-agnostic counterpart and gives future cross-method primitives a home (the planned `glide/monitors/core.py` mirrors this split). The helper is a free function, mirroring the existing `*_core` modules, and every caller invokes it directly rather than wrapping it. The helper centralizes the NaN convention (`labeled_mask = ~np.isnan(y_true)`) and **also returns that mask** alongside the three arrays, so each caller runs `_validate_sample_sizes` on the returned mask without recomputing `~np.isnan(y_true)` — the duplicate the inline copies would otherwise leave behind. Validation deliberately stays at the call sites rather than moving inside the helper: the upcoming `PPIMeanMonitor` calls this helper repeatedly *after* its own `_preprocess` has already validated each batch, and on pooled "earlier batches" arrays where a sample-size check is meaningless. An in-helper `_validate_sample_sizes` would turn every such call into a redundant downstream re-check, which the codebase forbids. Keeping the helper a pure partition primitive lets each consumer validate (or not) at its own boundary.

## Implementation

Create `glide/estimators/core.py` with the helper. Its imports are minimal:

```python
from typing import Tuple

import numpy as np
```

The helper returns the three split arrays followed by the boolean `labeled_mask` it computed:

```python
def _split_labeled_unlabeled(y_true: NDArray, y_proxy: NDArray) -> Tuple[NDArray, NDArray, NDArray, NDArray]:
    y_proxy_labeled = y_proxy[labeled_mask]
    y_proxy_unlabeled = y_proxy[~labeled_mask]
    return y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, labeled_mask
```

Each call site delegates the split to the helper and validates on the returned mask, so `~np.isnan(y_true)` is computed exactly once:

```python
# Before (ppi.py / ptd.py _preprocess)
labeled_mask = ~np.isnan(y_true_all)
_validate_sample_sizes(labeled_mask)
y_true = y_true_all[labeled_mask]
y_proxy_labeled = y_proxy_all[labeled_mask]
y_proxy_unlabeled = y_proxy_all[~labeled_mask]
return y_true, y_proxy_labeled, y_proxy_unlabeled

# After
y_true, y_proxy_labeled, y_proxy_unlabeled, labeled_mask = _split_labeled_unlabeled(y_true_all, y_proxy_all)
_validate_sample_sizes(labeled_mask)
return y_true, y_proxy_labeled, y_proxy_unlabeled
```

```python
# Before (stratified_core._preprocess, inside the per-stratum loop)
labeled_mask = ~np.isnan(stratum_y_true)
_validate_y_proxy(stratum_y_proxy, stratum_id)
_validate_sample_sizes(labeled_mask, stratum_id)
y_true_filtered = stratum_y_true[labeled_mask]
y_proxy_labeled = stratum_y_proxy[labeled_mask]
y_proxy_unlabeled = stratum_y_proxy[~labeled_mask]
strata.append((y_true_filtered, y_proxy_labeled, y_proxy_unlabeled))

# After
y_true_filtered, y_proxy_labeled, y_proxy_unlabeled, labeled_mask = _split_labeled_unlabeled(
    stratum_y_true, stratum_y_proxy
)
_validate_y_proxy(stratum_y_proxy, stratum_id)
_validate_sample_sizes(labeled_mask, stratum_id)
strata.append((y_true_filtered, y_proxy_labeled, y_proxy_unlabeled))
```

All three call sites (`ppi.py`, `ptd.py`, `stratified_core.py`) import the helper from `glide.estimators.core`. Consumers that do not need the mask (e.g. `PPIMeanMonitor`) drop it with `*arrays, _ = _split_labeled_unlabeled(...)`.

**Files affected:**

| File | What changes |
|---|---|
| `glide/estimators/core.py` | New module holding `_split_labeled_unlabeled`. |
| `glide/estimators/ppi.py` | `_preprocess` calls the helper (imported from `core`) instead of splitting inline. |
| `glide/estimators/ptd.py` | `_preprocess` calls the helper (imported from `core`) instead of splitting inline. |
| `glide/estimators/stratified_core.py` | Per-stratum split delegates to the helper; import it from `core`. |
| `tests/unit/estimators/test_core.py` | New test module with a unit test for `_split_labeled_unlabeled`. |

## Corner cases

- The helper must preserve the exact ordering of the existing inline split: labeled entries (in original order) for `y_true_labeled`/`y_proxy_labeled`, unlabeled entries for `y_proxy_unlabeled`. The existing estimator tests assert numerical results and will catch any reordering.
- The `_validate_sample_sizes` and `_validate_y_proxy` calls stay at the call sites and run **after** the split: `_validate_sample_sizes` is called with the `labeled_mask` the helper returns, so the split must come first. The helper is a pure partition primitive and never raises; it always produces a valid partition regardless of the input.

## Acceptance criteria

- [ ] `_split_labeled_unlabeled` lives in a new `glide/estimators/core.py` and is imported from there by `ppi.py`, `ptd.py` and `stratified_core.py`, with no remaining inline copy of the NaN split in any of them and no import of it from `ppi_core.py`.
- [ ] The existing estimator unit and functional tests pass unchanged.
- [ ] `_split_labeled_unlabeled` has its own unit test in `tests/unit/estimators/test_core.py`, asserting the three returned arrays and the returned `labeled_mask` against hardcoded expected values for an input with both labeled and unlabeled entries.

imerad:
### Description

- Extracts the repeated NaN-based labeled/unlabeled split (`labeled_mask = ~np.isnan(y_true)`) from `PPIMeanEstimator._preprocess`, `PTDMeanEstimator._preprocess`, and `stratified_core._preprocess` into a single shared helper `_split_labeled_unlabeled` in a new `glide/estimators/core.py`. The helper was also propagated to `MultiPPIMeanEstimator` and `ClusteredPPIMeanEstimator` (via `clustered_core.py`), and delegation tests were added for all call sites.
- Closes #312

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/estimators/clustered_core.py
+++ b/glide/estimators/clustered_core.py
@@ -10,6 +10,7 @@ from glide.core.validation import (
     _validate_unique_clusters,
     _validate_y_true,
 )
+from glide.estimators.core import _split_labeled_unlabeled
 
 
 def _preprocess(
@@ -22,7 +23,7 @@ def _preprocess(
     _validate_has_no_nan(y_proxy, "y_proxy")
     _validate_has_no_nan(clusters, "clusters")
 
-    labeled_mask = ~np.isnan(y_true)
+    y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, labeled_mask = _split_labeled_unlabeled(y_true, y_proxy)
     labeled_clusters = clusters[labeled_mask]
     unlabeled_clusters = clusters[~labeled_mask]
 
@@ -47,9 +48,9 @@ def _preprocess(
         error_message=f"Need at least 2 fully unlabeled clusters; got {n_unlabeled_clusters}.",
     )
 
-    labeled_true_sums = np.bincount(labeled_cluster_indices, weights=y_true[labeled_mask])
-    labeled_proxy_sums = np.bincount(labeled_cluster_indices, weights=y_proxy[labeled_mask])
-    unlabeled_proxy_sums = np.bincount(unlabeled_cluster_indices, weights=y_proxy[~labeled_mask])
+    labeled_true_sums = np.bincount(labeled_cluster_indices, weights=y_true_labeled)
+    labeled_proxy_sums = np.bincount(labeled_cluster_indices, weights=y_proxy_labeled)
+    unlabeled_proxy_sums = np.bincount(unlabeled_cluster_indices, weights=y_proxy_unlabeled)
     labeled_sizes = np.bincount(labeled_cluster_indices)
     unlabeled_sizes = np.bincount(unlabeled_cluster_indices)
 
--- a/glide/estimators/core.py
+++ b/glide/estimators/core.py
@@ -1,0 +1,12 @@
+from typing import Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def _split_labeled_unlabeled(y_true: NDArray, y_proxy: NDArray) -> Tuple[NDArray, NDArray, NDArray, NDArray]:
+    labeled_mask = ~np.isnan(y_true)
+    y_true_labeled = y_true[labeled_mask]
+    y_proxy_labeled = y_proxy[labeled_mask]
+    y_proxy_unlabeled = y_proxy[~labeled_mask]
+    return y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, labeled_mask
--- a/glide/estimators/multi_ppi.py
+++ b/glide/estimators/multi_ppi.py
@@ -1,7 +1,6 @@
 from math import floor
 from typing import Tuple
 
-import numpy as np
 from numpy.typing import NDArray
 
 from glide.confidence_intervals import CLTConfidenceInterval
@@ -12,6 +11,7 @@ from glide.core.validation import (
     _validate_y_true,
 )
 from glide.estimators.classical import ClassicalMeanEstimator
+from glide.estimators.core import _split_labeled_unlabeled
 from glide.estimators.multi_ppi_core import (
     _compute_mean_estimate,
     _compute_std_estimate,
@@ -64,11 +64,10 @@ class MultiPPIMeanEstimator:
         _validate_equal_lengths(y_true_all, y_proxies_all, names=["y_true", "y_proxies"])
         _validate_y_proxies(y_proxies_all)
         _validate_y_true(y_true_all)
-        labeled_mask = ~np.isnan(y_true_all)
+        y_true, y_proxies_labeled, y_proxies_unlabeled, labeled_mask = _split_labeled_unlabeled(
+            y_true_all, y_proxies_all
+        )
         _validate_sample_sizes(labeled_mask)
-        y_true = y_true_all[labeled_mask]
-        y_proxies_labeled = y_proxies_all[labeled_mask]
-        y_proxies_unlabeled = y_proxies_all[~labeled_mask]
         return y_true, y_proxies_labeled, y_proxies_unlabeled
 
     def estimate(
--- a/glide/estimators/ppi.py
+++ b/glide/estimators/ppi.py
@@ -1,7 +1,6 @@
 from math import floor
 from typing import Tuple
 
-import numpy as np
 from numpy.typing import NDArray
 
 from glide.confidence_intervals import CLTConfidenceInterval
@@ -12,6 +11,7 @@ from glide.core.validation import (
     _validate_y_true,
 )
 from glide.estimators.classical import ClassicalMeanEstimator
+from glide.estimators.core import _split_labeled_unlabeled
 from glide.estimators.ppi_core import (
     _compute_mean_estimate,
     _compute_std_estimate,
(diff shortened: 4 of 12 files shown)


imerad:
## Code review — `ref/labeled_unlabeled_split`

The extraction of `_split_labeled_unlabeled` is clean and the refactoring is mechanically correct — no crashes or wrong results introduced. A few structural issues worth addressing before merge.

---

### Issues

**1. Validate-before-work order inverted in `stratified_core.py`**

`_split_labeled_unlabeled` now runs *before* `_validate_y_proxy(stratum_y_proxy, stratum_id)` inside the per-stratum loop (line 31). The global `_validate_y_proxy(y_proxy)` on line 23 does not cover per-stratum constancy — a `y_proxy` that is non-constant globally can still be constant within a single stratum. So the per-stratum check is not redundant, and the split now executes before it fires.

Old ordering: compute trivial `labeled_mask` → validate → split  
New ordering: split → validate

Every other `_preprocess` in this codebase validates before consuming. The fix is to call `_validate_y_proxy` before `_split_labeled_unlabeled` in the loop.

---

**2. `_split_labeled_unlabeled` belongs in `glide/core/`, not `glide/estimators/core.py`**

`glide/core/` is the established home for shared, cross-cutting helpers (per CLAUDE.md: *"Centralized parameter validation helpers … Import from here rather than duplicating checks across modules"*). Placing this helper in `glide/estimators/` causes two problems:

- `glide/core/validation.py` already inlines `labeled = y_true[~np.isnan(y_true)]` independently (line 36) — partial duplication that stays permanent because `glide/core/` cannot import from `glide/estimators/` without a circular dependency.
- The name `glide/estimators/core.py` collides conceptually with `glide/core/`, creating two "core" namespaces with different purposes.

Suggested placement: `glide/core/data.py` or alongside `validation.py`.

---

**3. Missing 2D test case in `test_core.py`**

`multi_ppi._preprocess` calls `_split_labeled_unlabeled(y_true_all, y_proxies_all)` where `y_proxies_all` has shape `(n, k)`. `test_core.py` has a single test covering only the 1D case. CLAUDE.md: *"Each distinct scenario (input combination, edge case, or error condition) gets its own test function."*

The 2D path works correctly today (NumPy row-indexing), but it is not covered by unit tests — delegation tests for `MultiPPIMeanEstimator` mock out `_split_labeled_unlabeled`, and functional tests are excluded from the 100% coverage requirement.

A `test_split_labeled_unlabeled_2d` test with a `(4, 2)` `y_proxy` would cover this.

---

**4. `labeled_mask` returned by `_split_labeled_unlabeled` creates an implicit caller obligation**

In `ppi.py`, `ptd.py`, and `multi_ppi.py`, `labeled_mask` is the 4th return value and is used only to call `_validate_sample_sizes` before being discarded. This makes sample-size validation an invisible caller responsibility — a future estimator can call `_split_labeled_unlabeled` without the follow-up and silently skip the check. `clustered_core.py` already omits `_validate_sample_sizes` (intentionally, with its own cluster-count checks), making it impossible to distinguish intentional omissions from accidental ones.

Options: move the common `_validate_sample_sizes(labeled_mask)` call inside `_split_labeled_unlabeled` for the base case, or document the obligation explicitly.

---

### Minor

- `test_stratified_core.py:46` (pre-existing): `mock_validate_equal_lengths.assert_called_once_with(y_true, y_proxy, groups, ...)` passes NumPy arrays to `assert_called_once_with`. CLAUDE.md requires `assert_called_once()` followed by `np.testing.assert_array_equal` for each array argument.
- `test_stratified_core.py:77` (pre-existing): `assert len(strata) == 2` — CLAUDE.md prefers asserting exact contents over aggregate length checks.

gmartinon-ed (changes requested):
Mon seul feedback :

les fonctions de test `test_preprocess_delegates_to_validation` ont des noms un peu misleading maintenant, car elles ne délèguent plus qu'à des fonctions de validation. Donc peut-être juste les renommer `test_preprocess_delegates`

gmartinon-ed (approved):

PR #331, opened 2026-07-02
Add empirical-Bernstein confidence sequence

### LINKED ISSUE #314: ConfidenceSequence: anytime-valid bounds for drift monitoring
# TODO

- Functions or methods ?
- Le confidence_level ne sert à rien ! A supprimer dans le ticket et dans les classes, c'est au niveau des monitors que le confidence_level se gère
- Renommer `running_means` en `running_mean_estimates`

## Background

A *confidence interval* bounds a single estimate; a *confidence sequence* bounds a whole sequence of estimates simultaneously, so a user may inspect it after every new batch of data without the false-alarm probability creeping up over time (the "peeking" problem). It is the statistical object the drift monitors rely on: each computes a per-batch estimate of the quality metric, and the sequence delivers an anytime-valid bound on the *running mean* of those estimates that holds at all looks at once.

The bound used is the *empirical-Bernstein* sequence: a concentration bound whose width adapts to the empirically observed variability of the estimates instead of assuming a known variance. In math notation this is written as a deviation budget `u(v)` at accumulated variance `v`, the unique root in `s` of `∫₀¹ exp(β·s − ψ_E(β)·v) dβ = 1/δ` with `ψ_E(β) = −log(1−β) − β`, found numerically because the uniform mixture has no closed form. This package builds and tests that machinery in isolation, ahead of any monitor, mirroring how `glide/confidence_intervals/` is a standalone layer the estimators depend on.

## Design choices

The package `glide/confidence_sequences/` mirrors `glide/confidence_intervals/`: a `ConfidenceSequence` structural protocol in `base.py` (the time-uniform analogue of the `ConfidenceInterval` protocol) and one concrete `EmpiricalBernsteinConfidenceSequence` in `empirical_bernstein.py` (as `CLTConfidenceInterval`/`BootstrapConfidenceInterval` implement the interval protocol). The value object is **passive**: it holds the parallel `running_means`/`confidence_bounds` arrays already in original metric units, exactly as `CLTConfidenceInterval` holds final bounds rather than raw data, and exposes `test_null_hypothesis(h0_value, alternative)` returning the per-look boolean alarm vector. The heavy sequence math (running mean, variance process, boundary root-finding, harmful-side bound) lives in one free function `_compute_empirical_bernstein_bounds` that the monitors call before constructing the value object, just as an estimator computes a mean and standard error before handing them to a passive interval. Keeping the math in a free function (rather than the constructor) is what lets each monitor do its own unit↔metric mapping around it: that mapping is method-specific (the classical monitor un-maps risk space, the prediction-powered monitor un-maps an extra affine normalization first), so the value object only ever sees final metric-oriented arrays. The empirical-Bernstein boundary helpers (`_compute_mixture_wealth`, `_compute_mixture_boundary`) live here too, as the sequence's own internals. `test_null_hypothesis` supports only the two directional alternatives (`"larger"` for a risk, `"smaller"` for a performance) and raises on `"two-sided"`, a deliberate, documented divergence from the interval protocol's two-sided default, because the underlying sequence is one-sided. Like the interval protocol, the sequence is parameterized by its `confidence_level`: under no drift, the probability it *ever* fails to cover the running mean is at most `1 - confidence_level`.

### ConfidenceSequence protocol

**Location:** `glide/confidence_sequences/base.py`

**Purpose:** Structural protocol any sequence implementation satisfies, so result objects can hold one without depending on a concrete class. Mirrors `glide/confidence_intervals/base.py`.

```python
from typing import Literal, Protocol

class ConfidenceSequence(Protocol):
    """Structural protocol for anytime-valid confidence sequences.

    The time-uniform analogue of :class:`ConfidenceInterval`: instead of a single
    bound it carries a bound per look (one per batch), valid simultaneously at all
    looks. Any class implementing this protocol can be used as a
    ``confidence_sequence`` in result objects like ``MeanMonitoringResult``.
    """

    confidence_level: float
    running_means: NDArray
    confidence_bounds: NDArray

    def test_null_hypothesis(
        self,
        h0_value: float,
        alternative: Literal["larger", "smaller"] = "larger",
    ) -> NDArray:
        """Test the running mean against a value at every look.

        Returns the boolean per-look alarm vector, with time-uniform family-wise
        error control over all looks.
        """
        ...
```

### EmpiricalBernsteinConfidenceSequence and the shared builder

**Location:** `glide/confidence_sequences/empirical_bernstein.py`

**Purpose:** The empirical-Bernstein boundary internals, the shared sequence builder the monitors call, and the passive value object the monitoring results hold.

The boundary `_compute_mixture_boundary` has two facts the unit tests anchor on: it depends only on the variance process and `miscoverage`, and at variance process `0` the integral collapses to the closed form `(exp(s) − 1) / s`, so the returned `u` satisfies `(np.exp(u) - 1) / u == 1 / miscoverage`. The builder `_compute_empirical_bernstein_bounds` is space-agnostic: the caller supplies per-batch estimates and a predictable seed center already in `[0, 1]` (risk space for the classical monitor, affine-normalized space for the prediction-powered one), and it returns the parallel running-mean and one-sided harmful-side bound arrays in that same space.

```python
from dataclasses import dataclass, field
from typing import Literal, Tuple

import numpy as np
from scipy.integrate import quad
from scipy.optimize import brentq

from glide.core.validation import _validate_bounds, _validate_literal


def _compute_mixture_wealth(deviation: float, variance_process: float) -> float:
    # Mixture (uniform density on (0, 1)) of the betting wealth
    # exp(β·deviation − ψ_E(β)·variance_process), with ψ_E(β) = −log(1−β) − β. This
    # integral is a nonnegative supermartingale under the no-drift null, so by Ville's
    # inequality it ever reaches 1/δ with probability at most δ.
    def integrand(betting_parameter: float) -> float:
        psi_exponential = -np.log1p(-betting_parameter) - betting_parameter
        value = np.exp(betting_parameter * deviation - psi_exponential * variance_process)
        return value

    integral, _ = quad(integrand, 0.0, 1.0)
    return integral


def _compute_mixture_boundary(variance_process: float, miscoverage: float) -> float:
    # Solve _compute_mixture_wealth(u, variance_process) == 1 / miscoverage for the
    # deviation budget u: the edge of the acceptance region. brentq is a bracketing
    # solver, so it needs an interval on which excess_wealth changes sign. The lower
    # end u = 0 is always negative (the wealth there is <= 1 < 1 / miscoverage). The
    # root moves outward with variance_process and 1 / miscoverage, so no fixed upper
    # end works: double from 1 until excess_wealth is non-negative, which must happen
    # because the wealth grows without bound in u.
    wealth_target = 1.0 / miscoverage

    def excess_wealth(deviation: float) -> float:
        value = _compute_mixture_wealth(deviation, variance_process) - wealth_target
        return value

    upper_bracket = 1.0
    while excess_wealth(upper_bracket) < 0.0:
        upper_bracket *= 2.0
    boundary = brentq(excess_wealth, 0.0, upper_bracket)
    return boundary


def _compute_empirical_bernstein_bounds(
    batch_estimates: NDArray,
    seed_center: float,
    miscoverage: float,
) -> Tuple[NDArray, NDArray]:
    # The shared confidence-sequence math the monitors call. Operates in whatever
    # bounded space the caller supplies: per-batch estimates and seed_center must
    # already lie in [0, 1].
    n_batches = len(batch_estimates)
    batch_counts = np.arange(1, n_batches + 1)
    running_means = np.cumsum(batch_estimates) / batch_counts
    # The variance process accumulates squared deviations from a *predictable* center
    # (known before each batch arrives): the previous running mean, seeded with
    # seed_center before the first batch.
    predictable_centers = np.hstack([np.array([seed_center]), running_means[:-1]])
    variance_process = np.cumsum((batch_estimates - predictable_centers) ** 2)
    # Each boundary value requires scalar root-finding, hence the loop.
    boundaries = np.array([_compute_mixture_boundary(value, miscoverage) for value in variance_process])
    lower_bounds = running_means - boundaries / batch_counts
    return running_means, lower_bounds


@dataclass
class EmpiricalBernsteinConfidenceSequence:
    """Anytime-valid empirical-Bernstein confidence sequence on a running mean.

    Holds the per-look running means and the one-sided anytime-valid bound on the
    side where drift is harmful (a lower bound for a risk, an upper bound for a
    performance, after the monitor has mapped the sequence back to the original
    metric orientation). The bounds hold simultaneously at all looks, so testing
    after every batch does not inflate the false-alarm probability.

    Parameters
    ----------
    running_means : NDArray
        Per-look running mean of the per-batch estimates, in original metric units.
    confidence_bounds : NDArray
        Per-look harmful-side anytime-valid bound, in original metric units.
        Confidence level of the sequence: under no drift, the probability that it
        ever fails to cover the running mean is at most ``1 - confidence_level``.
        Default is 0.8.

    References
    ----------
    Waudby-Smith, Ian, and Aaditya Ramdas. "Estimating means of bounded random
    variables by betting." Journal of the Royal Statistical Society Series B:
    Statistical Methodology 86, no. 1 (2024): 1-27.

    no. 2 (2021): 1055-1080.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
    >>> sequence = EmpiricalBernsteinConfidenceSequence(
    ...     running_means=np.array([0.4, 0.6]),
    ...     confidence_bounds=np.array([0.1, 0.55]),
    ...     confidence_level=0.8,
    ... )
    >>> sequence.test_null_hypothesis(0.5, alternative="larger")
    array([False,  True])
    """

    running_means: NDArray
    confidence_bounds: NDArray
    _confidence_level: float = field(init=False, repr=False)

    def __init__(self, running_means: NDArray, confidence_bounds: NDArray, confidence_level: float = 0.8) -> None:
        self.running_means = running_means
        self.confidence_bounds = confidence_bounds
        self.confidence_level = confidence_level

    @property
    def confidence_level(self) -> float:
        return self._confidence_level

    @confidence_level.setter
    def confidence_level(self, value: float) -> None:
        _validate_bounds(value, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False)
        self._confidence_level = value

    def test_null_hypothesis(
        self,
        h0_value: float,
        alternative: Literal["larger", "smaller"] = "larger",
    ) -> NDArray:
        """Test the running mean against ``h0_value`` at every look.

        Parameters
        ----------
        h0_value : float
            The threshold the harmful-side bound is tested against (for a monitor,
            the user-supplied business threshold).
        alternative : str, optional
            ``'larger'`` (default) when the metric is a risk: alarm where the lower
            bound exceeds ``h0_value``. ``'smaller'`` when it is a performance: alarm
            where the upper bound falls below ``h0_value``. A confidence sequence is
            one-sided, so ``'two-sided'`` is not accepted.

        Returns
        -------
        NDArray
            Boolean per-look alarm vector, ``True`` once the bound has crossed
            ``h0_value``. Time-uniform family-wise error is controlled at
            ``1 - confidence_level`` over all looks.

        Raises
        ------
        ValueError
            If ``alternative`` is not ``'larger'`` or ``'smaller'``.

        Examples
        --------
        >>> import numpy as np
        >>> from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
        >>> sequence = EmpiricalBernsteinConfidenceSequence(
        ...     running_means=np.array([0.4, 0.6]),
        ...     confidence_bounds=np.array([0.1, 0.55]),
        ... )
        >>> sequence.test_null_hypothesis(0.5, alternative="larger")
        array([False,  True])
        """
        alternatives = ["larger", "smaller"]
        _validate_literal(alternative, "alternative", alternatives)
        if alternative == alternatives[0]:
            alarms = self.confidence_bounds > h0_value
        else:
            alarms = self.confidence_bounds < h0_value
        return alarms
```

### Package exports

**Location:** `glide/confidence_sequences/__init__.py`

```python
from glide.confidence_sequences.empirical_bernstein import EmpiricalBernsteinConfidenceSequence

__all__ = [
    "ConfidenceSequence",
    "EmpiricalBernsteinConfidenceSequence",
]
```

Unit tests mirror `tests/unit/confidence_intervals/` (no test for the protocol, one file per concrete class): `tests/unit/confidence_sequences/test_empirical_bernstein.py` covers the boundary helpers, the builder and the value object. As with the interval layer, `--doctest-modules` collects the docstring examples.

## Functional tests

**Boundary analytic anchor at zero variance process**

At variance process `0` the uniform-mixture wealth has the closed form `(exp(s) − 1) / s`, so the returned boundary `u` must invert it exactly.

```python
def test_compute_mixture_boundary_zero_variance(miscoverage):
    boundary = _compute_mixture_boundary(0.0, miscoverage)

    assert (np.exp(boundary) - 1) / boundary == pytest.approx(1 / miscoverage)
```

**Boundary monotonicity**

The deviation budget grows with the accumulated variance and with `1 / miscoverage` (a smaller miscoverage gives a wider budget).

```python
def test_compute_mixture_boundary_increases_with_variance():
    small = _compute_mixture_boundary(0.5, 0.1)
    large = _compute_mixture_boundary(2.0, 0.1)

    assert small < large


def test_compute_mixture_boundary_increases_with_confidence():
    loose = _compute_mixture_boundary(0.5, 0.2)
    tight = _compute_mixture_boundary(0.5, 0.05)

    assert loose < tight
```

**Builder reproduces the running mean and a shrinking bound**

On a hardcoded sequence, `_compute_empirical_bernstein_bounds` returns the cumulative mean and a harmful-side bound that lies strictly below it. Both arrays are asserted against exact hardcoded values (the expected `lower_bounds` are obtained from a correct first run of the implementation and then fixed in the test).

```python
def test_compute_empirical_bernstein_bounds():
    batch_estimates = np.array([0.4, 0.6, 0.5])
    running_means, lower_bounds = _compute_empirical_bernstein_bounds(batch_estimates, seed_center=0.5, miscoverage=0.2)

    np.testing.assert_allclose(running_means, np.array([0.4, 0.5, 0.5]))
    np.testing.assert_allclose(lower_bounds, EXPECTED_LOWER_BOUNDS, atol=1e-6)
```

The `running_means` are derived analytically from the inputs (cumulative mean of `[0.4, 0.6, 0.5]`). The `EXPECTED_LOWER_BOUNDS` array is the one quantity the implementer must capture from a first correct run of `_compute_empirical_bernstein_bounds` (the boundary has no closed form, so the values cannot be hand-derived) and then hardcode in the test, exactly as the `CLTConfidenceInterval` ±1.96 anchor is hardcoded.

The empirical-Bernstein coverage guarantee itself (type-I control under no drift, power under a shift) is validated end-to-end through the monitor Monte Carlo tests in the monitor tickets, not re-simulated here.

## Corner cases

- `_compute_empirical_bernstein_bounds` on a single look returns one running mean equal to that estimate and one bound; with estimates in `[0, 1]` and a modest miscoverage the bound can fall below `0`, which is expected finite-sample safety (a very wide early bound), not a bug.
- `EmpiricalBernsteinConfidenceSequence` with `confidence_level` outside `(0, 1)` → `ValueError` from the setter, with the bad value in the message.
- `test_null_hypothesis` with `alternative="two-sided"` (or any value other than `"larger"`/`"smaller"`) → `ValueError`; the sequence is one-sided.
- `test_null_hypothesis` returns an all-`False` vector when no bound crosses `h0_value`, and flips to `True` from the first crossing onward only if the (monotone) bound stays across; it does not assume monotonicity, so a bound that crosses and retreats yields `True` only at the crossing looks.

## Acceptance criteria

- [ ] `glide/confidence_sequences/` exists with `base.py` (the `ConfidenceSequence` protocol), `empirical_bernstein.py` and `__init__.py`, mirroring `glide/confidence_intervals/`.
- [ ] `_compute_mixture_boundary(0.0, miscoverage)` satisfies `(np.exp(u) - 1) / u == pytest.approx(1 / miscoverage)`, and is strictly increasing in both the variance process and `1 / miscoverage` (unit tests with hardcoded orderings).
- [ ] `_compute_empirical_bernstein_bounds` returns the cumulative running mean and a harmful-side bound strictly below it on a hardcoded sequence (asserted against expected values).
- [ ] `EmpiricalBernsteinConfidenceSequence.test_null_hypothesis` returns the boolean per-look alarm vector for `"larger"` and `"smaller"` (unit tests against hardcoded bounds and threshold) and raises `ValueError` on `"two-sided"`.
- [ ] The boundary helpers and the builder live only in `glide/confidence_sequences/empirical_bernstein.py`; no monitor module redefines them (verified once the monitors land).
- [ ] `ConfidenceSequence` and `EmpiricalBernsteinConfidenceSequence` are importable from `glide.confidence_sequences`.

imerad:
### Description

- Adds the `glide/confidence_sequences/` package with a `ConfidenceSequence` structural protocol and `EmpiricalBernsteinConfidenceSequence`: an anytime-valid one-sided bound on a running mean that holds simultaneously at all looks, with no false-alarm inflation from peeking. The package mirrors `glide/confidence_intervals/` in structure and serves as the foundational layer the drift monitors will depend on.
- Closes #314

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/confidence_sequences/empirical_bernstein.py
+++ b/glide/confidence_sequences/empirical_bernstein.py
@@ -1,0 +1,133 @@
+from dataclasses import dataclass
+from typing import Literal, Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+from scipy.integrate import quad
+from scipy.optimize import brentq
+
+from glide.core.validation import _validate_literal
+
+
+def _compute_mixture_wealth(deviation: float, variance_process: float) -> float:
+    # ψ_E(β) = −log(1−β) − β is the cumulant of the exponential-family mixture.
+    def integrand(betting_parameter: float) -> float:
+        psi_exponential = -np.log1p(-betting_parameter) - betting_parameter
+        value = np.exp(betting_parameter * deviation - psi_exponential * variance_process)
+        return value
+
+    integral, _ = quad(integrand, 0.0, 1.0)
+    return integral
+
+
+def _compute_mixture_boundary(variance_process: float, miscoverage: float) -> float:
+    wealth_target = 1.0 / miscoverage
+
+    def excess_wealth(deviation: float) -> float:
+        value = _compute_mixture_wealth(deviation, variance_process) - wealth_target
+        return value
+
+    # No fixed upper bracket works because the root grows with variance_process and
+    # 1/miscoverage; double until excess_wealth turns non-negative.
+    upper_bracket = 1.0
+    while excess_wealth(upper_bracket) < 0.0:
+        upper_bracket *= 2.0
+    boundary = brentq(excess_wealth, 0.0, upper_bracket)
+    return boundary
+
+
+def _compute_empirical_bernstein_bounds(
+    batch_estimates: NDArray,
+    seed_center: float,
+    miscoverage: float,
+) -> Tuple[NDArray, NDArray]:
+    n_batches = len(batch_estimates)
+    batch_counts = np.arange(1, n_batches + 1)
+    running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
+    # Centers must be predictable (known before each batch arrives): use the previous
+    # running mean, seeded with seed_center before the first batch.
+    predictable_centers = np.hstack([np.array([seed_center]), running_mean_estimates[:-1]])
+    variance_process = np.cumsum((batch_estimates - predictable_centers) ** 2)
+    boundaries = np.array([_compute_mixture_boundary(value, miscoverage) for value in variance_process])
+    lower_bounds = running_mean_estimates - boundaries / batch_counts
+    return running_mean_estimates, lower_bounds
+
+
+@dataclass
+class EmpiricalBernsteinConfidenceSequence:
+    """Anytime-valid empirical-Bernstein confidence sequence on a running mean.
+
+    Holds the per-look running means and the one-sided anytime-valid bound on the
+    side where drift is harmful (a lower bound for a risk, an upper bound for a
+    performance, after the monitor has mapped the sequence back to the original
+    metric orientation). The bounds hold simultaneously at all looks, so testing
+    after every batch does not inflate the false-alarm probability.
+
+    Parameters
+    ----------
+    running_mean_estimates : NDArray
+        Per-look running mean of the per-batch estimates, in original metric units.
+    confidence_bounds : NDArray
+        Per-look harmful-side anytime-valid bound, in original metric units.
+
+    References
+    ----------
+    Waudby-Smith, Ian, and Aaditya Ramdas. "Estimating means of bounded random
+    variables by betting." Journal of the Royal Statistical Society Series B:
+    Statistical Methodology 86, no. 1 (2024): 1-27.
+
+    Howard, Steven R., Aaditya Ramdas, Jon McAuliffe, and Jasjeet Sekhon. "Time-uniform,
+    nonparametric, nonasymptotic confidence sequences." The Annals of Statistics 49,
+    no. 2 (2021): 1055-1080.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
+    >>> sequence = EmpiricalBernsteinConfidenceSequence(
+    ...     running_mean_estimates=np.array([0.4, 0.6]),
+    ...     confidence_bounds=np.array([0.1, 0.55]),
+    ... )
+    >>> sequence.test_null_hypothesis(0.5, alternative="larger")
+    array([False,  True])
+    """
+
+    running_mean_estimates: NDArray
+    confidence_bounds: NDArray
+
+    def test_null_hypothesis(
+        self,
+        h0_value: float,
+        alternative: Literal["larger", "smaller"] = "larger",
+    ) -> NDArray:
+        """Test the running mean against ``h0_value`` at every look.
+
+        Parameters
+        ----------
+        h0_value : float
+            The threshold the harmful-side bound is tested against (for a monitor,
+            the user-supplied business threshold).
+        alternative : str, optional
+            ``'larger'`` (default) when the metric is a risk: alarm where the lower
+            bound exceeds ``h0_value``. ``'smaller'`` when it is a performance: alarm
+            where the upper bound falls below ``h0_value``. A confidence sequence is
+            one-sided, so ``'two-sided'`` is not accepted.
+
+        Returns
+        -------
+        NDArray
+            Boolean per-look alarm vector, ``True`` once the bound has crossed
+            ``h0_value``. Time-uniform family-wise error is controlled over all looks.
+
+        Raises
+        ------
+        ValueError
+            If ``alternative`` is not ``'larger'`` or ``'smaller'``.
+        """
+        alternatives = ["larger", "smaller"]
+        _validate_literal(alternative, "alternative", alternatives)
+        if alternative == alternatives[0]:
+            alarms = self.confidence_bounds > h0_value
+        else:
+            alarms = self.confidence_bounds < h0_value
+        return alarms
--- a/glide/confidence_sequences/empirical_bernstein.py
+++ b/glide/confidence_sequences/empirical_bernstein.py
@@ -10,7 +10,6 @@ from glide.core.validation import _validate_bounds, _validate_literal
 
 
 def _compute_mixture_wealth(deviation: float, variance_process_value: float) -> float:
-    # ψ_E(β) = −log(1−β) − β is the cumulant of the exponential-family mixture.
     def integrand(betting_parameter: float) -> float:
         psi_exponential = -np.log(1 - betting_parameter) - betting_parameter
         value = np.exp(betting_parameter * deviation - psi_exponential * variance_process_value)
@@ -27,8 +26,6 @@ def _compute_mixture_boundary(variance_process_value: float, miscoverage: float)
         value = _compute_mixture_wealth(deviation, variance_process_value) - wealth_target
         return value
 
-    # No fixed upper bracket works because the root grows with variance_process_value and
-    # 1/miscoverage; double until excess_wealth turns non-negative.
     upper_bracket = 1.0
     while excess_wealth(upper_bracket) < 0.0:
         upper_bracket *= 2.0
@@ -45,8 +42,6 @@ def _compute_empirical_bernstein_bounds(
     n_batches = len(batch_estimates)
     batch_counts = np.arange(1, n_batches + 1)
     running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
-    # Centers must be predictable (known before each batch arrives): use the previous
-    # running mean, seeded with seed_center before the first batch.
     predictable_centers = np.hstack([np.array([seed_center]), running_mean_estimates[:-1]])
     variance_process = np.cumsum((batch_estimates - predictable_centers) ** 2)
     boundaries = np.array([_compute_mixture_boundary(value, miscoverage) for value in variance_process])
--- a/glide/confidence_sequences/empirical_bernstein.py
+++ b/glide/confidence_sequences/empirical_bernstein.py
@@ -6,7 +6,7 @@ from numpy.typing import NDArray
 from scipy.integrate import quad
 from scipy.optimize import brentq
 
-from glide.core.validation import _validate_literal
+from glide.core.validation import _validate_bounds, _validate_literal
 
 
 def _compute_mixture_wealth(deviation: float, variance_process_value: float) -> float:
@@ -41,6 +41,7 @@ def _compute_empirical_bernstein_bounds(
     seed_center: float,
     miscoverage: float,
 ) -> Tuple[NDArray, NDArray]:
+    _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=1.0, left_inclusive=False, right_inclusive=False)
     n_batches = len(batch_estimates)
     batch_counts = np.arange(1, n_batches + 1)
     running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
--- a/glide/confidence_sequences/empirical_bernstein.py
+++ b/glide/confidence_sequences/empirical_bernstein.py
@@ -9,25 +9,25 @@ from scipy.optimize import brentq
 from glide.core.validation import _validate_literal
 
 
-def _compute_mixture_wealth(deviation: float, variance_process: float) -> float:
+def _compute_mixture_wealth(deviation: float, variance_process_value: float) -> float:
     # ψ_E(β) = −log(1−β) − β is the cumulant of the exponential-family mixture.
     def integrand(betting_parameter: float) -> float:
-        psi_exponential = -np.log1p(-betting_parameter) - betting_parameter
-        value = np.exp(betting_parameter * deviation - psi_exponential * variance_process)
+        psi_exponential = -np.log(1 - betting_parameter) - betting_parameter
+        value = np.exp(betting_parameter * deviation - psi_exponential * variance_process_value)
         return value
 
     integral, _ = quad(integrand, 0.0, 1.0)
     return integral
 
 
-def _compute_mixture_boundary(variance_process: float, miscoverage: float) -> float:
+def _compute_mixture_boundary(variance_process_value: float, miscoverage: float) -> float:
     wealth_target = 1.0 / miscoverage
 
     def excess_wealth(deviation: float) -> float:
-        value = _compute_mixture_wealth(deviation, variance_process) - wealth_target
+        value = _compute_mixture_wealth(deviation, variance_process_value) - wealth_target
         return value
 
-    # No fixed upper bracket works because the root grows with variance_process and
+    # No fixed upper bracket works because the root grows with variance_process_value and
     # 1/miscoverage; double until excess_wealth turns non-negative.
     upper_bracket = 1.0
     while excess_wealth(upper_bracket) < 0.0:
--- a/tests/unit/confidence_sequences/test_empirical_bernstein.py
+++ b/tests/unit/confidence_sequences/test_empirical_bernstein.py
@@ -48,6 +48,16 @@ def test_compute_mixture_boundary_increases_with_confidence():
 # --- _compute_empirical_bernstein_bounds ---
 
 
+def test_compute_empirical_bernstein_bounds_delegates_to_validation():
+    batch_estimates = np.array([0.4, 0.6, 0.5])
+    with patch.object(empirical_bernstein_module, "_validate_bounds") as mock_validate_bounds:
+        _compute_empirical_bernstein_bounds(batch_estimates, seed_center=0.5, miscoverage=0.2)
+
+        mock_validate_bounds.assert_called_once_with(
+            0.2, "miscoverage", lower=0.0, upper=1.0, left_inclusive=False, right_inclusive=False
+        )
+
+
 def test_compute_empirical_bernstein_bounds():
     batch_estimates = np.array([0.4, 0.6, 0.5])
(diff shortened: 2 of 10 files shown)


imerad:
## Code review

Reviewed the diff for `glide/confidence_sequences/` (protocol + `EmpiricalBernsteinConfidenceSequence`) and the accompanying test/doc changes. Findings below, most severe first (all verified by reading/running the code).

### 1. `_compute_empirical_bernstein_bounds` is never wired into the public class
`glide/confidence_sequences/empirical_bernstein.py:39-53` computes `running_mean_estimates`/bounds from raw `batch_estimates`, but it's only called from tests (verified via repo-wide grep — zero non-test call sites). `EmpiricalBernsteinConfidenceSequence` (`empirical_bernstein.py:56-96`) is a plain dataclass that requires already-computed `running_mean_estimates`/`confidence_bounds` as constructor args, unlike `CLTConfidenceInterval`/`BootstrapConfidenceInterval` in `glide/confidence_intervals/`, which compute their bounds internally from raw inputs in `__init__`. As shipped there's no public, supported way to go from batch data to a `ConfidenceSequence` instance — a caller would need to import a private, underscore-prefixed function. This also breaks the PR description's claim that the package "mirrors `glide/confidence_intervals/` in structure." Suggest a classmethod (e.g. `from_batch_estimates(...)`) mirroring the CLT/Bootstrap pattern.

### 2. `miscoverage` has no input validation and crashes with unclear errors
`glide/confidence_sequences/empirical_bernstein.py:23-24` (`wealth_target = 1.0 / miscoverage`). Verified directly:
- `miscoverage=0` → bare `ZeroDivisionError`
- `miscoverage=1.5` or `miscoverage=-0.1` → `ValueError: f(a) and f(b) must have different signs` (leaking a scipy internal error)

Neither matches the codebase convention (e.g. `_validate_bounds`) of raising a clear `'miscoverage' must be in (0, 1); got ...`-style message.

### 3. Test coverage regression: "invalid alternative raises ValueError" is no longer tested end-to-end
`tests/unit/confidence_intervals/test_bootstrap.py` and `test_clt.py` each deleted `test_null_hypothesis_invalid_alternative` and replaced it with a `test_null_hypothesis_delegates_to_validation` test that mocks out `_validate_literal` — so the real raise path for `BootstrapConfidenceInterval`/`CLTConfidenceInterval` is no longer exercised anywhere. The same gap exists for the new `EmpiricalBernsteinConfidenceSequence` (`tests/unit/confidence_sequences/test_empirical_bernstein.py`): only the delegation test was added, no direct "raises ValueError on bad alternative" test. Suggest keeping one lightweight unmocked `pytest.raises(ValueError, match=...)` test per class alongside the delegation test.

### 4. Per-look bound computation is expensive and scales with batch count
`glide/confidence_sequences/empirical_bernstein.py:51`. Each look requires a `brentq` root-find (with its own bracket-doubling loop) wrapping a `scipy.integrate.quad` numerical integration, repeated once per batch in the list comprehension. Measured locally: ~0.3s for 500 batches. Worth a look given the use case (a monitor recomputing after every new batch over potentially long-running sequences) — an incremental/vectorized approach or caching across successive calls may be needed before this is used in production monitors.

### 5. Minor: dangling forward reference in docstring
`glide/confidence_sequences/base.py:12` says the protocol can be used "in result objects like `MeanMonitoringResult`", but no such class exists anywhere in the codebase yet (repo-wide grep confirms). Fine if intentionally forward-looking to the drift-monitor work mentioned in the PR description, but worth flagging so it doesn't get missed once that lands.

---
*Generated via automated code review (Claude Code).*

glide/confidence_sequences/empirical_bernstein.py
```diff
@@ -0,0 +1,134 @@
...
+import numpy as np
+from numpy.typing import NDArray
+from scipy.integrate import quad
+from scipy.optimize import brentq
+
+from glide.core.validation import _validate_bounds, _validate_literal
+
+
+def _compute_mixture_wealth(deviation: float, variance_process_value: float) -> float:
+    # ψ_E(β) = −log(1−β) − β is the cumulant of the exponential-family mixture.
```
gmartinon-ed: remove comment

glide/confidence_sequences/empirical_bernstein.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _compute_mixture_boundary(variance_process_value: float, miscoverage: float) -> float:
+    wealth_target = 1.0 / miscoverage
+
+    def excess_wealth(deviation: float) -> float:
+        value = _compute_mixture_wealth(deviation, variance_process_value) - wealth_target
+        return value
+
+    # No fixed upper bracket works because the root grows with variance_process_value and
+    # 1/miscoverage; double until excess_wealth turns non-negative.
```
gmartinon-ed: remove comment

glide/confidence_sequences/empirical_bernstein.py
```diff
@@ -0,0 +1,134 @@
...
+    batch_estimates: NDArray,
+    seed_center: float,
+    miscoverage: float,
+) -> Tuple[NDArray, NDArray]:
+    _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=1.0, left_inclusive=False, right_inclusive=False)
+    n_batches = len(batch_estimates)
+    batch_counts = np.arange(1, n_batches + 1)
+    running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
+    # Centers must be predictable (known before each batch arrives): use the previous
+    # running mean, seeded with seed_center before the first batch.
```
gmartinon-ed: remove comment

tests/unit/confidence_sequences/test_empirical_bernstein.py
```diff
@@ -0,0 +1,103 @@
...
+    assert wealth == pytest.approx(1.227, abs=1e-3)
+
+
+# --- _compute_mixture_boundary ---
+
+
+@pytest.mark.parametrize("miscoverage", [0.1, 0.2, 0.5])
+def test_compute_mixture_boundary_zero_variance(miscoverage):
+    boundary = _compute_mixture_boundary(0.0, miscoverage)
+    assert (np.exp(boundary) - 1) / boundary == pytest.approx(1 / miscoverage, abs=1e-6)
```
gmartinon-ed: Un peu complexe ce assert introduire des variables intermédiaires

tests/unit/confidence_sequences/test_empirical_bernstein.py
```diff
@@ -0,0 +1,103 @@
...
+
+import glide.confidence_sequences.empirical_bernstein as empirical_bernstein_module
+from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
+from glide.confidence_sequences.empirical_bernstein import (
+    _compute_empirical_bernstein_bounds,
+    _compute_mixture_boundary,
+    _compute_mixture_wealth,
+)
+
+EXPECTED_LOWER_BOUNDS = np.array([-2.27458313, -0.86521862, -0.41014575])
```
gmartinon-ed: On s'en sert qu'une seule fois j'ai l'impression, donc autant le rapatrier dans son test

gmartinon-ed (changes requested):

glide/confidence_sequences/empirical_bernstein.py
imerad: Done

glide/confidence_sequences/empirical_bernstein.py
imerad: Done

glide/confidence_sequences/empirical_bernstein.py
imerad: Done

tests/unit/confidence_sequences/test_empirical_bernstein.py
imerad: Fait

tests/unit/confidence_sequences/test_empirical_bernstein.py
imerad: Fait

gmartinon-ed (approved):

PR #333, opened 2026-07-03
Add mean monitoring result dataclasses

### LINKED ISSUE #315: Mean monitoring result dataclasses
# TODO
- Renommer `batch_estimates` en `batch_mean_estimates`

## Background

The drift monitors each return a result object describing a monitoring run over a batched dataset: the per-batch estimates, the running mean of those estimates, the anytime-valid confidence bound on that running mean, the per-batch alarm flags, and the threshold the bound is tested against. This is the monitoring counterpart of the inference results that estimators return (`glide/mean_inference_results/`), and it is extracted into its own ticket so the monitor tickets can focus on the detection logic alone.

The result is a passive data container: it holds a `ConfidenceSequence` (the anytime-valid bound object, ticket `feat-confidence-sequence.md`) plus a few scalar fields, and derives everything else (the per-look running means and bounds, the alarm vector, whether and when a drift was detected) as read-only properties. This mirrors how the inference results hold a `ConfidenceInterval` rather than loose bound arrays.

## Design choices

A new package `glide/mean_monitoring_results/` mirrors the structure of `glide/mean_inference_results/`: a base class in `base.py` holding the shared fields and display logic, and one thin subclass per monitor (`classical.py`, `prediction_powered.py`), just as `MeanInferenceResult` is subclassed by `ClassicalMeanInferenceResult` and `PredictionPoweredMeanInferenceResult`. Here the embedded `confidence_sequence` comes last in the field list, after the scalar and array fields. The inference base currently lists `confidence_interval` first and its subclasses redundantly redeclare the base fields; aligning it (moving `confidence_interval` last on `MeanInferenceResult` and dropping the redeclarations so the subclasses inherit that order) is split out into its own refactoring ticket `ref-confidence-interval-field-order.md`, since it also touches both inference subclasses and is independent of monitoring. This ticket does not depend on that one. The `__init__.py` exports all three names.

Like the inference results, which hold a `confidence_interval: ConfidenceInterval` rather than loose bound arrays, the monitoring base holds a `confidence_sequence: ConfidenceSequence` (from `glide/confidence_sequences/`) and re-derives `running_means`, `confidence_bounds` and `alarms` from it as read-only properties (`alarms` via `confidence_sequence.test_null_hypothesis(alarm_threshold, alternative)`, the alternative chosen from `higher_is_better`: `"larger"` for a risk, `"smaller"` for a performance). The shared monitoring machinery (the running means, bounds, alarms, alarm bookkeeping and display) is method-agnostic, so it all lives on the base `MeanMonitoringResult`.

What differs across methods is the per-batch sample-size bookkeeping, and this is exactly what gives each subclass its own content, mirroring the inference subclasses (`ClassicalMeanInferenceResult` carries `n`; `PredictionPoweredMeanInferenceResult` carries `n_true` and `n_proxy`). The monitoring counterpart is per-batch rather than scalar, since a monitoring run spans many batches: `ClassicalMeanMonitoringResult` adds `batch_n: NDArray` (the number of labeled samples in each batch), and `PredictionPoweredMeanMonitoringResult` adds `batch_n_true: NDArray` and `batch_n_proxy: NDArray` (the per-batch labeled and total counts). These slot in smoothly because each monitor already partitions every batch into labeled/unlabeled blocks while computing its per-batch estimate, so the counts are a free by-product collected into arrays of length `n_batches` alongside `batch_estimates`. They land after the base fields (the embedded `confidence_sequence` stays last on the base, the per-batch count arrays follow it on each subclass), exactly as the inference subclasses append their scalars after the base's fields. The `monitor_name` field records which monitor produced a given result, mirroring `estimator_name` on the inference results.

This package depends only on the `ConfidenceSequence` protocol from `glide/confidence_sequences/`, not on the monitors themselves, so it lands right after that package and is tested in isolation before either monitor exists.

### MeanMonitoringResult

**Location:** `glide/mean_monitoring_results/base.py`

**Purpose:** Shared base dataclass for monitoring results, holding the per-batch estimates, the alarm threshold the bounds are tested against, and the embedded confidence sequence from which the running means, bounds and alarms are derived. Mirrors `MeanInferenceResult`, which embeds a `ConfidenceInterval`.

```python
from typing import List, Optional

from glide.confidence_sequences import ConfidenceSequence


@dataclass(repr=False)
class MeanMonitoringResult:
    """Result of a drift-monitoring procedure over a batched dataset.

    Holds the embedded ``confidence_sequence`` and exposes its per-look arrays as
    properties. Every array is indexed by batch in chronological order and is
    expressed in the original metric units. For batch ``t``:

    - ``running_means[t]`` is the mean of the per-batch estimates up to ``t``;
    - ``confidence_bounds[t]`` is the anytime-valid bound on that running mean,
      on the side where drift is harmful (a lower bound when the metric is a
      risk, an upper bound when it is a performance);
    - ``alarms[t]`` is ``True`` once ``confidence_bounds[t]`` has crossed
      ``alarm_threshold``, signalling a drift.

    ``alarm_threshold`` is the user-supplied business threshold the confidence
    bound is tested against: the worst metric value the user is willing to
    tolerate, in metric units.

    Because the threshold is a known constant, the only false-alarm budget is
    the embedded sequence's, reported as ``confidence_level`` (one minus its
    miscoverage): under no drift, the probability of ever raising a false alarm
    is at most ``1 - confidence_level``.
    """

    metric_name: str
    monitor_name: str
    higher_is_better: bool
    alarm_threshold: float
    batch_identifiers: NDArray
    batch_estimates: NDArray
    confidence_sequence: ConfidenceSequence

    @property
    def running_means(self) -> NDArray:
        result = self.confidence_sequence.running_means
        return result

    @property
    def confidence_bounds(self) -> NDArray:
        result = self.confidence_sequence.confidence_bounds
        return result

    @property
    def alarms(self) -> NDArray:
        alternative = "smaller" if self.higher_is_better else "larger"
        result = self.confidence_sequence.test_null_hypothesis(self.alarm_threshold, alternative)
        return result

    @property
    def n_batches(self) -> int:
        result = len(self.batch_estimates)
        return result

    @property
    def drift_detected(self) -> bool:
        result = bool(self.alarms.any())
        return result

    @property
    def first_alarm_index(self) -> Optional[int]:
        if not self.drift_detected:
            return None
        result = int(self.alarms.argmax())
        return result

    def __str__(self) -> str:
        confidence_level = self.confidence_sequence.confidence_level
        lines: List[str] = [
            f"Metric: {self.metric_name}",
            f"Monitor: {self.monitor_name}",
            f"Number of Batches: {self.n_batches}",
            f"Drift Detected: {self.drift_detected}",
            f"First Alarm Index: {self.first_alarm_index}",
            f"Alarm Threshold: {self.alarm_threshold:.3f}",
            f"Running Mean: {self.running_means[-1]:.3f}",
            f"Confidence Bound: {self.confidence_bounds[-1]:.3f}",
            f"Confidence Level: {confidence_level:.2f}",
        ]
        return "\n".join(lines)

    def summary(self) -> str:
        """Return a formatted summary of the monitoring result."""
        return self.__str__()

    def __repr__(self) -> str:
        return self.__str__()
```

### ClassicalMeanMonitoringResult

**Location:** `glide/mean_monitoring_results/classical.py`

**Purpose:** Result returned by `ClassicalMeanMonitor`. Mirrors `ClassicalMeanInferenceResult`, which carries `n`: it adds the per-batch labeled sample count `batch_n`.

```python

from glide.mean_monitoring_results.base import MeanMonitoringResult


@dataclass(repr=False)
class ClassicalMeanMonitoringResult(MeanMonitoringResult):
    """Monitoring result for the classical (proxy-free) monitor.

    Adds the per-batch labeled sample count to the shared base. ``batch_n[t]``
    is the number of labeled samples in batch ``t`` and has the same length as
    ``batch_estimates``.
    """

    batch_n: NDArray
```

### PredictionPoweredMeanMonitoringResult

**Location:** `glide/mean_monitoring_results/prediction_powered.py`

**Purpose:** Result returned by `PPIMeanMonitor`. Mirrors `PredictionPoweredMeanInferenceResult`, which carries `n_true` and `n_proxy`: it adds the per-batch labeled and total counts `batch_n_true` and `batch_n_proxy`.

```python

from glide.mean_monitoring_results.base import MeanMonitoringResult


@dataclass(repr=False)
class PredictionPoweredMeanMonitoringResult(MeanMonitoringResult):
    """Monitoring result for the prediction-powered monitor.

    Adds the per-batch labeled and total sample counts to the shared base.
    ``batch_n_true[t]`` is the number of labeled samples in batch ``t`` and
    ``batch_n_proxy[t]`` the total number of samples (labeled plus unlabeled);
    both have the same length as ``batch_estimates``.
    """

    batch_n_true: NDArray
    batch_n_proxy: NDArray
```

### Package exports

**Location:** `glide/mean_monitoring_results/__init__.py`

```python
from glide.mean_monitoring_results.base import MeanMonitoringResult
from glide.mean_monitoring_results.classical import ClassicalMeanMonitoringResult
from glide.mean_monitoring_results.prediction_powered import PredictionPoweredMeanMonitoringResult

__all__ = [
    "MeanMonitoringResult",
    "ClassicalMeanMonitoringResult",
    "PredictionPoweredMeanMonitoringResult",
]
```

Unit tests mirror the layout: `tests/unit/mean_monitoring_results/test_base.py`, `tests/unit/mean_monitoring_results/test_classical.py`, `tests/unit/mean_monitoring_results/test_prediction_powered.py`. Tests construct results with a real `EmpiricalBernsteinConfidenceSequence` (cheap to build from small hardcoded `running_means`/`confidence_bounds` arrays), so the property delegation and the `alarms` derivation exercise the genuine sequence object. The subclass test files additionally assert that the per-batch count arrays (`batch_n` for classical; `batch_n_true`/`batch_n_proxy` for prediction-powered) round-trip from construction, against small hardcoded arrays. Because `--doctest-modules` is active, the docstrings are collected; the classes carry no doctests (passive containers), so the property and `__str__` behaviour is covered by the unit tests below.

## Corner cases

- `alarms` is derived on each access from `confidence_sequence.test_null_hypothesis(alarm_threshold, alternative)`, with `alternative` `"larger"` when `higher_is_better` is `False` (risk) and `"smaller"` when `True` (performance); it is never stored as a field.
- `drift_detected` is `False` and `first_alarm_index` is `None` when `alarms` is all-`False`; `first_alarm_index` is the index of the first `True` otherwise (a `0`-th batch alarm returns `0`, not `None`).
- `n_batches` equals `len(batch_estimates)` and counts every batch (there is no reference batch to exclude).
- The confidence level is read inline from `confidence_sequence.confidence_level` inside `__str__` and is not exposed as a property or stored as a field, mirroring how `MeanInferenceResult` reads `confidence_interval.confidence_level` inline rather than re-exposing it.
- `__str__`, `summary` and `__repr__` return the same string, and `summary()` reads the last entry of the `running_means` and `confidence_bounds` properties plus the confidence level, so a result with at least one batch always formats without error.

## Acceptance criteria

- [ ] `glide/mean_monitoring_results/` exists with `base.py`, `classical.py`, `prediction_powered.py` and `__init__.py`, mirroring `glide/mean_inference_results/`.
- [ ] `MeanMonitoringResult` holds the seven fields above (no `source_miscoverage`; the list ends in `confidence_sequence: ConfidenceSequence`) and exposes `running_means`, `confidence_bounds` and `alarms` as properties delegating to it; `ClassicalMeanMonitoringResult` adds `batch_n: NDArray` and `PredictionPoweredMeanMonitoringResult` adds `batch_n_true: NDArray` and `batch_n_proxy: NDArray`, each after the base fields.
- [ ] The per-batch count arrays round-trip from construction (unit tests asserting `batch_n`, `batch_n_true` and `batch_n_proxy` against small hardcoded arrays).
- [ ] `running_means` and `confidence_bounds` return the embedded sequence's arrays, and `alarms` returns its `test_null_hypothesis(alarm_threshold, alternative)` vector with `alternative` chosen from `higher_is_better` (unit test asserting `"larger"` for a risk and `"smaller"` for a performance against a hardcoded sequence).
- [ ] `first_alarm_index` returns `None` on an all-`False` `alarms` array and the index of the first `True` otherwise (unit test driven by a hardcoded confidence sequence, including the `index 0` case).
- [ ] `drift_detected` and `n_batches` return hardcoded expected values on a small constructed result.
- [ ] `__str__`, `__repr__` and `summary()` return identical strings, asserted against a hardcoded expected string on a small constructed result.
- [ ] All three classes are importable from `glide.mean_monitoring_results`.

imerad:
### Description

- Adds `glide.mean_monitoring_results` module with `MeanMonitoringResult` (base), `ClassicalMeanMonitoringResult` and `PredictionPoweredMeanMonitoringResult`: result objects returned by anytime-valid drift-monitoring procedures over batched data, exposing running means, confidence bounds and alarm state derived from an embedded `ConfidenceSequence`.
- Closes #315
- Noteworthy: `running_means` was renamed to `running_mean_estimates` on the `ConfidenceSequence` object. `confidence_level` is now supplied as a field at construction on `MeanMonitoringResult`, since it was removed from `ConfidenceSequence`.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/tests/unit/mean_monitoring_results/test_base.py
+++ b/tests/unit/mean_monitoring_results/test_base.py
@@ -1,0 +1,104 @@
+import numpy as np
+
+from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
+from glide.mean_monitoring_results import MeanMonitoringResult
+
+# --- MeanMonitoringResult (common attributes and properties) ---
+
+_SEQUENCE = EmpiricalBernsteinConfidenceSequence(
+    running_mean_estimates=np.array([0.4, 0.6]),
+    confidence_bounds=np.array([0.1, 0.55]),
+)
+_RESULT_HIGHER_IS_BETTER = MeanMonitoringResult(
+    metric_name="accuracy",
+    monitor_name="Test",
+    higher_is_better=True,
+    alarm_threshold=0.5,
+    confidence_level=0.95,
+    batch_identifiers=np.array([0, 1]),
+    batch_mean_estimates=np.array([0.4, 0.8]),
+    confidence_sequence=_SEQUENCE,
+)
+_RESULT_LOWER_IS_BETTER = MeanMonitoringResult(
+    metric_name="loss",
+    monitor_name="Test",
+    higher_is_better=False,
+    alarm_threshold=0.5,
+    confidence_level=0.95,
+    batch_identifiers=np.array([0, 1]),
+    batch_mean_estimates=np.array([0.4, 0.8]),
+    confidence_sequence=_SEQUENCE,
+)
+_RESULT_NO_ALARM = MeanMonitoringResult(
+    metric_name="accuracy",
+    monitor_name="Test",
+    higher_is_better=True,
+    alarm_threshold=-1.0,
+    confidence_level=0.95,
+    batch_identifiers=np.array([0, 1]),
+    batch_mean_estimates=np.array([0.4, 0.8]),
+    confidence_sequence=_SEQUENCE,
+)
+
+
+def test_base_running_means():
+    np.testing.assert_array_equal(_RESULT_HIGHER_IS_BETTER.running_means, np.array([0.4, 0.6]))
+
+
+def test_base_confidence_bounds():
+    np.testing.assert_array_equal(_RESULT_HIGHER_IS_BETTER.confidence_bounds, np.array([0.1, 0.55]))
+
+
+def test_base_alarms_higher_is_better():
+    np.testing.assert_array_equal(_RESULT_HIGHER_IS_BETTER.alarms, np.array([True, False]))
+
+
+def test_base_alarms_lower_is_better():
+    np.testing.assert_array_equal(_RESULT_LOWER_IS_BETTER.alarms, np.array([False, True]))
+
+
+def test_base_n_batches():
+    assert _RESULT_HIGHER_IS_BETTER.n_batches == 2
+
+
+def test_base_drift_detected_true():
+    assert _RESULT_HIGHER_IS_BETTER.drift_detected is True
+
+
+def test_base_drift_detected_false():
+    assert _RESULT_NO_ALARM.drift_detected is False
+
+
+def test_base_first_alarm_index_at_zero():
+    assert _RESULT_HIGHER_IS_BETTER.first_alarm_index == 0
+
+
+def test_base_first_alarm_index_not_at_zero():
+    assert _RESULT_LOWER_IS_BETTER.first_alarm_index == 1
+
+
+def test_base_first_alarm_index_none():
+    assert _RESULT_NO_ALARM.first_alarm_index is None
+
+
+def test_base_repr_equals_str_equals_summary():
+    assert repr(_RESULT_HIGHER_IS_BETTER) == str(_RESULT_HIGHER_IS_BETTER)
+    assert str(_RESULT_HIGHER_IS_BETTER) == _RESULT_HIGHER_IS_BETTER.summary()
+
+
+# --- MeanMonitoringResult.__str__ ---
+
+
+def test_base_str():
+    expected = (
+        "Metric: accuracy\n"
+        "Monitor: Test\n"
+        "Number of Batches: 2\n"
+        "Drift Detected: True\n"
+        "First Alarm Index: 0\n"
+        "Alarm Threshold: 0.500\n"
+        "Running Mean: 0.600\n"
+        "Confidence Bound: 0.550\n"
+        "Confidence Level: 0.95"
+    )
+    assert str(_RESULT_HIGHER_IS_BETTER) == expected
--- a/tests/unit/mean_monitoring_results/test_base.py
+++ b/tests/unit/mean_monitoring_results/test_base.py
@@ -44,7 +44,7 @@ def result_without_alarm(sequence):
     )
 
 
-def test_base_properties(result_with_alarm, result_without_alarm):
+def test_base_properties_with_alarm(result_with_alarm):
     np.testing.assert_array_equal(result_with_alarm.running_means, np.array([0.4, 0.6]))
     np.testing.assert_array_equal(result_with_alarm.confidence_bounds, np.array([0.1, 0.55]))
     assert result_with_alarm.n_batches == 2
@@ -52,6 +52,8 @@ def test_base_properties(result_with_alarm, result_without_alarm):
     assert result_with_alarm.drift_detected is True
     assert result_with_alarm.first_alarm_index == 0
 
+
+def test_base_properties_without_alarm(result_without_alarm):
     np.testing.assert_array_equal(result_without_alarm.alarms, np.array([False, False]))
     assert result_without_alarm.drift_detected is False
     assert result_without_alarm.first_alarm_index is None
--- a/tests/unit/mean_monitoring_results/test_base.py
+++ b/tests/unit/mean_monitoring_results/test_base.py
@@ -1,6 +1,9 @@
+from unittest.mock import patch
+
 import numpy as np
 import pytest
 
+import glide.mean_monitoring_results.base as base_module
 from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
 from glide.mean_monitoring_results import MeanMonitoringResult
 
@@ -23,7 +26,6 @@ def result_with_alarm(sequence):
         higher_is_better=True,
         alarm_threshold=0.5,
         confidence_level=0.95,
-        batch_identifiers=np.array([0, 1]),
         batch_mean_estimates=np.array([0.4, 0.8]),
         confidence_sequence=sequence,
     )
@@ -37,7 +39,6 @@ def result_without_alarm(sequence):
         higher_is_better=False,
         alarm_threshold=2.0,
         confidence_level=0.95,
-        batch_identifiers=np.array([0, 1]),
         batch_mean_estimates=np.array([0.4, 0.8]),
         confidence_sequence=sequence,
     )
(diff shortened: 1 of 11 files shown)


imerad:
## Code review (automated)

Overall: this is a clean, purely-additive PR — mirrors the existing `mean_inference_results` pattern well, docs/CHANGELOG/mkdocs entries are correctly placed, and the `higher_is_better` → alarm-direction logic is correct (verified numerically against the test fixtures). No blocking issues found; the items below are worth a look before a real "monitor" class starts constructing these results in production.

### 1. Silent data-mismatch risk (PLAUSIBLE)
**`glide/mean_monitoring_results/classical.py:17`** (and similarly `prediction_powered.py:18-19`)

`batch_n` (resp. `batch_n_true`/`batch_n_proxy`) isn't validated against `batch_mean_estimates`'s length, despite the docstring promising they're equal length. A future off-by-one bug in a monitor implementation would construct the object successfully and `__str__()`/`summary()` would silently print a count from the wrong batch — no exception, just a misleading number exactly when someone is reading it to trust a drift alarm.

### 2. Dual source of truth for alarm direction (PLAUSIBLE)
**`glide/mean_monitoring_results/base.py:55`**

```python
alternative = "smaller" if self.higher_is_better else "larger"
```

`higher_is_better` is stored independently from the orientation already baked into `confidence_sequence.confidence_bounds` (per its own docstring, the bound is already computed "on the side where drift is harmful"). Nothing checks the two agree. If a future caller passes a `higher_is_better` inconsistent with how the confidence sequence was actually oriented, `alarms` silently inverts with no error.

### 3. Unguarded empty-array crash (PLAUSIBLE)
**`glide/mean_monitoring_results/base.py:84-85`**

`_common_lines()` indexes `running_means[-1]` / `confidence_bounds[-1]` unconditionally. A `MeanMonitoringResult` constructed with zero batches (e.g. a monitor run that produced no completed batches) would raise `IndexError` on `str()`/`repr()`/`.summary()` instead of a meaningful message.

### 4. Dead field (CONFIRMED)
**`glide/mean_monitoring_results/base.py:39`**

`batch_identifiers` is a required constructor field with no reader anywhere — not exposed via any property, not used in `_common_lines()`/`alarms`/etc., and not consumed by `glide/io/export.py`. Every caller (including all three new test fixtures) must supply it just to satisfy the dataclass constructor.

### 5. Duplicated summary boilerplate (PLAUSIBLE, minor/borderline)
**`glide/mean_monitoring_results/base.py:90-98`**

`__str__`/`summary`/`__repr__` (and the `_common_lines()`-extension idiom in `classical.py`/`prediction_powered.py`) are a byte-for-byte duplicate of the pattern in `glide/mean_inference_results/base.py`, with no shared mixin. Minor — only ~9 lines duplicated, and a shared abstraction may not be worth it yet, but flagging per the "No Redundancy" convention.

---
*Generated by an automated code review pass.*

tests/unit/mean_monitoring_results/test_base.py
```diff
@@ -0,0 +1,110 @@
...
+    np.testing.assert_array_equal(result_with_alarm.running_means, np.array([0.4, 0.6]))
+    np.testing.assert_array_equal(result_with_alarm.confidence_bounds, np.array([0.1, 0.55]))
+    assert result_with_alarm.n_batches == 2
+    np.testing.assert_array_equal(result_with_alarm.alarms, np.array([True, False]))
+    assert result_with_alarm.drift_detected is True
+    assert result_with_alarm.first_alarm_index == 0
+
+    np.testing.assert_array_equal(result_without_alarm.alarms, np.array([False, False]))
+    assert result_without_alarm.drift_detected is False
+    assert result_without_alarm.first_alarm_index is None
```
gmartinon-ed, line 59: A séparer dans un second test ?

gmartinon-ed (approved):

tests/unit/mean_monitoring_results/test_base.py
imerad: Fait

PR #334, opened 2026-07-06
Add ClassicalMeanMonitor for drift detection over batched data

### LINKED ISSUE #316: ClassicalMeanMonitor: anytime-valid drift detection over batched production data
# TODO
- Investigate if `_unique_ordered_batches` is really needed, given that `np.unique` has an argument `sorted=True` by default.
- Investigate if we can reduce the complexity of this function (e.g., first identify unique batches with a single pass, and then check unicity on the result)
- Rename `risk_batch_estimate` in `risk_batch_mean_estimate`
- Investigate if we can vectorize the batch_mean_estimate with `np.bincount(..., weight=...)` for this estimator, and also for the forthcoming PPIMeanMonitor

## Background

After a GenAI system is deployed, its quality metric is re-estimated on successive batches of production data, for instance by collecting a fresh batch every week. When estimation relies on *human labels only* (no proxy predictions), each batch yields an ordinary sample-mean estimate of the metric over its labeled observations. The natural question is: "does this sequence of batches prove that the metric has drifted past an acceptable business threshold?". Comparing each new batch estimate to the threshold with an ordinary confidence interval is statistically invalid: each comparison carries its own false-alarm probability, so checking after every batch makes a false alarm almost certain over time (the classical "multiple testing" problem, here called *peeking*).

The fix is a *confidence sequence*: a sequence of confidence bounds that holds simultaneously at all times, so the user may look after every batch and the total false-alarm probability stays below a chosen budget. The monitor builds an anytime-valid bound on the *running mean* (the average of the per-batch sample means up to now) using an empirical-Bernstein bound (a concentration bound whose width adapts to the empirically observed variability of the estimates, instead of assuming a known variance). In math notation, for a risk (lower is better), the alarm rule at batch t is: raise drift if `L_t > threshold`, where `L_t` is the anytime-valid lower bound on the running risk after t batches and `threshold` is a user-supplied value (the worst running risk the user is willing to tolerate, in metric units). Under no drift (the running risk stays at or below the threshold), the probability of *ever* raising a false alarm is at most the single miscoverage budget `1 - confidence_level`.

This is the classical baseline of `PPIMeanMonitor` (ticket `feat-ppi-mean-monitor.md`): same confidence-sequence machinery, but the per-batch estimate is the plain sample mean rather than a prediction-powered estimate. It is the monitoring counterpart of how `ClassicalMeanEstimator` is the proxy-free baseline of `PPIMeanEstimator`.

## Design choices

The monitor lives in `glide/monitors/classical.py`, class `ClassicalMeanMonitor`, mirroring `glide/estimators/classical.py` (a `_preprocess`/`_postprocess` pair around a public `detect` method; the symmetric `_postprocess` is new relative to the estimators, warranted by the round-trip risk-orientation scaling the monitor adds). The anytime-valid confidence-sequence machinery is method-agnostic and lives in its own package `glide/confidence_sequences/`. What stays in the shared `glide/monitors/core.py` is only the monitor-specific glue: the batch-chronology derivation (`_unique_ordered_batches`) and the risk-orientation scaling (`_scale_to_unit_risk`, `_scale_from_unit_risk`). Arbitrary metric scales (a Likert scale, 0–20, 0–100, ...) are handled exactly as in `PPIMeanMonitor`, through `metric_lower_bound`/`metric_upper_bound`: `_scale_to_unit_risk` maps any such range onto `[0, 1]` before the sequence sees the data.

The classical monitor returns `ClassicalMeanMonitoringResult`, a thin subclass of the shared `MeanMonitoringResult` base.

The monitor is **stateless** and consumes raw data: `detect(y, batches, higher_is_better, threshold)` extends `ClassicalMeanEstimator.estimate(y)` with `batches` the way the stratified estimators extend with `groups`, except `batches` is order-significant. Like `ClassicalMeanEstimator`, unlabeled entries are passed as `NaN` and dropped per batch. Chronology rides on row order: the data is passed oldest batch first, every batch is monitored, and each batch must occupy a contiguous block of rows, which `_preprocess` enforces (interleaved rows raise). Identifier values are never compared, so any label type works (integers, dates, free-form strings), sidestepping the `"batch10" < "batch2"` lexicographic trap. The user accumulates production data however they like and calls `detect` whenever they like on the whole accumulated dataset; anytime validity is what makes repeated calls on growing histories jointly valid. The `higher_is_better` boolean handles performance metrics by internally flipping the metric into "risk orientation" (lower is better) so a single code path serves both cases.

The alarm target is the user-supplied `threshold`, mapped into risk space once. The empirical-Bernstein variance process is seeded with that same threshold (the only predictable center available before the first batch); the seed influences only the first look's width and washes out as batches accumulate. The lone false-alarm budget is the sequence's, `1 - confidence_level`.

### Shared internal helpers

**Location:** `glide/monitors/core.py`

**Purpose:** The monitor-specific glue shared by every monitor: the batch-chronology derivation (`_unique_ordered_batches`) and the metric↔risk orientation scaling (`_scale_to_unit_risk`, `_scale_from_unit_risk`).

```python
from typing import Union

import numpy as np

def _unique_ordered_batches(batches: NDArray) -> NDArray:
    # Deduplicate the batches to their distinct labels, ordered by first appearance
    # rather than by value. np.unique returns
    # the labels sorted by value (which would put "batch10" before "batch2"), together
    # with the row index where each first appears; reordering the labels by that index
    # recovers oldest-first order and makes any label type (ints, dates, strings) work.
    unique_identifiers, first_appearance_indices = np.unique(batches, return_index=True)
    chronological_order = np.argsort(first_appearance_indices)
    batch_identifiers = unique_identifiers[chronological_order]
    n_distinct_batches = len(batch_identifiers)
    _validate_bounds(
        n_distinct_batches,
        "batches",
        lower=1,
        error_message=(
            f"'batches' must contain at least 1 batch to monitor; got {n_distinct_batches}."
        ),
    )
    # Each batch must occupy a single contiguous block; interleaved rows signal unsorted
    # input, the most common chronology mistake, so we reject them. The number of maximal
    # runs of equal identifiers must equal the number of batches.
    block_starts = np.hstack([[True], batches[1:] != batches[:-1]])
    if np.count_nonzero(block_starts) != n_distinct_batches:
        raise ValueError(
            "'batches' must be grouped into contiguous blocks ordered oldest first; "
            "found interleaved batches. Sort the data by batch before calling detect."
        )
    return batch_identifiers


def _scale_to_unit_risk(
    values: Union[float, NDArray],
    metric_lower_bound: float,
    metric_upper_bound: float,
    higher_is_better: bool,
) -> Union[float, NDArray]:
    # Map the metric onto [0, 1] with "lower is better" orientation, so one code
    # path serves both risks and performances.
    scaled = (values - metric_lower_bound) / (metric_upper_bound - metric_lower_bound)
    if higher_is_better:
        scaled = 1.0 - scaled
    return scaled


def _scale_from_unit_risk(
    values: Union[float, NDArray],
    metric_lower_bound: float,
    metric_upper_bound: float,
    higher_is_better: bool,
) -> Union[float, NDArray]:
    if higher_is_better:
        values = 1.0 - values
    original = metric_lower_bound + values * (metric_upper_bound - metric_lower_bound)
    return original
```

### ClassicalMeanMonitor

**Location:** `glide/monitors/classical.py`

**Purpose:** Stateless anytime-valid drift detector over a batched dataset of labels, mirroring `ClassicalMeanEstimator`.

```python

import numpy as np

from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
from glide.core.validation import (
    _validate_bounds,
    _validate_equal_lengths,
    _validate_min_samples,
)
from glide.mean_monitoring_results import ClassicalMeanMonitoringResult
from glide.monitors.core import (
    _unique_ordered_batches,
    _scale_from_unit_risk,
    _scale_to_unit_risk,
)


class ClassicalMeanMonitor:
    """Anytime-valid drift monitor over a batched dataset of labels.

    It uses the plain sample
    mean per batch. Unlabeled entries are passed as ``np.nan`` and dropped per batch. The user accumulates non-overlapping
    production batches and calls :meth:`detect` on the whole accumulated dataset at
    any time. Data is passed oldest batch first, and every batch is monitored. The
    monitor computes a sample-mean estimate per batch, builds an anytime-valid
    confidence sequence on the running mean of those estimates, and raises a drift
    alarm when it crosses a user-supplied ``threshold`` (the worst metric value the
    user is willing to tolerate). Because the bounds are valid at all times
    simultaneously, the user may call :meth:`detect` after every new batch without
    inflating the false-alarm probability.

    Notes
    -----
    The confidence sequence is built over the per-batch sample means as its
    atoms, so the monitored quantity is the equal-weighted running mean of the
    per-batch estimates. This matches the way human annotations arrive in
    practice (in batches, not as a per-sample stream) and equals a per-sample
    running mean only when every batch shares the same labeled count. The bound
    width splits into two parts: the underlying statistical precision, set by the
    total number of labels (the usual root-n term, unaffected by how the labels
    are split across batches), plus the anytime-valid overhead, the price for
    being allowed to peek after every batch, which decays as one over the number
    of batches. Both parts shrink as data accumulates; the overhead is simply
    what keeps the earliest looks wide, so no alarm can fire in the first few
    batches. This is the expected, modest cost of valid repeated looking, not a
    loss of sample efficiency.

    References
    ----------
    no. 2 (2021): 1055-1080.

    Podkopaev, Aleksandr, and Aaditya Ramdas. "Tracking the risk of a deployed model
    and detecting harmful distribution shifts." International Conference on Learning
    Representations (ICLR), 2022.

    Waudby-Smith, Ian, and Aaditya Ramdas. "Estimating means of bounded random
    variables by betting." Journal of the Royal Statistical Society Series B:
    Statistical Methodology 86, no. 1 (2024): 1-27.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.monitors import ClassicalMeanMonitor
    >>> pre_drift_batch = np.array([0.0, 0.2, np.nan, np.nan])
    >>> post_drift_batch = np.array([0.8, 1.0, np.nan, np.nan])
    >>> y = np.hstack([pre_drift_batch, np.tile(post_drift_batch, 50)])
    >>> monitor = ClassicalMeanMonitor()
    >>> result = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
    >>> result.drift_detected
    True
    >>> result.first_alarm_index
    11
    """

    def _preprocess(
        self,
        y: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
    ) -> Tuple[NDArray, float, NDArray]:
        _validate_equal_lengths(y, batches, names=["y", "batches"])
        labeled_mask = ~np.isnan(y)
        labeled_values = y[labeled_mask]
        _validate_bounds(
            labeled_values,
            "y",
            lower=metric_lower_bound,
            upper=metric_upper_bound,
            error_message=f"'y' values must lie within [{metric_lower_bound!r}, {metric_upper_bound!r}].",
        )
        batch_identifiers = _unique_ordered_batches(batches)
        for identifier in batch_identifiers:
            batch_mask = batches == identifier
            batch_labeled_mask = batch_mask & labeled_mask
            _validate_min_samples(y[batch_labeled_mask], "y")
        risk_y = _scale_to_unit_risk(y, metric_lower_bound, metric_upper_bound, higher_is_better)
        risk_threshold = _scale_to_unit_risk(threshold, metric_lower_bound, metric_upper_bound, higher_is_better)
        return risk_y, risk_threshold, batch_identifiers

    def _postprocess(
        self,
        risk_running_means: NDArray,
        risk_lower_bounds: NDArray,
        risk_batch_estimates: NDArray,
        higher_is_better: bool,
        running_means = _scale_from_unit_risk(risk_running_means, metric_lower_bound, metric_upper_bound, higher_is_better)
        confidence_bounds = _scale_from_unit_risk(risk_lower_bounds, metric_lower_bound, metric_upper_bound, higher_is_better)
        batch_estimates = _scale_from_unit_risk(risk_batch_estimates, metric_lower_bound, metric_upper_bound, higher_is_better)
        return running_means, confidence_bounds, batch_estimates

    def detect(
        self,
        y: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
        metric_name: str = "Metric",
    ) -> ClassicalMeanMonitoringResult:

        Splits the data by batch, computes a sample-mean estimate per batch, and
        builds an anytime-valid empirical-Bernstein confidence sequence on the
        running mean of those estimates. An alarm is raised at every batch where the
        sequence crosses the user-supplied ``threshold``.

        Rows must be ordered oldest batch first and grouped into contiguous blocks;
        identifier values are not compared, so any label type works. Batches must be
        non-overlapping (no shared samples), and successive calls must be made on
        growing histories of the same data; passing the full accumulated dataset at
        every call makes the anytime-valid guarantee hold jointly over all calls.
        Alternatively the data may be restricted to the most recent batches, in which
        case the guarantee holds within each restriction but not across the moving
        history as a whole.

        Parameters
        ----------
        y : NDArray
            Array of observations, shape ``(n_samples,)``. Unlabeled entries are
            ``np.nan`` and are dropped per batch; every batch must keep at least 2
            labeled entries.
        batches : NDArray
            Array of batch identifiers, shape ``(n_samples,)``. Rows must be ordered
            oldest batch first and grouped into contiguous blocks. Identifier values
            are not compared, so any hashable label type works (integers, dates,
            free-form strings).
        higher_is_better : bool
            ``False`` when the metric is a risk (drift means the metric increased),
            ``True`` when it is a performance (drift means the metric decreased).
        threshold : float
            The metric value the running mean is monitored against, in metric units:
            the worst level the user is willing to tolerate. An alarm fires once the
            anytime-valid bound proves the running metric has crossed it (the running
            risk exceeds it for a risk, the running performance falls below it for a
            performance). Must lie within ``[metric_lower_bound, metric_upper_bound]``.
        metric_name : str, optional
            How confident each alarm should be. At the default ``0.8`` the monitor
            raises "80% confident" alarms: under no drift it has at most a 20% chance
            of *ever* raising a false alarm, no matter how often you look. Raising it
            (say to ``0.95``) demands more evidence before alarming, so alarms become
            more trustworthy but arrive later; lowering it detects sooner at the cost
            of more false alarms.
        metric_lower_bound : float, optional
            Known lower bound of the metric. Defaults to ``0.0``.
        metric_upper_bound : float, optional
            Known upper bound of the metric. Defaults to ``1.0``.

        Returns
        -------
        ClassicalMeanMonitoringResult
            Per-batch estimates, running means, anytime-valid confidence bounds,
            alarm flags, and the alarm threshold, all in the original metric
            orientation.

        Raises
        ------
        ValueError
            - If ``y`` and ``batches`` have different lengths.
            - If any labeled value falls outside ``[metric_lower_bound, metric_upper_bound]``.
            - If ``batches`` is empty (no batch to monitor).
            - If batches are interleaved rather than grouped into contiguous blocks.
            - If any batch has fewer than 2 labeled (non-NaN) samples.
            - If ``metric_lower_bound >= metric_upper_bound``.
            - If ``threshold`` falls outside ``[metric_lower_bound, metric_upper_bound]``.
            - If ``confidence_level`` is not in (0, 1).
        """
        _validate_bounds(
            confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
        )
        _validate_bounds(
            metric_lower_bound,
            "metric_lower_bound",
            upper=metric_upper_bound,
            right_inclusive=False,
            error_message=(
                f"'metric_lower_bound' must be strictly smaller than 'metric_upper_bound'; "
                f"got {metric_lower_bound!r} and {metric_upper_bound!r}."
            ),
        )
        _validate_bounds(threshold, "threshold", lower=metric_lower_bound, upper=metric_upper_bound)
        risk_y, risk_threshold, batch_identifiers = self._preprocess(
            y, batches, higher_is_better, threshold, metric_lower_bound, metric_upper_bound
        )

        risk_batch_estimates_list: List[float] = []
        batch_n_list: List[int] = []
            batch_mask = batches == batch_identifiers[position]
            batch_values = risk_y[batch_mask]
            risk_batch_estimate = np.nanmean(batch_values)
            risk_batch_estimates_list.append(risk_batch_estimate)
            batch_n_list.append(np.count_nonzero(~np.isnan(batch_values)))
        risk_batch_estimates = np.array(risk_batch_estimates_list)
        batch_n = np.array(batch_n_list)

        risk_running_means, risk_lower_bounds = _compute_empirical_bernstein_bounds(
            risk_batch_estimates, risk_threshold, miscoverage
        )

        running_means, confidence_bounds, batch_estimates = self._postprocess(
            risk_running_means, risk_lower_bounds, risk_batch_estimates, higher_is_better, metric_lower_bound, metric_upper_bound
        )
        confidence_sequence = EmpiricalBernsteinConfidenceSequence(
            running_means=running_means,
        )
            metric_name=metric_name,
            alarm_threshold=threshold,
            batch_identifiers=batch_identifiers,
            batch_estimates=batch_estimates,
            batch_n=batch_n,
        )
        return result
```

`glide/monitors/__init__.py` exports `ClassicalMeanMonitor`; add it to `glide/__init__.py` exports as well. New unit tests for this ticket: `tests/unit/monitors/test_classical.py` and `tests/unit/monitors/test_core.py` (the latter covering only `_unique_ordered_batches` and the orientation scaling). The boundary helpers and the sequence builder are tested by `feat-confidence-sequence.md`, and the result dataclasses by `feat-mean-monitoring-results.md`; neither is re-tested here.

## Functional tests

**Per-batch estimates match ClassicalMeanEstimator**

Each entry of `batch_estimates` must equal the `ClassicalMeanEstimator` point estimate computed on that batch's data alone (with default bounds and `higher_is_better=False`, no orientation flip applies). Every batch is monitored, so `batch_estimates[0]` is the first batch (identifier 0).

```python
def test_detect_batch_estimates_match_classical_estimator(y, batches):
    result = ClassicalMeanMonitor().detect(y, batches, higher_is_better=False, threshold=0.5)
    batch_mask = batches == 0
    estimator_result = ClassicalMeanEstimator().estimate(y[batch_mask])

    assert result.batch_estimates[0] == pytest.approx(estimator_result.mean)
```

**Prefix consistency (statelessness is sound)**

Calling `detect` on the data restricted to the first k batches must produce exactly the first k entries of the arrays returned by calling `detect` on the full dataset (every batch is monitored, so there is no reference batch to exclude). This is the property that makes repeated calls on a growing history jointly valid.

```python
def test_detect_prefix_consistency():
    first_block = np.array([0.1, 0.2, 0.15, 0.25])
    later_block = np.array([0.3, 0.4, 0.35, 0.45])
    y = np.hstack([first_block, np.tile(later_block, 3)])
    batches = np.repeat(np.arange(4), 4)

    monitor = ClassicalMeanMonitor()
    full = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
    prefix_mask = batches <= 2
    prefix = monitor.detect(y[prefix_mask], batches[prefix_mask], higher_is_better=False, threshold=0.5)

    np.testing.assert_allclose(prefix.running_means, full.running_means[:3])
    np.testing.assert_allclose(prefix.confidence_bounds, full.confidence_bounds[:3])
```

**Risk/performance symmetry**

Monitoring a performance metric with `higher_is_better=True` must be the exact mirror image of monitoring the flipped data (`1 - y`) as a risk: same alarm flags, mirrored estimates, running means and bounds. The threshold flips with the data (`1 - threshold`).

```python
def test_detect_higher_is_better_symmetry(y, batches):
    risk = ClassicalMeanMonitor().detect(y, batches, higher_is_better=False, threshold=0.3)
    performance = ClassicalMeanMonitor().detect(1.0 - y, batches, higher_is_better=True, threshold=0.7)

    np.testing.assert_array_equal(performance.alarms, risk.alarms)
    np.testing.assert_allclose(performance.confidence_bounds, 1.0 - risk.confidence_bounds)
```

## Corner cases

- An empty `batches` array → `ValueError` (there is no batch to monitor).
- A batch with fewer than 2 labeled (non-NaN) samples → `ValueError`; the variance process requires at least 2 observations.
- A labeled `y` value outside `[metric_lower_bound, metric_upper_bound]` → `ValueError`; boundedness of the data is what makes the empirical-Bernstein bound valid.
- `NaN` entries in `y` are unlabeled rows: dropped per batch and never raise, so the same `y_true` stream passed to `PPIMeanMonitor` works here unchanged.
- A single batch (the first look of a growing history): must work and return a wide bound; with a metric in `[0, 1]` and default settings, the deviation budget exceeds the range for the first few batches, so no alarm can mathematically fire there. This is expected finite-sample safety, not a bug.
- Interleaved batches (e.g. rows ordered `A, B, A, B`): `ValueError`; chronology is taken from row order, so each batch must be a contiguous block. Contiguous blocks in the wrong time order cannot be detected and are the user's responsibility (documented, not enforced).
- Free-form string identifiers (e.g. `"week 1"`, `"v2.3"`): fully supported, since identifier values are never compared; order comes from row position alone.
- `threshold` outside `[metric_lower_bound, metric_upper_bound]`, `metric_lower_bound >= metric_upper_bound`, `confidence_level` outside (0, 1) → `ValueError` with the bad value in the message.
- Constant batch estimates equal to the threshold: the variance process is 0 except for the first term, bounds shrink at rate ~1/t, no alarm ever fires.

## Acceptance criteria

- [ ] `ClassicalMeanMonitor.detect` consumes raw arrays `(y, batches)` plus a required `threshold` and scalar options; the class holds no state between calls, exposes no `update`/`observe` methods, and takes no proxy predictions, tuning parameter or normalization.
- [ ] `detect` computes no reference batch: every batch is monitored, the alarm fires when the anytime-valid bound crosses the user-supplied `threshold`, and the lone false-alarm knob is `confidence_level`.
- [ ] `ClassicalMeanMonitor` calls the shared `_compute_empirical_bernstein_bounds` builder from `glide/confidence_sequences/` and reuses `_scale_to_unit_risk`/`_scale_from_unit_risk` and `_unique_ordered_batches` from `glide/monitors/core.py`.
- [ ] `ClassicalMeanMonitor.detect` returns a result whose `confidence_sequence` is an `EmpiricalBernsteinConfidenceSequence`, and `result.running_means`/`result.confidence_bounds`/`result.alarms` read through it (unit test).
- [ ] The ClassicalMeanEstimator-equivalence, prefix-consistency and risk/performance symmetry functional tests pass.
- [ ] Deterministic behavioral functional tests confirm no alarm fires on a stationary stream below the threshold and an alarm fires on a stream well above it; the many-seed false-alarm-rate and detection-power study lives in the deep-dive notebook (`doc-pprm-deep-dive.md`), not in CI.
- [ ] `ClassicalMeanMonitor` drops `NaN` entries per batch: a unit test passing the same `y_true` stream (true labels plus `NaN`) used for `PPIMeanMonitor` runs without error and estimates each batch on its labeled entries only.
- [ ] `ClassicalMeanMonitor.detect` returns a `ClassicalMeanMonitoringResult`.
- [ ] `ClassicalMeanMonitor` is importable from `glide.monitors` and `glide`.
- [ ] `_preprocess` raises `ValueError` on interleaved batches (a unit test passes rows ordered `A, B, A, B`) and accepts free-form string identifiers (a unit test labels batches `"week 1"`, `"week 2"`, ... in row order and detects normally).
- [ ] The `detect` docstring states the usage requirements the API cannot fully enforce: oldest-first row order (only interleaving is caught), non-overlapping batches, growing history across calls.

imerad:
### Description

- Adds `glide.monitors` module with `ClassicalMeanMonitor`: an anytime-valid drift detector over a batched dataset of labels, using the plain sample mean per batch and an empirical-Bernstein confidence sequence on the running mean, mirroring `ClassicalMeanEstimator`. Also adds shared batch-chronology and risk-orientation helpers in `glide/monitors/core.py`, and a `docs/api/monitors.md` API reference page.
- Closes #316

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/monitors/classical.py
+++ b/glide/monitors/classical.py
@@ -5,7 +5,7 @@ from numpy.typing import NDArray
 
 from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
 from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
-from glide.core.validation import _validate_bounds, _validate_equal_lengths, _validate_min_samples
+from glide.core.validation import _validate_bounds, _validate_equal_lengths, _validate_has_no_nan, _validate_non_empty
 from glide.mean_monitoring_results import ClassicalMeanMonitoringResult
 from glide.monitors.core import _scale_from_unit_risk, _scale_to_unit_risk, _unique_ordered_batches
 
@@ -79,6 +79,8 @@ class ClassicalMeanMonitor:
         metric_lower_bound: float,
         metric_upper_bound: float,
     ) -> Tuple[NDArray, float, NDArray, NDArray]:
+        _validate_non_empty(y, "y")
+        _validate_has_no_nan(batches, "batches")
         _validate_bounds(
             confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
         )
@@ -105,7 +107,6 @@ class ClassicalMeanMonitor:
         _validate_equal_lengths(y, batches, names=["y", "batches"])
         labeled_mask = ~np.isnan(y)
         labeled_values = y[labeled_mask]
-        _validate_min_samples(labeled_values, "y")
         _validate_bounds(
             labeled_values,
             "y",
@@ -224,6 +225,7 @@ class ClassicalMeanMonitor:
         Raises
         ------
         ValueError
+            - If ``y`` is empty.
             - If ``y`` and ``batches`` have different lengths.
             - If any labeled value falls outside ``[metric_lower_bound, metric_upper_bound]``.
             - If ``batches`` contains NaN values.
--- a/glide/monitors/classical.py
+++ b/glide/monitors/classical.py
@@ -75,9 +75,33 @@ class ClassicalMeanMonitor:
         batches: NDArray,
         higher_is_better: bool,
         threshold: float,
+        confidence_level: float,
         metric_lower_bound: float,
         metric_upper_bound: float,
     ) -> Tuple[NDArray, float, NDArray, NDArray]:
+        _validate_bounds(
+            confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
+        )
+        _validate_bounds(
+            metric_lower_bound,
+            "metric_lower_bound",
+            upper=metric_upper_bound,
+            right_inclusive=False,
+            error_message=(
+                f"'metric_lower_bound' must be strictly smaller than 'metric_upper_bound'; "
+                f"got {metric_lower_bound!r} and {metric_upper_bound!r}."
+            ),
+        )
+        _validate_bounds(
+            threshold,
+            "threshold",
+            lower=metric_lower_bound,
+            upper=metric_upper_bound,
+            error_message=(
+                f"'threshold' must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
+                f"'metric_upper_bound'={metric_upper_bound!r}."
+            ),
+        )
         _validate_equal_lengths(y, batches, names=["y", "batches"])
         labeled_mask = ~np.isnan(y)
         labeled_values = y[labeled_mask]
@@ -209,32 +233,8 @@ class ClassicalMeanMonitor:
             - If ``threshold`` falls outside ``[metric_lower_bound, metric_upper_bound]``.
             - If ``confidence_level`` is not in (0, 1).
         """
-        _validate_bounds(
-            confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
-        )
-        _validate_bounds(
-            metric_lower_bound,
-            "metric_lower_bound",
-            upper=metric_upper_bound,
-            right_inclusive=False,
-            error_message=(
-                f"'metric_lower_bound' must be strictly smaller than 'metric_upper_bound'; "
-                f"got {metric_lower_bound!r} and {metric_upper_bound!r}."
-            ),
-        )
-        _validate_bounds(
-            threshold,
-            "threshold",
-            lower=metric_lower_bound,
-            upper=metric_upper_bound,
-            error_message=(
-                f"'threshold' must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
-                f"'metric_upper_bound'={metric_upper_bound!r}."
-            ),
-        )
-
         risk_y, risk_threshold, batch_codes, batch_n = self._preprocess(
-            y, batches, higher_is_better, threshold, metric_lower_bound, metric_upper_bound
+            y, batches, higher_is_better, threshold, confidence_level, metric_lower_bound, metric_upper_bound
         )
         batch_sums = np.bincount(batch_codes, weights=risk_y)
         risk_batch_mean_estimates = batch_sums / batch_n
--- a/glide/monitors/classical.py
+++ b/glide/monitors/classical.py
@@ -1,0 +1,269 @@
+from typing import Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
+from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
+from glide.core.validation import _validate_bounds, _validate_equal_lengths, _validate_min_samples
+from glide.mean_monitoring_results import ClassicalMeanMonitoringResult
+from glide.monitors.core import _scale_from_unit_risk, _scale_to_unit_risk, _unique_ordered_batches
+
+
+class ClassicalMeanMonitor:
+    """Anytime-valid drift monitor over a batched dataset of labels.
+
+    It uses the plain sample mean per batch. Unlabeled entries are passed as
+    ``np.nan`` and dropped per batch. To be used on accumulated non-overlapping
+    production batches by calling :meth:`detect` on the whole accumulated dataset at
+    any time. Data is passed oldest batch first, and every batch is monitored. The
+    monitor computes a sample-mean estimate per batch, builds an anytime-valid
+    confidence sequence on the running mean of those estimates, and raises a drift
+    alarm when it crosses a user-supplied ``threshold`` (the worst tolerable metric value).
+    Because the bounds are valid at all times simultaneously, :meth:`detect` may be called
+    after every new batch without inflating the false-alarm probability.
+
+    Notes
+    -----
+    The confidence sequence is built over the per-batch sample means as its
+    atoms, so the monitored quantity is the equal-weighted running mean of the
+    per-batch estimates. This matches the way human annotations arrive in
+    practice (in batches, not as a per-sample stream) and equals a per-sample
+    running mean only when every batch shares the same labeled count. The bound
+    width splits into two parts: the underlying statistical precision, set by the
+    total number of labels (the usual root-n term, unaffected by how the labels
+    are split across batches), plus the anytime-valid overhead, the price for
+    being allowed to peek after every batch, which decays as one over the number
+    of batches. Both parts shrink as data accumulates; the overhead is simply
+    what keeps the earliest looks wide, so no alarm can fire in the first few
+    batches. This is the expected, modest cost of valid repeated looking, not a
+    loss of sample efficiency.
+
+    References
+    ----------
+    Howard, Steven R., Aaditya Ramdas, Jon McAuliffe, and Jasjeet Sekhon. "Time-uniform,
+    nonparametric, nonasymptotic confidence sequences." The Annals of Statistics 49,
+    no. 2 (2021): 1055-1080.
+
+    Podkopaev, Aleksandr, and Aaditya Ramdas. "Tracking the risk of a deployed model
+    and detecting harmful distribution shifts." International Conference on Learning
+    Representations (ICLR), 2022.
+
+    Waudby-Smith, Ian, and Aaditya Ramdas. "Estimating means of bounded random
+    variables by betting." Journal of the Royal Statistical Society Series B:
+    Statistical Methodology 86, no. 1 (2024): 1-27.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from glide.monitors import ClassicalMeanMonitor
+    >>> pre_drift_batch = np.array([0.0, 0.2, np.nan, np.nan])
+    >>> post_drift_batch = np.array([0.8, 1.0, np.nan, np.nan])
+    >>> y = np.hstack([pre_drift_batch, np.tile(post_drift_batch, 50)])
+    >>> batches = np.repeat(np.arange(51), 4)
+    >>> monitor = ClassicalMeanMonitor()
+    >>> result = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
+    >>> result.drift_detected
+    True
+    >>> result.first_alarm_index
+    11
+    """
+
+    def _preprocess(
+        self,
+        y: NDArray,
+        batches: NDArray,
+        higher_is_better: bool,
+        threshold: float,
+        metric_lower_bound: float,
+        metric_upper_bound: float,
+    ) -> Tuple[NDArray, float, NDArray, NDArray]:
+        _validate_equal_lengths(y, batches, names=["y", "batches"])
+        labeled_mask = ~np.isnan(y)
+        labeled_values = y[labeled_mask]
+        _validate_min_samples(labeled_values, "y")
+        _validate_bounds(
+            labeled_values,
+            "y",
+            lower=metric_lower_bound,
+            upper=metric_upper_bound,
+            error_message=(
+                f"'y' values must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
+                f"'metric_upper_bound'={metric_upper_bound!r}; got values in "
+                f"[{labeled_values.min()!r}, {labeled_values.max()!r}]."
+            ),
+        )
+        batches_labeled = batches[labeled_mask]
+        batch_identifiers, batch_codes = _unique_ordered_batches(batches_labeled)
+        batch_n = np.bincount(batch_codes)
+        worst_batch_position = np.argmin(batch_n)
+        _validate_bounds(
+            batch_n[worst_batch_position],
+            "y",
+            lower=2,
+            error_message=(
+                f"'y' must have at least 2 non-NaN values per batch; got {batch_n[worst_batch_position]} "
+                f"in batch '{batch_identifiers[worst_batch_position]}'."
+            ),
+        )
+        risk_y = _scale_to_unit_risk(labeled_values, metric_lower_bound, metric_upper_bound, higher_is_better)
+        risk_threshold = _scale_to_unit_risk(threshold, metric_lower_bound, metric_upper_bound, higher_is_better)
+        return risk_y, risk_threshold, batch_codes, batch_n
+
+    def _postprocess(
+        self,
+        risk_running_means: NDArray,
+        risk_confidence_bounds: NDArray,
+        risk_batch_mean_estimates: NDArray,
+        higher_is_better: bool,
+        metric_lower_bound: float,
+        metric_upper_bound: float,
+    ) -> Tuple[NDArray, NDArray, NDArray]:
+        running_means = _scale_from_unit_risk(
+            risk_running_means, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        confidence_bounds = _scale_from_unit_risk(
+            risk_confidence_bounds, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        batch_mean_estimates = _scale_from_unit_risk(
+            risk_batch_mean_estimates, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        return running_means, confidence_bounds, batch_mean_estimates
+
+    def detect(
+        self,
+        y: NDArray,
+        batches: NDArray,
+        higher_is_better: bool,
+        threshold: float,
+        metric_name: str = "Metric",
+        confidence_level: float = 0.8,
+        metric_lower_bound: float = 0.0,
+        metric_upper_bound: float = 1.0,
+    ) -> ClassicalMeanMonitoringResult:
+        """Detect a drift of the running mean across a batched dataset.
+
+        Splits the data by batch, computes a sample-mean estimate per batch, and
+        builds an anytime-valid empirical-Bernstein confidence sequence on the
+        running mean of those estimates. An alarm is raised at every batch where the
+        sequence crosses the user-supplied ``threshold``.
+
+        Rows must be ordered oldest batch first and grouped into contiguous blocks;
+        identifier values are not compared, so any label type works. Batches must be
+        non-overlapping (no shared samples), and successive calls must be made on
+        growing histories of the same data; passing the full accumulated dataset at
+        every call makes the anytime-valid guarantee hold jointly over all calls.
+        Alternatively the data may be restricted to the most recent batches, in which
+        case the guarantee holds within each restriction but not across the moving
+        history as a whole.
+
+        Parameters
+        ----------
+        y : NDArray
+            Array of observations, shape ``(n_samples,)``. Unlabeled entries are
+            ``np.nan`` and are dropped per batch; every batch must keep at least 2
+            labeled entries.
+        batches : NDArray
+            Array of batch identifiers, shape ``(n_samples,)``. Rows must be ordered
+            oldest batch first and grouped into contiguous blocks. Identifier values
+            are not compared, so any hashable label type works (integers, dates,
+            free-form strings).
+        higher_is_better : bool
+            ``False`` when the metric is a risk (drift means the metric increased),
+            ``True`` when it is a performance (drift means the metric decreased).
+        threshold : float
+            The metric value the running mean is monitored against, in metric units:
+            the worst level the user is willing to tolerate. An alarm fires once the
+            anytime-valid bound proves the running metric has crossed it (the running
+            risk exceeds it for a risk, the running performance falls below it for a
+            performance). Must lie within ``[metric_lower_bound, metric_upper_bound]``.
+        metric_name : str, optional
+            Human-readable label for the metric. Defaults to ``"Metric"``.
+        confidence_level : float, optional
+            How confident each alarm should be. At the default ``0.8`` the monitor
+            raises "80% confident" alarms: under no drift it has at most a 20% chance
+            of raising a false alarm. Raising this value demands more evidence before
+            alarming, so alarms become more trustworthy but arrive later; lowering it
+            detects sooner at the cost of more false alarms.
+        metric_lower_bound : float, optional
+            Known lower bound of the metric. Defaults to ``0.0``.
+        metric_upper_bound : float, optional
+            Known upper bound of the metric. Defaults to ``1.0``.
+
+        Returns
+        -------
+        ClassicalMeanMonitoringResult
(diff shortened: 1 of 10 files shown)


imerad:
### Code review

Found 4 issues:

1. Batches with zero labeled samples are silently dropped instead of raising the documented error

`_preprocess` filters `y`/`batches` down to labeled rows (`labeled_mask = ~np.isnan(y)`) *before* computing batch chronology and counts. A batch that is entirely unlabeled therefore never appears in `batch_identifiers`/`batch_codes`/`batch_n` at all, so the "at least 2 non-NaN values per batch" check (and the docstring's `Raises` promise "If any batch has fewer than 2 labeled (non-NaN) samples") never fires for it — the batch is just skipped, and `detect` returns one fewer monitored batch with no error. Reproduced locally: `y=[nan, nan, 0.1, 0.2, 0.3, 0.4]`, `batches=[0, 0, 1, 1, 2, 2]` runs successfully and only monitors batches 1 and 2 (`batch_n == [2, 2]`), even though batch 0 has zero labels. This contradicts "every batch is monitored" (used elsewhere to justify prefix-consistency) and can silently hide a labeling-pipeline outage for a whole production batch.

https://github.com/EmertonData/glide/blob/1af29b20851fa170323264dc5a5c8f46bff0c96e/glide/monitors/classical.py#L105-L132

2. Private, underscore-prefixed helper imported across package boundaries

`_compute_empirical_bernstein_bounds` is imported directly from `glide.confidence_sequences.empirical_bernstein` into `glide.monitors.classical`. This is exactly the gap flagged on [PR #331](https://github.com/EmertonData/glide/pull/331#issuecomment-4874587801)'s review: "there's no public, supported way to go from batch data to a `ConfidenceSequence` instance — a caller would need to import a private, underscore-prefixed function," with a suggested fix of a classmethod like `from_batch_estimates(...)` on `EmpiricalBernsteinConfidenceSequence`. That fix was never made, and this PR is the first production caller to actually reach across the package boundary for the private helper instead of the class's own constructor.

https://github.com/EmertonData/glide/blob/1af29b20851fa170323264dc5a5c8f46bff0c96e/glide/monitors/classical.py#L6-L8

3. Per-batch minimum-sample check duplicates the existing `_validate_min_samples` helper

`glide/core/validation.py` already centralizes this exact check via `_validate_min_samples(values, name, stratum_id)`, used as-is for the analogous per-stratum case in `StratifiedClassicalMeanEstimator`. Here it's reimplemented by hand with `_validate_bounds(batch_n[worst_batch_position], "y", lower=2, error_message=...)`, duplicating the "at least 2 non-NaN values per <group>" logic and message format. Per CLAUDE.md's "No Redundancy" and "Consistency and Propagation" sections this should reuse the shared helper (e.g. by validating the worst batch's actual labeled slice with `stratum_id=batch_identifiers[worst_batch_position]`) rather than re-deriving the same check.

https://github.com/EmertonData/glide/blob/1af29b20851fa170323264dc5a5c8f46bff0c96e/glide/monitors/classical.py#L120-L132

Compare to the existing pattern:

https://github.com/EmertonData/glide/blob/1af29b20851fa170323264dc5a5c8f46bff0c96e/glide/estimators/stratified_classical.py#L94-L99

4. No unit test exercises any of the `Raises` paths end-to-end

`detect`'s docstring lists 7 `ValueError` conditions, but `test_classical.py` only has a `test_preprocess_delegates_to_validation` test that mocks out `_validate_bounds`/`_validate_min_samples`/`_unique_ordered_batches` and asserts on call arguments — none of the real raise paths (including the "fewer than 2 labeled samples per batch" one from issue 1) are exercised with real data. This is likely why issue 1 wasn't caught.

https://github.com/EmertonData/glide/blob/1af29b20851fa170323264dc5a5c8f46bff0c96e/tests/unit/monitors/test_classical.py#L27-L96

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

glide/monitors/classical.py
```diff
@@ -0,0 +1,255 @@
...
+    >>> pre_drift_batch = np.array([0.0, 0.2, np.nan, np.nan])
+    >>> post_drift_batch = np.array([0.8, 1.0, np.nan, np.nan])
+    >>> y = np.hstack([pre_drift_batch, np.tile(post_drift_batch, 50)])
+    >>> batches = np.repeat(np.arange(51), 4)
+    >>> monitor = ClassicalMeanMonitor()
+    >>> result = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
+    >>> result.drift_detected
+    True
+    >>> result.first_alarm_index
+    11
```
gmartinon-ed, line 53: Il est trop fort ce Claude, il avait déjà prédit 11 dans le ticket !

glide/monitors/core.py
```diff
@@ -0,0 +1,58 @@
+from typing import Tuple, Union, overload
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def _unique_ordered_batches(batches: NDArray) -> Tuple[NDArray, NDArray]:
+    block_starts = np.ones(len(batches), dtype=bool)
+    block_starts[1:] = batches[1:] != batches[:-1]
```
gmartinon-ed, line 9: Plus efficace que le argsort ?

glide/monitors/core.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@overload
+def _scale_to_unit_risk(
+    values: float, metric_lower_bound: float, metric_upper_bound: float, higher_is_better: bool
+) -> float: ...
+@overload
+def _scale_to_unit_risk(
+    values: NDArray, metric_lower_bound: float, metric_upper_bound: float, higher_is_better: bool
+) -> NDArray: ...
```
gmartinon-ed, line 28: Pourquoi on a besoin de ces overloads ?
On a déjà `values: Union[float, NDArray]` en dessous

glide/monitors/core.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@overload
+def _scale_from_unit_risk(
+    values: float, metric_lower_bound: float, metric_upper_bound: float, higher_is_better: bool
+) -> float: ...
+@overload
+def _scale_from_unit_risk(
+    values: NDArray, metric_lower_bound: float, metric_upper_bound: float, higher_is_better: bool
+) -> NDArray: ...
```
gmartinon-ed, line 48: idem

tests/unit/monitors/test_core.py
```diff
@@ -0,0 +1,72 @@
...
+
+# --- _unique_ordered_batches ---
+
+
+def test_unique_ordered_batches_integers():
+    batches = np.array([2, 2, 0, 0, 0, 1])
+    batch_identifiers, batch_codes = _unique_ordered_batches(batches)
+
+    expected_batch_identifiers = np.array([2, 0, 1])
+    expected_batch_codes = np.array([0, 0, 1, 1, 1, 2])
```
gmartinon-ed, line 23: On place plutôt les expected avant le call de fonction par convention : setup / execution / assertion

gmartinon-ed (changes requested):

glide/monitors/classical.py
imerad: Je pense qu'il l'avait simulé :)

glide/monitors/core.py
imerad: Je m'étais posé la même question. Il s'avère que sans ces overloads, ty ne peut pas deviner qu'un float devrait mener à un float en sortie et un NDArray à un NDArray. Donc le type-check ne passerait plus

glide/monitors/core.py
imerad: Voir réponse précédente

glide/monitors/core.py
imerad: Argsort serait en n log(n) hors ceci est linéaire, donc oui c'est plus efficace. On fait quand même appel à `np.unique` plus tard, mais sur un tableau (a priori) bien plus petit

tests/unit/monitors/test_core.py
imerad: Traité

gmartinon-ed (approved):

PR #335, opened 2026-07-07
Add PPIMeanMonitor for anytime-valid drift detection (PPRM)

### LINKED ISSUE #317: PPIMeanMonitor: anytime-valid drift detection over batched production data (PPRM)
# TODO :

Investigate if we can vectorize the batch_mean_estimate with `np.bincount(..., weight=...)`

# PPIMeanMonitor: anytime-valid drift detection over batched production data (PPRM)

## Background

After a GenAI system is deployed, its quality metric is typically re-estimated on successive batches of production data, for instance by collecting a fresh batch every week with a few human labels and many proxy labels. The natural question is then: "does this sequence of batches prove that the metric has drifted past an acceptable business threshold?". Comparing each new batch estimate to the threshold with an ordinary confidence interval is statistically invalid: each comparison carries its own false-alarm probability, so checking after every batch makes a false alarm almost certain over time (this is the classical "multiple testing" problem, here called *peeking*).

Zhang et al. solve this with PPRM (Prediction-Powered Risk Monitoring). The tool is a *confidence sequence*: a sequence of confidence bounds that holds simultaneously at all times, so the user may look after every batch and the total false-alarm probability stays below a chosen budget. PPRM builds an anytime-valid bound on the *running mean* (the average of the per-batch metric values up to now) from per-batch PPI estimates, using an empirical-Bernstein bound (a concentration bound whose width adapts to the empirically observed variability of the estimates, instead of assuming a known variance). In math notation, for a risk (lower is better), the alarm rule at batch t is: raise drift if `L_t > threshold`, where `L_t` is the anytime-valid lower bound on the running risk after t batches and `threshold` is a user-supplied value (the worst running risk the user is willing to tolerate, in metric units). The guarantee is that under no drift (the running risk stays at or below the threshold), the probability of *ever* raising a false alarm is at most the single miscoverage budget `1 - confidence_level`.

## Design choices

The monitor lives in `glide/monitors/ppi.py` (class `PPIMeanMonitor`, with PPI-specific internals in `glide/monitors/ppi_core.py`), mirroring `glide/estimators/ppi.py` in structure (a `_preprocess`/`_postprocess` pair around a single public `detect` method; the symmetric `_postprocess` is new relative to the estimators, warranted by the round-trip risk-orientation scaling the monitor adds). This ticket adds `PPIMeanMonitor`, extends `glide/monitors/__init__.py`, and adds `PPIMeanMonitor` to `glide/__init__.py`.

The central interface decision, requested explicitly: GLIDE must not orchestrate data collection, so the monitor is **stateless** and consumes raw data. The signature `detect(y_true, y_proxy, batches, higher_is_better, threshold)` extends `PPIMeanEstimator.estimate(y_true, y_proxy)` exactly the way the stratified estimators extend it with `groups`: `batches` assigns each sample to a batch. Chronology rides on row order rather than identifier values, so the data is passed oldest batch first. Identifier values are never compared, so any label type works (integers, dates, free-form strings), sidestepping the `"batch10" < "batch2"` lexicographic trap. Unlike the order-free `groups` of the stratified estimators, `batches` is therefore order-significant — honest to the time-series nature of monitoring — and each batch must occupy a contiguous block of rows, which `_preprocess` enforces (interleaved rows signal unsorted input and raise). The user accumulates production data however they like and calls `detect` whenever they like on the whole accumulated dataset; anytime validity is exactly what makes repeated calls on growing histories jointly valid. This design keeps coupling minimal: the inference result dataclasses and `glide/io` are untouched, and the monitor reuses the existing `_compute_tuning_parameter` and `_compute_mean_estimate` helpers from `glide/estimators/ppi_core.py` and the method-agnostic `_split_labeled_unlabeled` from `glide/estimators/core.py`, leaving existing estimators unchanged. It also reuses the shared confidence-sequence builder `_compute_empirical_bernstein_bounds` (and the `EmpiricalBernsteinConfidenceSequence` value object the result embeds) from `glide/confidence_sequences/`, plus the monitor primitives `_scale_to_unit_risk`, `_scale_from_unit_risk` and `_unique_ordered_batches` from `glide/monitors/core.py`, so `_preprocess`, `_postprocess` and `detect` add no batch-chronology, boundary or scaling logic of their own: `_preprocess` calls the scale-in helper, `_postprocess` calls the scale-out helper, and `detect` orchestrates.

Owning the raw data lets the monitor implement the paper's power tuning faithfully: for each batch t, the tuning parameter is computed on *all previous* batches (pooling their observations), never on batch t itself, which makes it *predictable* as required by the paper's unbiasedness lemma. The very first batch has no predecessor, so it falls back to the classic PPI weight of 1 (a constant is trivially predictable); power tuning kicks in from the second batch onward. Predictability is the validity-critical part; the choice of *which* prior batches to pool is pure power tuning (any predictable choice is valid, only the width changes), so we use the full prior history rather than the paper's sliding window (a deliberate adaptation detailed below). The tuning parameter is clipped to `[0, max_tuning_parameter]`, and this cap is what makes the normalization well-defined rather than a cosmetic guard: with a metric in [0, 1] a PPI estimate provably lies in `[-max_tuning_parameter, 1 + max_tuning_parameter]`, so the paper's affine map `(x + max_tuning_parameter) / (1 + 2·max_tuning_parameter)` sends it exactly onto [0, 1], which the empirical-Bernstein bound requires (it is only valid for variables in a known bounded range). The clip does real work because the power-tuning parameter is a variance-minimizing regression slope (`Cov(y_true, y_proxy) / Var(y_proxy)` up to a finite-sample factor), not a mixing weight confined to [0, 1]: an under-dispersed proxy drives it above 1 and an anti-correlated proxy drives it below 0, so the standard PPI++ formula (which GLIDE leaves unclipped in the estimators) can return any real value. Without the cap, the tuning ratio can leave [0, 1] on a strong or perverse proxy, the estimate leaves any fixed interval, and the normalization breaks. Clipping to [0, max_tuning_parameter] keeps the per-batch estimate in the fixed range the bound needs, with the lower 0 falling back to the classical estimate when the proxy is anti-correlated. The trade-off: a larger cap permits more proxy reliance but widens the denominator, so the un-normalized confidence sequence scales up and detection slows. The `higher_is_better` boolean handles performance metrics by internally flipping the metric into "risk orientation" (lower is better) so a single code path serves both cases; the threshold is flipped on the same path, so the alarm rule `L_t > threshold` holds in risk space for both metric orientations.

The empirical-Bernstein variance process needs a *predictable* center before the first batch arrives (the seed `c_0`). With no reference batch to estimate it from, the monitor seeds with the threshold itself (mapped into the normalized space the bound runs in): it is the one meaningful constant available before any data, it is predictable by construction, and seeding at the worst tolerated value is mildly conservative (it can only widen the first variance term, never break validity). The seed influences only the first look's width and washes out as batches accumulate.

Two deliberate adaptations of the paper:

- the mixture density over the betting parameter is the uniform density on (0, 1): any probability density yields a valid bound, and the uniform one is parameter-free and horizon-agnostic. (A conjugate gamma mixture would give a closed-form, somewhat tighter boundary, but only at the price of a sharpness-tuning hyperparameter that a non-expert cannot set intuitively and that affects detection speed only, never validity;
- the power-tuning parameter pools the **full prior history** (all earlier batches), where the paper uses a fixed-length sliding window. We deliberately drop the window and keep full-history tuning as the only mode. It is parameter-free (the window length is a knob no non-expert can set), but the deeper reason is that a windowed λ is internally inconsistent: it would weight the *tuning parameter* by recency while the *proxy labels* it is computed from carry no matching weighting, half-correcting for a drifting proxy without down-weighting the stale labels themselves. The situation the window targets — the proxy's own calibration drifting over time, e.g. an LLM-as-judge that keeps improving — is handled coherently by the stateless API instead: the user reissues `detect` on a recent slice (the most recent batches), which resets *both* the tuning parameter and the labels it sees to the current regime. That restriction is itself a valid ordered, non-overlapping sequence, so the anytime-valid guarantee holds within it (the same escape hatch the caveats below describe for stale-history dilution).

Caveats that the docstring must state because the stateless API cannot fully enforce them: rows must be ordered oldest batch first (the contiguity check catches interleaved/shuffled input loudly, but contiguous blocks in the wrong time order cannot be detected and would silently misorder the history), batches must be non-overlapping (no shared samples), and successive calls must be made on growing histories of the same data (no cherry-picking of batches, which would break the martingale guarantee). The method detects drifts of the *running average*, so detection is inherently multi-batch: evidence must accumulate for the lower bound to escalate, and a short transient drift gets diluted as more batches arrive. Restricting the data to the most recent batches removes the dilution from a long stale history (it makes the running mean reflect the current regime), and in that case it genuinely *accelerates* detection: a sustained recent shift that an unbounded history would dilute below the threshold can escalate within a few batches once the stale mass is dropped. What windowing does **not** do is enable detection of an abrupt step between two consecutive batches: the restricted history restarts the confidence sequence at its widest, so a minimum number of batches is still required before any alarm can fire. Single-batch step detection is a different problem (a two-sample / change-point comparison, which would use the per-batch PPI estimates and their standard errors rather than the running mean) and is out of scope for this monitor.

### Result dataclass (prerequisite)

`PPIMeanMonitor.detect` returns `PredictionPoweredMeanMonitoringResult`, a thin subclass of the shared `MeanMonitoringResult` base. Both are implemented by the prerequisite ticket `feat-mean-monitoring-results.md`; this ticket only imports and returns the class.

### PPI-specific internal helpers

**Location:** `glide/monitors/ppi_core.py`

**Purpose:** The *proxy-specific* primitives for the monitor, mirroring how `glide/estimators/ppi_core.py` holds the estimator internals as free functions. What remains PPI-specific is the clipped power-tuning parameter and the affine estimate normalization that the proxy makes necessary.

```python
from typing import Union

import numpy as np

from glide.estimators.ppi_core import _compute_tuning_parameter


def _compute_clipped_tuning_parameter(
    y_true_labeled: NDArray,
    y_proxy_labeled: NDArray,
    y_proxy_unlabeled: NDArray,
    power_tuning: bool,
    max_tuning_parameter: float,
) -> float:
    tuning_parameter = _compute_tuning_parameter(y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, power_tuning)
    clipped_tuning_parameter = min(max(tuning_parameter, 0.0), max_tuning_parameter)
    return clipped_tuning_parameter


def _normalize_to_unit_interval(
    values: Union[float, NDArray],
    max_tuning_parameter: float,
) -> Union[float, NDArray]:
    normalization = 1.0 + 2.0 * max_tuning_parameter
    normalized = (values + max_tuning_parameter) / normalization
    return normalized


def _denormalize_from_unit_interval(
    values: Union[float, NDArray],
    max_tuning_parameter: float,
) -> Union[float, NDArray]:
    normalization = 1.0 + 2.0 * max_tuning_parameter
    original = values * normalization - max_tuning_parameter
    return original
```

### PPIMeanMonitor

**Location:** `glide/monitors/ppi.py`

**Purpose:** Stateless anytime-valid drift detector over a batched dataset of true and proxy labels, mirroring `PPIMeanEstimator`.

```python

import numpy as np

from glide.core.validation import (
    _validate_bounds,
    _validate_equal_lengths,
    _validate_sample_sizes,
    _validate_y_proxy,
    _validate_y_true,
)
from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
from glide.estimators.ppi_core import _compute_mean_estimate
from glide.mean_monitoring_results import PredictionPoweredMeanMonitoringResult
from glide.monitors.core import (
    _unique_ordered_batches,
    _scale_from_unit_risk,
    _scale_to_unit_risk,
)
from glide.monitors.ppi_core import (
    _compute_clipped_tuning_parameter,
    _denormalize_from_unit_interval,
    _normalize_to_unit_interval,
)


class PPIMeanMonitor:
    """Anytime-valid drift monitor over a batched dataset of true and proxy labels.

    This class implements the PPRM method. The user accumulates non-overlapping
    production batches (each with a few true labels and many proxy labels) and calls
    :meth:`detect` on the whole accumulated dataset at any time. Data is passed oldest
    batch first, and every batch is monitored. The monitor computes a
    prediction-powered estimate per batch with a power-tuning parameter computed only
    from past batches, builds an anytime-valid confidence sequence on the running mean
    of those estimates, and raises a drift alarm when it crosses a user-supplied
    ``threshold`` (the worst metric value the user is willing to tolerate). Because the
    bounds are valid at all times simultaneously, the user may call :meth:`detect`
    after every new batch without inflating the false-alarm probability.

    References
    ----------
    Zhang, Guangyi, Yunlong Cai, Guanding Yu, and Osvaldo Simeone. "Prediction-powered
    risk monitoring of deployed models for detecting harmful distribution shifts."
    arXiv preprint arXiv:2602.02229 (2026).

    no. 2 (2021): 1055-1080.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.monitors import PPIMeanMonitor
    >>> pre_drift_y_true = np.array([0.0, 0.2, np.nan, np.nan])
    >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
    >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
    >>> post_drift_y_proxy = np.array([0.8, 1.0, 0.8, 1.0])
    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
    >>> monitor = PPIMeanMonitor()
    >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
    >>> result.drift_detected
    True
    >>> result.first_alarm_index
    22
    """

    def _preprocess(
        self,
        y_true: NDArray,
        y_proxy: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
    ) -> Tuple[NDArray, NDArray, float, NDArray]:
        _validate_equal_lengths(y_true, y_proxy, batches, names=["y_true", "y_proxy", "batches"])
        _validate_y_proxy(y_proxy)
        _validate_y_true(y_true)
        labeled_values = y_true[labeled_mask]
        _validate_bounds(
            labeled_values,
            "y_true",
            lower=metric_lower_bound,
            upper=metric_upper_bound,
            error_message=f"'y_true' values must lie within [{metric_lower_bound!r}, {metric_upper_bound!r}].",
        )
        _validate_bounds(
            y_proxy,
            "y_proxy",
            lower=metric_lower_bound,
            upper=metric_upper_bound,
            error_message=f"'y_proxy' values must lie within [{metric_lower_bound!r}, {metric_upper_bound!r}].",
        )
        batch_identifiers = _unique_ordered_batches(batches)
        for identifier in batch_identifiers:
            batch_mask = batches == identifier
            batch_labeled_mask = labeled_mask[batch_mask]
            _validate_sample_sizes(batch_labeled_mask)
        risk_y_true = _scale_to_unit_risk(y_true, metric_lower_bound, metric_upper_bound, higher_is_better)
        risk_y_proxy = _scale_to_unit_risk(y_proxy, metric_lower_bound, metric_upper_bound, higher_is_better)
        risk_threshold = _scale_to_unit_risk(threshold, metric_lower_bound, metric_upper_bound, higher_is_better)
        return risk_y_true, risk_y_proxy, risk_threshold, batch_identifiers

    def _postprocess(
        self,
        risk_running_means: NDArray,
        risk_lower_bounds: NDArray,
        risk_batch_estimates: NDArray,
        higher_is_better: bool,
        running_means = _scale_from_unit_risk(risk_running_means, metric_lower_bound, metric_upper_bound, higher_is_better)
        confidence_bounds = _scale_from_unit_risk(risk_lower_bounds, metric_lower_bound, metric_upper_bound, higher_is_better)
        batch_estimates = _scale_from_unit_risk(risk_batch_estimates, metric_lower_bound, metric_upper_bound, higher_is_better)
        return running_means, confidence_bounds, batch_estimates

    def detect(
        self,
        y_true: NDArray,
        y_proxy: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
        metric_name: str = "Metric",
        power_tuning: bool = True,
        max_tuning_parameter: float = 1.0,

        Splits the data by batch, computes a prediction-powered estimate per batch
        with a power-tuning parameter computed only from past batches, and builds an
        anytime-valid empirical-Bernstein confidence sequence on the running mean of
        those estimates. An alarm is raised at every batch where the sequence crosses
        the user-supplied ``threshold``.

        Rows must be ordered oldest batch first and grouped into contiguous blocks;
        identifier values are not compared, so any label type works. Batches must be
        non-overlapping (no shared samples), and successive calls must be made on
        growing histories of the same data; passing the full accumulated dataset at
        every call makes the anytime-valid guarantee hold jointly over all calls.
        Alternatively the data may be restricted to the most recent batches, in which
        case the guarantee holds within each restriction but not across the moving
        history as a whole.

        Parameters
        ----------
        y_true : NDArray
            Array of labeled observations, shape ``(n_samples,)``.
        y_proxy : NDArray
        batches : NDArray
            Array of batch identifiers, shape ``(n_samples,)``. Rows must be ordered
            oldest batch first and grouped into contiguous blocks. Identifier values
            are not compared, so any hashable label type works (integers, dates,
            free-form strings).
        higher_is_better : bool
            ``False`` when the metric is a risk (drift means the metric increased),
            ``True`` when it is a performance (drift means the metric decreased).
        threshold : float
            The metric value the running mean is monitored against, in metric units:
            the worst level the user is willing to tolerate. An alarm fires once the
            anytime-valid bound proves the running metric has crossed it (the running
            risk exceeds it for a risk, the running performance falls below it for a
            performance). Must lie within ``[metric_lower_bound, metric_upper_bound]``.
        metric_name : str, optional
            How confident each alarm should be. At the default ``0.8`` the monitor
            raises "80% confident" alarms: under no drift it has at most a 20% chance
            of *ever* raising a false alarm, no matter how often you look. Raising it
            (say to ``0.95``) demands more evidence before alarming, so alarms become
            more trustworthy but arrive later; lowering it detects sooner at the cost
            of more false alarms. The default is looser than GLIDE's fixed-sample
            ``0.95`` because anytime-valid bounds are conservative, so ``0.8`` keeps
            detection timely.
        power_tuning : bool, optional
            If ``True`` (default), compute the power-tuning parameter of each batch on
            all previous batches (the first batch, having no predecessor, uses 1). If
            ``False``, use 1 everywhere (classic PPI).
        metric_lower_bound : float, optional
            Known lower bound of the metric. Defaults to ``0.0``.
        metric_upper_bound : float, optional
            Known upper bound of the metric. Defaults to ``1.0``.
        max_tuning_parameter : float, optional
            Upper clip for the power-tuning parameter. Larger values allow more
            reliance on the proxy but widen the confidence sequence through the
            affine normalization. Defaults to ``1.0``.

        Returns
        -------
        PredictionPoweredMeanMonitoringResult
            Per-batch estimates, running means, anytime-valid confidence bounds,
            alarm flags, and the alarm threshold, all in the original metric
            orientation.

        Raises
        ------
        ValueError
            - If ``y_true``, ``y_proxy`` and ``batches`` have different lengths.
            - If any proxy value is NaN.
            - If any value falls outside ``[metric_lower_bound, metric_upper_bound]``.
            - If ``batches`` is empty (no batch to monitor).
            - If batches are interleaved rather than grouped into contiguous blocks.
            - If any batch has fewer than 2 labeled or fewer than 2 unlabeled samples.
            - If proxy values are constant across the prior batches (with ``power_tuning=True``).
            - If ``metric_lower_bound >= metric_upper_bound``.
            - If ``threshold`` falls outside ``[metric_lower_bound, metric_upper_bound]``.
            - If ``confidence_level`` is not in (0, 1).
            - If ``max_tuning_parameter`` is not strictly positive.
        """
        _validate_bounds(
            confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
        )
        _validate_bounds(max_tuning_parameter, "max_tuning_parameter", lower=0, left_inclusive=False)
        _validate_bounds(
            metric_lower_bound,
            "metric_lower_bound",
            upper=metric_upper_bound,
            right_inclusive=False,
            error_message=(
                f"'metric_lower_bound' must be strictly smaller than 'metric_upper_bound'; "
                f"got {metric_lower_bound!r} and {metric_upper_bound!r}."
            ),
        )
        _validate_bounds(threshold, "threshold", lower=metric_lower_bound, upper=metric_upper_bound)
        risk_y_true, risk_y_proxy, risk_threshold, batch_identifiers = self._preprocess(
            y_true, y_proxy, batches, higher_is_better, threshold, metric_lower_bound, metric_upper_bound
        )

        risk_batch_estimates_list: List[float] = []
        batch_n_true_list: List[int] = []
        batch_n_proxy_list: List[int] = []
            if position == 0:
                tuning_parameter = 1.0
            else:
                earlier_identifiers = batch_identifiers[:position]
                earlier_mask = np.isin(batches, earlier_identifiers)
                *earlier_arrays, _ = _split_labeled_unlabeled(risk_y_true[earlier_mask], risk_y_proxy[earlier_mask])
                tuning_parameter = _compute_clipped_tuning_parameter(
                    *earlier_arrays, power_tuning, max_tuning_parameter
                )
            batch_mask = batches == batch_identifiers[position]
            y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, _ = _split_labeled_unlabeled(
                risk_y_true[batch_mask], risk_y_proxy[batch_mask]
            )
            risk_batch_estimate = _compute_mean_estimate(y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, tuning_parameter)
            risk_batch_estimates_list.append(risk_batch_estimate)
            batch_n_true_list.append(len(y_true_labeled))
            batch_n_proxy_list.append(len(y_true_labeled) + len(y_proxy_unlabeled))
        risk_batch_estimates = np.array(risk_batch_estimates_list)
        batch_n_true = np.array(batch_n_true_list)
        batch_n_proxy = np.array(batch_n_proxy_list)

        normalized_estimates = _normalize_to_unit_interval(risk_batch_estimates, max_tuning_parameter)
        normalized_threshold = _normalize_to_unit_interval(risk_threshold, max_tuning_parameter)

        normalized_running_means, normalized_lower_bounds = _compute_empirical_bernstein_bounds(
            normalized_estimates, normalized_threshold, miscoverage
        )

        risk_running_means = _denormalize_from_unit_interval(normalized_running_means, max_tuning_parameter)
        risk_lower_bounds = _denormalize_from_unit_interval(normalized_lower_bounds, max_tuning_parameter)
        running_means, confidence_bounds, batch_estimates = self._postprocess(
            risk_running_means, risk_lower_bounds, risk_batch_estimates, higher_is_better, metric_lower_bound, metric_upper_bound
        )
        confidence_sequence = EmpiricalBernsteinConfidenceSequence(
            running_means=running_means,
        )
            metric_name=metric_name,
            alarm_threshold=threshold,
            batch_identifiers=batch_identifiers,
            batch_estimates=batch_estimates,
            batch_n_true=batch_n_true,
            batch_n_proxy=batch_n_proxy,
        )
        return result
```

`glide/monitors/__init__.py` gains `PPIMeanMonitor`; add it to `glide/__init__.py` exports as well. New unit tests cover only the PPI-specific pieces: `tests/unit/monitors/test_ppi.py` and `tests/unit/monitors/test_ppi_core.py`.

## Functional tests

**Per-batch estimates match PPIMeanEstimator without power tuning**

With `power_tuning=False` the tuning parameter is 1 everywhere, so each entry of `batch_estimates` must equal the `PPIMeanEstimator(power_tuning=False)` point estimate computed on that batch's data alone. The assertion must cover a batch *after* the first: the first batch uses the tuning parameter 1 unconditionally (it has no predecessor), so checking only `batch_estimates[0]` would pass even if `power_tuning=False` were ignored and power tuning applied throughout. Asserting every batch removes the question of which index to pick and exercises the `power_tuning=False` code path on batches that would otherwise be power-tuned.

```python
def test_detect_batch_estimates_match_ppi_estimator(y_true, y_proxy, batches):
    result = PPIMeanMonitor().detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5, power_tuning=False)
    estimator = PPIMeanEstimator()

    for position, identifier in enumerate(result.batch_identifiers):
        batch_mask = batches == identifier
        estimator_result = estimator.estimate(y_true[batch_mask], y_proxy[batch_mask], power_tuning=False)
        assert result.batch_estimates[position] == pytest.approx(estimator_result.mean)
```

**Prefix consistency (statelessness is sound)**

Calling `detect` on the data restricted to the first k batches must produce exactly the first k entries of the arrays returned by calling `detect` on the full dataset (every batch is monitored, so there is no reference batch to exclude). This is the property that makes repeated calls on a growing history jointly valid.

```python
def test_detect_prefix_consistency():
    # Four batches in row order (0, 1, 2, 3), each with two labeled then two unlabeled
    # rows; every batch is monitored.
    first_block = np.array([0.1, 0.2, np.nan, np.nan])
    later_block = np.array([0.3, 0.4, np.nan, np.nan])
    y_true = np.hstack([first_block, np.tile(later_block, 3)])
    y_proxy = np.tile(np.array([0.1, 0.2, 0.15, 0.25]), 4)
    # Integer identifiers in chronological row order, so a prefix is a plain mask.
    batches = np.repeat(np.arange(4), 4)

    monitor = PPIMeanMonitor()
    full = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
    # Restrict to the first three batches (identifiers 0, 1, 2).
    prefix_mask = batches <= 2
    prefix = monitor.detect(
        y_true[prefix_mask], y_proxy[prefix_mask], batches[prefix_mask], higher_is_better=False, threshold=0.5
    )

    # The prefix run reproduces the first three entries of the full run; this is
    # exactly what makes repeated calls on a growing history jointly valid.
    np.testing.assert_allclose(prefix.running_means, full.running_means[:3])
    np.testing.assert_allclose(prefix.confidence_bounds, full.confidence_bounds[:3])
```

**Risk/performance symmetry**

Monitoring a performance metric with `higher_is_better=True` must be the exact mirror image of monitoring the flipped data (`1 - y`) as a risk: same alarm flags, mirrored estimates, running means and bounds. The threshold flips with the data (`1 - threshold`).

```python
def test_detect_higher_is_better_symmetry(y_true, y_proxy, batches):
    risk = PPIMeanMonitor().detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.3)
    performance = PPIMeanMonitor().detect(1.0 - y_true, 1.0 - y_proxy, batches, higher_is_better=True, threshold=0.7)

    np.testing.assert_array_equal(performance.alarms, risk.alarms)
    np.testing.assert_allclose(performance.confidence_bounds, 1.0 - risk.confidence_bounds)
```

## Corner cases

- An empty `batches` array → `ValueError` (there is no batch to monitor).
- A batch with fewer than 2 labeled or fewer than 2 unlabeled samples → `ValueError`, same requirement as `PPIMeanEstimator` applied per batch.
- A `y_true` or `y_proxy` value outside `[metric_lower_bound, metric_upper_bound]` → `ValueError`; boundedness of the data is what makes the empirical-Bernstein bound valid.
- Constant proxy values across the prior batches and `power_tuning=True` → `ValueError` raised by `_compute_tuning_parameter` (zero variance). The first batch is exempt, since it uses the constant weight 1 and never calls the tuning estimator.
- A single batch (the first look of a growing history): must work and return a wide bound; with a metric in `[0, 1]` and default settings, the deviation budget exceeds the normalized range for the first few batches, so no alarm can mathematically fire there. This is expected finite-sample safety, not a bug.
- Interleaved batches (e.g. rows ordered `A, B, A, B`): `ValueError`; chronology is taken from row order, so each batch must be a contiguous block. Contiguous blocks in the wrong time order cannot be detected and are the user's responsibility (documented, not enforced).
- Free-form string identifiers (e.g. `"week 1"`, `"v2.3"`): fully supported, since identifier values are never compared; order comes from row position alone.
- `threshold` outside `[metric_lower_bound, metric_upper_bound]`, `metric_lower_bound >= metric_upper_bound`, `confidence_level` outside (0, 1), `max_tuning_parameter <= 0` → `ValueError` with the bad value in the message.
- Constant batch estimates equal to the threshold: the variance process is 0 except for the first term, bounds shrink at rate ~1/t, no alarm ever fires.

## Acceptance criteria

- [ ] `PPIMeanMonitor.detect` consumes raw arrays `(y_true, y_proxy, batches)` plus a required `threshold` and scalar options; the class holds no state between calls, exposes no `update`/`observe` methods, and the existing estimators, the inference result dataclasses and `glide/io` are all unchanged.
- [ ] `detect` computes no reference batch and takes no `source_miscoverage` or `tolerance`: every batch is monitored, the alarm fires when the anytime-valid bound crosses the user-supplied `threshold`, and the lone false-alarm knob is `confidence_level`.
- [ ] `PPIMeanMonitor` reuses the shared `_split_labeled_unlabeled` helper from `glide/estimators/core.py` (introduced by the prerequisite refactoring ticket) and adds no split logic of its own.
- [ ] The power-tuning parameter of each batch is computed exclusively from batches that strictly precede it, with the first batch (no predecessor) using the classic PPI weight of 1 (verified by a unit test asserting each `batch_estimates` entry equals an independently computed estimate whose tuning parameter is fitted on all prior batches only, that the first entry equals the `power_tuning=False` estimate, and that altering the last batch's data leaves every earlier `batch_estimates` entry unchanged).
- [ ] `PPIMeanMonitor.detect` returns a result whose `confidence_sequence` is an `EmpiricalBernsteinConfidenceSequence`, and `result.running_means`/`result.confidence_bounds`/`result.alarms` read through it (unit test).
- [ ] The PPIMeanEstimator-equivalence, prefix-consistency and risk/performance symmetry functional tests pass.
- [ ] Deterministic behavioral functional tests confirm no alarm fires on a stationary stream below the threshold and an alarm fires on a stream well above it; the many-seed false-alarm-rate and detection-power study (with and without power tuning) lives in the deep-dive notebook (`doc-pprm-deep-dive.md`), not in CI.
- [ ] `PPIMeanMonitor.detect` returns a `PredictionPoweredMeanMonitoringResult` (the dataclass itself is delivered and tested by `feat-mean-monitoring-results.md`).
- [ ] `PPIMeanMonitor` is importable from `glide.monitors` and `glide`.
- [ ] `_preprocess` raises `ValueError` on interleaved batches (a unit test passes rows ordered `A, B, A, B`) and accepts free-form string identifiers (a unit test labels batches `"week 1"`, `"week 2"`, ... in row order and detects normally).
- [ ] The `detect` docstring states the usage requirements the API cannot fully enforce: oldest-first row order (only interleaving is caught), non-overlapping batches, growing history across calls.

imerad:
### Description

- Adds `PPIMeanMonitor` to `glide.monitors`: an anytime-valid drift monitor over batched production data that combines a small set of true labels with a large set of proxy labels via a per-batch prediction-powered inference (PPI) estimate. It builds a running empirical-Bernstein confidence sequence and raises a drift alarm when the sequence crosses a user-supplied threshold, staying valid under repeated calls as new batches arrive.
- Adds `glide/monitors/ppi_core.py` with shared internals (tuning-parameter clipping, unit-interval normalization) used by the monitor.
- Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=176217941&issue=EmertonData%7Cglide%7C317) Closes #317

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/monitors/ppi.py
+++ b/glide/monitors/ppi.py
@@ -64,6 +64,8 @@ class PPIMeanMonitor:
     >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
     >>> result.drift_detected
     True
+    >>> result.first_alarm_index
+    23
     """
 
     def _preprocess(
@@ -243,10 +245,9 @@ class PPIMeanMonitor:
         confidence_level : float, optional
             How confident each alarm should be. At the default ``0.8`` the monitor
             raises "80% confident" alarms: under no drift it has at most a 20% chance
-            of *ever* raising a false alarm, no matter how often you look. Raising it
-            (say to ``0.95``) demands more evidence before alarming, so alarms become
-            more trustworthy but arrive later; lowering it detects sooner at the cost
-            of more false alarms.
+            of raising a false alarm. Raising this value demands more evidence before
+            alarming, so alarms become more trustworthy but arrive later; lowering it
+            detects sooner at the cost of more false alarms.
         power_tuning : bool, optional
             If ``True`` (default), compute the power-tuning parameter of each batch on
             all previous batches (the first batch, having no predecessor, uses 1). If
@@ -299,7 +300,7 @@ class PPIMeanMonitor:
 
         risk_batch_estimates = np.empty(n_batches)
         for position in range(n_batches):
-            if position == 0:
+            if position == 0 or (not power_tuning):
                 tuning_parameter = 1.0
             else:
                 earlier_mask = batch_codes < position
--- a/glide/monitors/ppi.py
+++ b/glide/monitors/ppi.py
@@ -1,0 +1,357 @@
+from typing import Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
+from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
+from glide.core.validation import (
+    _validate_bounds,
+    _validate_equal_lengths,
+    _validate_has_no_nan,
+    _validate_non_empty,
+    _validate_y_proxy,
+    _validate_y_true,
+)
+from glide.estimators.core import _split_labeled_unlabeled
+from glide.estimators.ppi_core import _compute_mean_estimate
+from glide.mean_monitoring_results import PredictionPoweredMeanMonitoringResult
+from glide.monitors.core import _scale_from_unit_risk, _scale_to_unit_risk, _unique_ordered_batches
+from glide.monitors.ppi_core import (
+    _compute_clipped_tuning_parameter,
+    _denormalize_from_unit_interval,
+    _normalize_to_unit_interval,
+)
+
+
+class PPIMeanMonitor:
+    """Anytime-valid drift monitor over a batched dataset of true and proxy labels.
+
+    It uses a per-batch Prediction-Powered Inference (PPI) estimate: a small set of
+    true labels and a large set of proxy labels, combined with a power-tuning
+    parameter fitted on the batches that strictly precede it (predictable, as
+    required for validity). To be used on accumulated non-overlapping production
+    batches by calling :meth:`detect` on the whole accumulated dataset at any time.
+    Data is passed oldest batch first, and every batch is monitored. The monitor
+    builds an anytime-valid confidence sequence on the running mean of the per-batch
+    PPI estimates, and raises a drift alarm when it crosses a user-supplied
+    ``threshold`` (the worst tolerable metric value). Because the bounds are valid at
+    all times simultaneously, :meth:`detect` may be called after every new batch
+    without inflating the false-alarm probability.
+
+    References
+    ----------
+    Zhang, Guangyi, Yunlong Cai, Guanding Yu, and Osvaldo Simeone. "Prediction-powered
+    risk monitoring of deployed models for detecting harmful distribution shifts."
+    arXiv preprint arXiv:2602.02229 (2026).
+
+    Howard, Steven R., Aaditya Ramdas, Jon McAuliffe, and Jasjeet Sekhon. "Time-uniform,
+    nonparametric, nonasymptotic confidence sequences." The Annals of Statistics 49,
+    no. 2 (2021): 1055-1080.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from glide.monitors import PPIMeanMonitor
+    >>> pre_drift_y_true = np.array([0.0, 0.2, np.nan, np.nan])
+    >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
+    >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
+    >>> post_drift_y_proxy = np.array([0.8, 1.0, 0.8, 1.0])
+    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
+    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
+    >>> batches = np.repeat(np.arange(51), 4)
+    >>> monitor = PPIMeanMonitor()
+    >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
+    >>> result.drift_detected
+    True
+    """
+
+    def _preprocess(
+        self,
+        y_true: NDArray,
+        y_proxy: NDArray,
+        batches: NDArray,
+        higher_is_better: bool,
+        threshold: float,
+        confidence_level: float,
+        max_tuning_parameter: float,
+        metric_lower_bound: float,
+        metric_upper_bound: float,
+    ) -> Tuple[NDArray, NDArray, float, NDArray, NDArray, NDArray]:
+        _validate_non_empty(y_true, "y_true")
+        _validate_equal_lengths(y_true, y_proxy, batches, names=["y_true", "y_proxy", "batches"])
+        _validate_has_no_nan(batches, "batches")
+        _validate_bounds(
+            confidence_level, "confidence_level", lower=0, upper=1, left_inclusive=False, right_inclusive=False
+        )
+        _validate_bounds(max_tuning_parameter, "max_tuning_parameter", lower=0, left_inclusive=False)
+        _validate_bounds(
+            metric_lower_bound,
+            "metric_lower_bound",
+            upper=metric_upper_bound,
+            right_inclusive=False,
+            error_message=(
+                f"'metric_lower_bound' must be strictly smaller than 'metric_upper_bound'; "
+                f"got {metric_lower_bound!r} and {metric_upper_bound!r}."
+            ),
+        )
+        _validate_bounds(
+            threshold,
+            "threshold",
+            lower=metric_lower_bound,
+            upper=metric_upper_bound,
+            error_message=(
+                f"'threshold' must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
+                f"'metric_upper_bound'={metric_upper_bound!r}."
+            ),
+        )
+        _validate_y_proxy(y_proxy)
+        _validate_y_true(y_true)
+        labeled_mask = ~np.isnan(y_true)
+        labeled_values = y_true[labeled_mask]
+        _validate_bounds(
+            labeled_values,
+            "y_true",
+            lower=metric_lower_bound,
+            upper=metric_upper_bound,
+            error_message=(
+                f"'y_true' values must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
+                f"'metric_upper_bound'={metric_upper_bound!r}; got values in "
+                f"[{labeled_values.min()!r}, {labeled_values.max()!r}]."
+            ),
+        )
+        _validate_bounds(
+            y_proxy,
+            "y_proxy",
+            lower=metric_lower_bound,
+            upper=metric_upper_bound,
+            error_message=(
+                f"'y_proxy' values must lie between 'metric_lower_bound'={metric_lower_bound!r} and "
+                f"'metric_upper_bound'={metric_upper_bound!r}; got values in "
+                f"[{y_proxy.min()!r}, {y_proxy.max()!r}]."
+            ),
+        )
+        batch_identifiers, batch_codes = _unique_ordered_batches(batches)
+        n_batches = len(batch_identifiers)
+        batch_n_true = np.bincount(batch_codes[labeled_mask], minlength=n_batches)
+        batch_n_proxy = np.bincount(batch_codes, minlength=n_batches)
+        batch_n_unlabeled = batch_n_proxy - batch_n_true
+
+        worst_labeled_position = np.argmin(batch_n_true)
+        _validate_bounds(
+            batch_n_true[worst_labeled_position],
+            "y_true",
+            lower=2,
+            error_message=(
+                f"'y_true' must have at least 2 labeled values per batch; got "
+                f"{batch_n_true[worst_labeled_position]} in batch '{batch_identifiers[worst_labeled_position]}'."
+            ),
+        )
+        worst_unlabeled_position = np.argmin(batch_n_unlabeled)
+        _validate_bounds(
+            batch_n_unlabeled[worst_unlabeled_position],
+            "y_true",
+            lower=2,
+            error_message=(
+                f"'y_true' must have at least 2 unlabeled values per batch; got "
+                f"{batch_n_unlabeled[worst_unlabeled_position]} in batch "
+                f"'{batch_identifiers[worst_unlabeled_position]}'."
+            ),
+        )
+
+        risk_y_true = _scale_to_unit_risk(y_true, metric_lower_bound, metric_upper_bound, higher_is_better)
+        risk_y_proxy = _scale_to_unit_risk(y_proxy, metric_lower_bound, metric_upper_bound, higher_is_better)
+        risk_threshold = _scale_to_unit_risk(threshold, metric_lower_bound, metric_upper_bound, higher_is_better)
+        return risk_y_true, risk_y_proxy, risk_threshold, batch_codes, batch_n_true, batch_n_proxy
+
+    def _postprocess(
+        self,
+        risk_running_means: NDArray,
+        risk_confidence_bounds: NDArray,
+        risk_batch_mean_estimates: NDArray,
+        higher_is_better: bool,
+        metric_lower_bound: float,
+        metric_upper_bound: float,
+    ) -> Tuple[NDArray, NDArray, NDArray]:
+        running_means = _scale_from_unit_risk(
+            risk_running_means, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        confidence_bounds = _scale_from_unit_risk(
+            risk_confidence_bounds, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        batch_mean_estimates = _scale_from_unit_risk(
+            risk_batch_mean_estimates, metric_lower_bound, metric_upper_bound, higher_is_better
+        )
+        return running_means, confidence_bounds, batch_mean_estimates
+
+    def detect(
+        self,
+        y_true: NDArray,
+        y_proxy: NDArray,
+        batches: NDArray,
+        higher_is_better: bool,
+        threshold: float,
+        metric_name: str = "Metric",
+        confidence_level: float = 0.8,
+        power_tuning: bool = True,
+        metric_lower_bound: float = 0.0,
+        metric_upper_bound: float = 1.0,
+        max_tuning_parameter: float = 1.0,
+    ) -> PredictionPoweredMeanMonitoringResult:
+        """Detect a drift of the running mean across a batched dataset.
+
+        Splits the data by batch, computes a prediction-powered estimate per batch
+        with a power-tuning parameter fitted only on batches that strictly precede
+        it (the first batch, having no predecessor, uses the classic PPI weight of
+        1), and builds an anytime-valid empirical-Bernstein confidence sequence on
+        the running mean of those estimates. An alarm is raised at every batch where
+        the sequence crosses the user-supplied ``threshold``.
+
+        Rows must be ordered oldest batch first and grouped into contiguous blocks;
+        identifier values are not compared, so any label type works. Batches must be
+        non-overlapping (no shared samples), and successive calls must be made on
+        growing histories of the same data; passing the full accumulated dataset at
+        every call makes the anytime-valid guarantee hold jointly over all calls.
+        Alternatively the data may be restricted to the most recent batches, in which
+        case the guarantee holds within each restriction but not across the moving
+        history as a whole.
+
+        Parameters
+        ----------
+        y_true : NDArray
+            Array of labeled observations, shape ``(n_samples,)``.
+            Labeled entries are finite; unlabeled entries are ``np.nan``.
+        y_proxy : NDArray
+            Array of proxy predictions, shape ``(n_samples,)``.
+            Must be fully populated (no NaN).
+        batches : NDArray
+            Array of batch identifiers, shape ``(n_samples,)``. Rows must be ordered
+            oldest batch first and grouped into contiguous blocks. Identifier values
+            are not compared, so any hashable label type works (integers, dates,
+            free-form strings).
+        higher_is_better : bool
+            ``False`` when the metric is a risk (drift means the metric increased),
+            ``True`` when it is a performance (drift means the metric decreased).
+        threshold : float
+            The metric value the running mean is monitored against, in metric units:
+            the worst level the user is willing to tolerate. An alarm fires once the
+            anytime-valid bound proves the running metric has crossed it (the running
+            risk exceeds it for a risk, the running performance falls below it for a
+            performance). Must lie within ``[metric_lower_bound, metric_upper_bound]``.
+        metric_name : str, optional
+            Human-readable label for the metric. Defaults to ``"Metric"``.
+        confidence_level : float, optional
+            How confident each alarm should be. At the default ``0.8`` the monitor
+            raises "80% confident" alarms: under no drift it has at most a 20% chance
+            of *ever* raising a false alarm, no matter how often you look. Raising it
+            (say to ``0.95``) demands more evidence before alarming, so alarms become
+            more trustworthy but arrive later; lowering it detects sooner at the cost
+            of more false alarms.
+        power_tuning : bool, optional
+            If ``True`` (default), compute the power-tuning parameter of each batch on
+            all previous batches (the first batch, having no predecessor, uses 1). If
+            ``False``, use 1 everywhere (classic PPI).
+        metric_lower_bound : float, optional
+            Known lower bound of the metric. Defaults to ``0.0``.
+        metric_upper_bound : float, optional
+            Known upper bound of the metric. Defaults to ``1.0``.
+        max_tuning_parameter : float, optional
+            Upper clip for the power-tuning parameter. Larger values allow more
+            reliance on the proxy but widen the confidence sequence through the
+            affine normalization. Defaults to ``1.0``.
+
+        Returns
+        -------
+        PredictionPoweredMeanMonitoringResult
+            Per-batch estimates, running means, anytime-valid confidence bounds,
+            alarm flags, and the alarm threshold, all in the original metric
+            orientation.
+
+        Raises
+        ------
+        ValueError
+            - If ``y_true`` is empty.
+            - If ``y_true``, ``y_proxy`` and ``batches`` have different lengths.
+            - If ``batches`` contains NaN values (numeric dtype) or None values (non-numeric dtype).
+            - If ``confidence_level`` is not in (0, 1).
+            - If ``max_tuning_parameter`` is not strictly positive.
+            - If ``metric_lower_bound >= metric_upper_bound``.
+            - If ``threshold`` falls outside ``[metric_lower_bound, metric_upper_bound]``.
+            - If any proxy value is NaN or all proxy values are identical.
+            - If labeled ``y_true`` values are constant.
+            - If any value falls outside ``[metric_lower_bound, metric_upper_bound]``.
+            - If batches are interleaved rather than grouped into contiguous blocks.
+            - If any batch has fewer than 2 labeled or fewer than 2 unlabeled samples.
+            - If proxy values are constant across the prior batches (with ``power_tuning=True``).
+        """
+        risk_y_true, risk_y_proxy, risk_threshold, batch_codes, batch_n_true, batch_n_proxy = self._preprocess(
+            y_true,
+            y_proxy,
+            batches,
+            higher_is_better,
+            threshold,
+            confidence_level,
+            max_tuning_parameter,
+            metric_lower_bound,
+            metric_upper_bound,
+        )
+        n_batches = len(batch_n_true)
+
+        risk_batch_estimates = np.empty(n_batches)
+        for position in range(n_batches):
+            if position == 0:
+                tuning_parameter = 1.0
+            else:
+                earlier_mask = batch_codes < position
+                y_true_earlier, y_proxy_labeled_earlier, y_proxy_unlabeled_earlier, _ = _split_labeled_unlabeled(
+                    risk_y_true[earlier_mask], risk_y_proxy[earlier_mask]
+                )
+                tuning_parameter = _compute_clipped_tuning_parameter(
+                    y_true_earlier,
+                    y_proxy_labeled_earlier,
+                    y_proxy_unlabeled_earlier,
+                    power_tuning,
+                    max_tuning_parameter,
+                )
+            batch_mask = batch_codes == position
+            y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, _ = _split_labeled_unlabeled(
+                risk_y_true[batch_mask], risk_y_proxy[batch_mask]
+            )
+            risk_batch_estimates[position] = _compute_mean_estimate(
+                y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, tuning_parameter
+            )
+
+        normalized_estimates = _normalize_to_unit_interval(risk_batch_estimates, max_tuning_parameter)
+        normalized_threshold = _normalize_to_unit_interval(risk_threshold, max_tuning_parameter)
+
+        miscoverage = 1.0 - confidence_level
+        normalized_running_means, normalized_lower_bounds = _compute_empirical_bernstein_bounds(
+            normalized_estimates, normalized_threshold, miscoverage
+        )
+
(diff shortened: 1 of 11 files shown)


imerad:
## Code review findings

Automated review of this branch's diff against `main`. Ranked most-severe first.

### 🔴 Correctness

**1. Unclipped first-batch tuning parameter breaks the normalization envelope — [`glide/monitors/ppi.py:303`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi.py#L303)**

The power-tuning parameter for the first batch (and every batch when `power_tuning=False`) is hardcoded to `1.0` and never clipped to `max_tuning_parameter`, but `_normalize_to_unit_interval`/`_denormalize_from_unit_interval` in `ppi_core.py` assume every batch estimate was produced with λ ∈ `[0, max_tuning_parameter]`.

`max_tuning_parameter` is only validated to be `> 0` (no coupling to `1.0`), so any call with `max_tuning_parameter < 1.0` lets the first batch's raw estimate fall outside the assumed range. Reproduced numerically: with `y_true ≈ [0.95, 0.97]`, `y_proxy_labeled ≈ [0.02, 0.03]`, `y_proxy_unlabeled ≈ [0.95, 0.97]`, the raw estimate is `1.895`; normalizing with `max_tuning_parameter=0.5` (envelope `[-0.5, 1.5]`) gives `1.1975`, outside `[0, 1]`.

This silently violates the boundedness assumption the empirical-Bernstein betting bound depends on for its anytime-valid guarantee — no exception is raised, the guarantee is just quietly broken. No test in this PR exercises `max_tuning_parameter != 1.0`.

**2. Undocumented negative-λ floor — [`glide/monitors/ppi_core.py:17`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi_core.py#L17)**

`_compute_clipped_tuning_parameter` silently floors a negative tuning parameter to `0.0`, but this floor is undocumented in the public `detect()` docstring, which only describes `max_tuning_parameter` as an "upper clip." When the proxy is negatively correlated with the true label, `_compute_tuning_parameter` can legitimately return a negative λ (as it does, unclipped, for `PPIMeanEstimator`). The monitor's undocumented lower floor makes its batch estimates diverge from what `PPIMeanEstimator` would produce on the same data whenever `power_tuning=True` — untested, since the only cross-check test in `tests/functional/monitors/test_ppi.py` uses `power_tuning=False`.

### 🟡 Efficiency

**3. O(n_batches²) tuning-parameter recomputation — [`glide/monitors/ppi.py:302`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi.py#L302)**

The per-batch loop recomputes covariance/variance over the entire growing prefix from scratch at every iteration (`earlier_mask = batch_codes < position`, then `np.cov`/`np.var` over all prior samples), making `detect()` quadratic in the number of batches. Since the monitor is documented to be called after every new batch on a growing production history, this becomes materially slow as history grows — unlike `ClassicalMeanMonitor.detect`, which computes all batch means with a single vectorized `np.bincount` call.

### 🟠 Design

**4. Confidence-sequence machinery hardcodes `[0,1]` — [`glide/confidence_sequences/empirical_bernstein.py:36`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/confidence_sequences/empirical_bernstein.py#L36)**

`_compute_empirical_bernstein_bounds` hardcodes an assumed `[0,1]` input range instead of accepting explicit bounds, pushing every caller to invent its own affine rescale. `ClassicalMeanMonitor` already relies on its metric-bound rescale landing batch means in `[0,1]`; `PPIMeanMonitor` now needs a *second*, independent normalize/denormalize pair (`ppi_core.py`) to force its wider-range estimate back into `[0,1]`. A future monitor type will have to hand-roll a third copy of this scheme instead of passing its known bounds into one general mechanism — and, as finding #1 shows, an incomplete per-caller rescale can silently break the shared function's core assumption with no validation catching it.

### 🔵 Reuse / duplication

**5. `_postprocess` is a verbatim duplicate — [`glide/monitors/ppi.py:169`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi.py#L169)**

`PPIMeanMonitor._postprocess` is byte-for-byte identical to `ClassicalMeanMonitor._postprocess` (`glide/monitors/classical.py:122-140`), duplicated instead of factored into a shared helper in `glide/monitors/core.py`. Against the repo's "No Redundancy" rule ("If two classes share logic ... extract it rather than duplicating").

**6. Validation blocks copied near-verbatim from `ClassicalMeanMonitor` — [`glide/monitors/ppi.py:114`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi.py#L114)**

The bounds/threshold validation block and the y-value-range and minimum-batch-count checks in `_preprocess` are copied near-verbatim from `ClassicalMeanMonitor._preprocess` (`glide/monitors/classical.py:69-117`), including identical error-message templates. The two "at least N per batch" blocks (labeled and unlabeled, lines 142-162) are also copy-pasted within the same file. A fix applied to one copy won't propagate to the others.

### ⚪ Conventions (CLAUDE.md)

**7. Threshold error message omits the bad value — [`glide/monitors/ppi.py:100`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/glide/monitors/ppi.py#L100)**

The `threshold` validation's `error_message` omits the actual `threshold` value that failed, violating "Always include the bad value in the message." Calling `detect(..., threshold=5.0, metric_lower_bound=0.0, metric_upper_bound=1.0)` raises a message that never shows `5.0`, unlike the neighboring `metric_lower_bound`/`y_true`/`y_proxy` checks in the same method.

**8. `np.concat` instead of `np.hstack` — [`tests/functional/monitors/test_ppi.py:36`](https://github.com/EmertonData/glide/blob/feat/ppi_mean_monitor/tests/functional/monitors/test_ppi.py#L36)**

`xi = np.concat(xis)` joins a list of same-shaped 1-D arrays — the canonical `np.hstack` use case per "Use `np.hstack` or `np.vstack` instead of `np.concatenate` whenever possible." This PR's own doctest in `glide/monitors/ppi.py` uses `np.hstack` for the identical pattern.

---
*Generated with [Claude Code](https://claude.com/claude-code)*

glide/monitors/ppi.py
```diff
@@ -0,0 +1,361 @@
...
+        the sequence crosses the user-supplied ``threshold``.
+
+        Rows must be ordered oldest batch first and grouped into contiguous blocks;
+        identifier values are not compared, so any label type works. Batches must be
+        non-overlapping (no shared samples), and successive calls must be made on
+        growing histories of the same data; passing the full accumulated dataset at
+        every call makes the anytime-valid guarantee hold jointly over all calls.
+        Alternatively the data may be restricted to the most recent batches, in which
+        case the guarantee holds within each restriction but not across the moving
+        history as a whole.
```
gmartinon-ed, line 215: Un peu trop redondant avec la docstring de classe à mon goût
Je pense que la docstring de classe est un peu trop détaillée

gmartinon-ed (changes requested):
Also : add the papers in the README (at least Zhang et al, but also Podkaev, and maybe howard and waudby-smith)

glide/monitors/ppi.py
imerad: Traité, j'ai élagué un peu des deux côtés.

gmartinon-ed (approved):

PR #347, opened 2026-07-17
Add AsymptoticConfidenceSequence

### LINKED ISSUE #348: AsymptoticConfidenceSequence: a variance-adaptive anytime-valid confidence sequence
## TODO

- [ ] Bien checker les sections pertinentes de l'article en termes d'exactitude de l'implémentation, et pertinence des hypothèses
- [ ] Challenger l'existence de ces classes AsymptoticConfidenceSequence et EmpiricalConfidenceSequence: vraiment nécessaires ? Pourquoi pas garde une simple classe unique ConfidenceSequence avec un attribut qui identifie la méthode de calcul ?

## Background

A confidence sequence is a sequence of confidence bounds, one per batch, that are valid at all batches simultaneously, so a monitor can check them after every new batch without inflating the false-alarm probability. The existing `EmpiricalBernsteinConfidenceSequence` treats each per-batch estimate as an arbitrary point in [0, 1]: its bound must guard against the worst case where the next point jumps anywhere in that range, so its width has a floor of roughly 2.66/t (at 80% confidence) after t batches, no matter how precise each batch estimate actually is. In practice each batch estimate comes with a known standard error (an error bar), often around 0.01 to 0.1, so the empirical-Bernstein bound is 5 to 15 times wider than the error bars justify, which makes drift detection very slow.

This ticket adds a second confidence sequence that consumes those error bars directly, following the asymptotic confidence sequences of Waudby-Smith et al. (Proposition B.2 of the reference below). Each batch estimate is treated as approximately Gaussian with its own standard error σ̂ᵢ, the accumulated variance Vₜ = Σᵢ σ̂ᵢ² plays the role of time, and the running mean μ̂ₜ of the batch estimates gets the anytime-valid lower bound: in math notation this is written as μ̂ₜ − sqrt( 2(Vₜρ² + 1)/(t²ρ²) · log(1 + sqrt(Vₜρ² + 1)/(2α)) ), where α is the miscoverage and ρ > 0 is a tuning scale choosing the batch at which the bound is tightest. The width now scales with the actual error bars instead of a worst-case constant. The trade-off is that the guarantee is asymptotic (it relies on batch estimates being approximately Gaussian) rather than exact, which is the same regime as the CLT-based confidence intervals used everywhere else in GLIDE.

## Design choices

The new code lives in `glide/confidence_sequences/asymptotic.py`, mirroring `empirical_bernstein.py` exactly: a private module-level computation function plus a public result dataclass. Both dataclasses would share identical fields and an identical `test_null_hypothesis` method, so the existing `ConfidenceSequence` protocol in `glide/confidence_sequences/base.py` is converted into a concrete base dataclass carrying that shared implementation, and both sequences inherit from it. Keeping the `ConfidenceSequence` name means the annotation in `glide/mean_monitoring_results/base.py` and the public export stay untouched. This deliberately diverges from the `confidence_intervals` package, where the protocol stays a protocol because `BootstrapConfidenceInterval` and `CLTConfidenceInterval` implement genuinely different `test_null_hypothesis` bodies; here the implementations would be copies, and a protocol plus a concrete base declaring the same shape would be redundant. The two subclasses are deliberately docstring-only and must not be collapsed into the base: the class is how a result object records which statistical guarantee (exact nonparametric versus asymptotic) backs its bounds, each class carries its own assumptions and `References` rendered as a separate API reference entry, and each is the landing spot for method-specific fields later (variance process, tuning scale). This ticket touches only `glide/confidence_sequences/` and its tests (plus the one import site above if the type checker needs it); wiring the sequence into a monitor is a separate ticket (`feat-asymptotic-ppi-mean-monitor.md`).

The tuning scale ρ exists because no anytime-valid bound can be tightest at all times simultaneously (a law-of-the-iterated-logarithm fact): every confidence sequence must decide where along the time axis to spend its tightness, and ρ is that dial. The empirical-Bernstein sequence makes the same decision implicitly (its uniform mixture over (0, 1) is the parameter-free but looser choice documented in the user guide); Robbins' Gaussian mixture just makes it explicit. Crucially, ρ affects tightness only, never validity: the false-alarm guarantee holds for any ρ > 0.

ρ is not exposed directly: callers pass `tightest_at_batch` (the batch index, counted from 1, where the bound should be tightest) and ρ is computed from the closed-form approximation of the reference paper (equation 50), with the miscoverage doubled because our bound is one-sided (footnote 11 of the paper). That formula requires the doubled miscoverage to stay below 1, hence the validation `miscoverage < 0.5`. The choice is deliberately forgiving, so users should leave the monitor-level default of 10 alone: the mis-tuning penalty is logarithmic and scale-free (independent of the standard errors). With `tightest_at_batch=10`, the bound is within 8% of the perfectly tuned width everywhere from batch 3 to batch 100, and only 20% wider at batch 500, against the 5 to 15 times gained over the empirical-Bernstein bound (numbers verified with a prototype at 80% confidence).

### ConfidenceSequence (protocol turned concrete base)

**Location:** `glide/confidence_sequences/base.py`

**Purpose:** Concrete base dataclass holding the per-look arrays and the alarm test shared by all confidence sequences.

Replace the `Protocol` with a `@dataclass` of the same name, moving the fields, the `test_null_hypothesis` method, and its docstring verbatim from `EmpiricalBernsteinConfidenceSequence` (drop the now-unused `Protocol` import). The class docstring describes the shared contract (per-look running means, harmful-side bounds, time-uniform validity) and, being concrete public API now, carries a runnable doctest:

```python
@dataclass
class ConfidenceSequence:
    """Anytime-valid confidence sequence on a running mean of per-batch estimates.

    Holds the per-look running means and the one-sided anytime-valid bound on the
    side where drift is harmful. The bounds hold simultaneously at all looks, so
    testing after every batch does not inflate the false-alarm probability.

    Parameters
    ----------
    running_mean_estimates : NDArray
        Per-look running mean of the per-batch estimates, in original metric units.
    confidence_bounds : NDArray
        Per-look harmful-side anytime-valid bound, in original metric units.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.confidence_sequences import ConfidenceSequence
    >>> sequence = ConfidenceSequence(
    ...     running_mean_estimates=np.array([0.4, 0.6]),
    ...     confidence_bounds=np.array([0.1, 0.55]),
    ... )
    >>> sequence.test_null_hypothesis(0.5, alternative="larger")
    array([False,  True])
    """

    running_mean_estimates: NDArray
    confidence_bounds: NDArray

    def test_null_hypothesis(
        self,
        h0_value: float,
        alternative: Literal["larger", "smaller"] = "larger",
    ) -> NDArray:
        # Body and docstring moved unchanged from EmpiricalBernsteinConfidenceSequence.
        ...
```

`EmpiricalBernsteinConfidenceSequence` then reduces to a docstring-only subclass (keep its current class docstring, references and doctest; delete the duplicated fields and method):

```python
class EmpiricalBernsteinConfidenceSequence(ConfidenceSequence):
    """Anytime-valid empirical-Bernstein confidence sequence on a running mean.

    [current docstring, References and Examples kept as they are]
    """
```

### _compute_asymptotic_bounds and AsymptoticConfidenceSequence

**Location:** `glide/confidence_sequences/asymptotic.py`

**Purpose:** Compute the one-sided Gaussian-mixture bounds from per-batch estimates and their standard errors, and expose the result as a confidence sequence.

```python
from typing import Tuple

import numpy as np

def _compute_asymptotic_bounds(
    batch_estimates: NDArray,
    batch_std_estimates: NDArray,
    miscoverage: float,
    tightest_at_batch: int,
) -> Tuple[NDArray, NDArray]:
    # The one-sided boundary uses the two-sided formula at twice the miscoverage,
    # which is only defined below 1, hence the 0.5 upper bound.
    _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=0.5, left_inclusive=False, right_inclusive=False)
    _validate_bounds(tightest_at_batch, "tightest_at_batch", lower=1)
    n_batches = len(batch_estimates)
    batch_counts = np.arange(1, n_batches + 1)
    running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
    # Intrinsic time of the Gaussian approximation: the accumulated squared
    # standard errors of the batch estimates, not their realized scatter.
    variance_process = np.cumsum(batch_std_estimates**2)
    target_position = min(tightest_at_batch, n_batches) - 1
    target_intrinsic_time = variance_process[target_position]
    _validate_bounds(
        target_intrinsic_time,
        "batch_std_estimates",
        lower=0.0,
        left_inclusive=False,
        error_message=(
            f"'batch_std_estimates' must accumulate a positive variance by batch {target_position + 1}; "
            f"got {target_intrinsic_time!r}."
        ),
    )
    doubled_miscoverage = 2.0 * miscoverage
    log_term = -2.0 * np.log(doubled_miscoverage)
    tuning_scale_squared = (log_term + np.log(log_term + 1.0)) / target_intrinsic_time
    scaled_times = variance_process * tuning_scale_squared + 1.0
    boundary_widths = np.sqrt(
        2.0
        * scaled_times
        / (batch_counts**2 * tuning_scale_squared)
        * np.log(1.0 + np.sqrt(scaled_times) / doubled_miscoverage)
    )
    lower_bounds = running_mean_estimates - boundary_widths
    return running_mean_estimates, lower_bounds


class AsymptoticConfidenceSequence(ConfidenceSequence):
    """Anytime-valid asymptotic confidence sequence on a running mean.

    A Gaussian-mixture confidence sequence whose width scales with the known
    standard errors of the per-batch estimates instead of a worst-case bounded
    range, so the bounds stay tight when per-batch estimates are precise. The
    price is an asymptotic (rather than exact) time-uniform guarantee: per-batch
    estimates must be approximately Gaussian with consistently estimated
    variances.

    Parameters
    ----------
    running_mean_estimates : NDArray
        Per-look running mean of the per-batch estimates, in original metric units.
    confidence_bounds : NDArray
        Per-look harmful-side anytime-valid bound, in original metric units.

    References
    ----------

    Robbins, Herbert. "Statistical methods related to the law of the iterated
    logarithm." The Annals of Mathematical Statistics 41, no. 5 (1970): 1397-1409.

    Examples
    --------
    >>> import numpy as np
    >>> from glide.confidence_sequences import AsymptoticConfidenceSequence
    >>> sequence = AsymptoticConfidenceSequence(
    ...     running_mean_estimates=np.array([0.4, 0.6]),
    ...     confidence_bounds=np.array([0.1, 0.55]),
    ... )
    >>> sequence.test_null_hypothesis(0.5, alternative="larger")
    array([False,  True])
    """
```

Export `AsymptoticConfidenceSequence` from `glide/confidence_sequences/__init__.py` (`ConfidenceSequence` is already exported and keeps its name) and add it to `docs/api/confidence_sequences.md` with a `:::` mkdocstrings entry next to the existing one.

**Tests.** `tests/unit/confidence_sequences/test_asymptotic.py` mirrors `test_empirical_bernstein.py` section by section (`# --- _compute_asymptotic_bounds ---` with `batch_estimates`/`batch_std_estimates` fixtures, then `# --- AsymptoticConfidenceSequence ---` with a `sequence` fixture and a `test_sequence_attributes` test). The `test_null_hypothesis_*` tests and the `sequence` fixture currently living in `test_empirical_bernstein.py` move to a new `tests/unit/confidence_sequences/test_base.py` targeting `ConfidenceSequence` directly (now instantiable). As in `test_empirical_bernstein.py`, all statistical properties are unit tests with hardcoded expected values (all verified against a prototype run; in the anchor test the standard errors deliberately differ so the variance-process accumulation is exercised, and its first look doubles as the single-batch case):

```python
@pytest.fixture
def batch_estimates():
    return np.array([0.4, 0.6])


@pytest.fixture
def batch_std_estimates():
    return np.array([0.1, 0.2])


def test_compute_asymptotic_bounds_delegates_to_validation(batch_estimates, batch_std_estimates):
    with patch.object(asymptotic_module, "_validate_bounds") as mock_validate_bounds:
        _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=1)

        assert mock_validate_bounds.call_count == 3
        assert mock_validate_bounds.call_args_list[0][0] == (0.2, "miscoverage")
        assert mock_validate_bounds.call_args_list[0][1] == {
            "lower": 0.0,
            "upper": 0.5,
            "left_inclusive": False,
            "right_inclusive": False,
        }
        assert mock_validate_bounds.call_args_list[1][0] == (1, "tightest_at_batch")
        assert mock_validate_bounds.call_args_list[1][1] == {"lower": 1}
        assert mock_validate_bounds.call_args_list[2][0][1] == "batch_std_estimates"
        assert "must accumulate a positive variance" in mock_validate_bounds.call_args_list[2][1]["error_message"]


def test_compute_asymptotic_bounds(batch_estimates, batch_std_estimates):
    running_means, lower_bounds = _compute_asymptotic_bounds(
        batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=1
    )
    np.testing.assert_allclose(running_means, np.array([0.4, 0.5]))
    np.testing.assert_allclose(lower_bounds, np.array([0.181036, 0.247749]), atol=1e-6)


def test_compute_asymptotic_bounds_width_shrinks_with_batches():
    running_means, lower_bounds = _compute_asymptotic_bounds(
        np.array([0.5, 0.5, 0.5]), np.array([0.05, 0.05, 0.05]), miscoverage=0.2, tightest_at_batch=1
    )
    np.testing.assert_allclose(running_means - lower_bounds, np.array([0.109482, 0.076885, 0.063525]), atol=1e-6)


def test_compute_asymptotic_bounds_widens_with_confidence(batch_estimates, batch_std_estimates):
    _, loose = _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.3, tightest_at_batch=1)
    _, tight = _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.1, tightest_at_batch=1)
    np.testing.assert_allclose(loose, np.array([0.195628, 0.274674]), atol=1e-6)
    np.testing.assert_allclose(tight, np.array([0.150778, 0.209157]), atol=1e-6)


def test_compute_asymptotic_bounds_target_clamped_to_last_batch(batch_estimates, batch_std_estimates):
    _, clamped = _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=5)
    _, exact = _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=2)
    np.testing.assert_allclose(clamped, exact)
```

The cross-family tightness property (asymptotic bounds beating empirical-Bernstein ones on precise batch estimates) is deliberately not tested here: it is covered end to end by the earlier-alarm functional tests of the monitor tickets, and there is no `tests/functional/confidence_sequences/` precedent to introduce.

## Corner cases

- `miscoverage >= 0.5` (i.e. confidence level of 50% or below): raise `ValueError` with the bad value, since the one-sided tuning formula is undefined there.
- `tightest_at_batch < 1`: raise `ValueError`.
- `tightest_at_batch` larger than the number of batches: no error; the target position clamps to the last available batch.
- All standard errors equal to zero up to the target batch: raise `ValueError` (positive accumulated variance validation above).

## Acceptance criteria

- [ ] `AsymptoticConfidenceSequence` is importable from `glide.confidence_sequences` and rendered in `docs/api/confidence_sequences.md`.
- [ ] `glide/confidence_sequences/base.py` no longer defines a `Protocol`: `ConfidenceSequence` is a concrete dataclass, both sequence classes inherit from it, and the `confidence_sequence` annotation in `glide/mean_monitoring_results/base.py` still type-checks unchanged.
- [ ] `_compute_asymptotic_bounds(np.array([0.4, 0.6]), np.array([0.1, 0.2]), 0.2, 1)` returns lower bounds `[0.181036, 0.247749]` (±1e-6).
- [ ] `EmpiricalBernsteinConfidenceSequence` inherits from `ConfidenceSequence` and its existing unit and doctest behavior is unchanged (tests relocated, not rewritten).
- [ ] `tests/unit/confidence_sequences/test_asymptotic.py` mirrors the section layout of `test_empirical_bernstein.py` and the five unit tests above pass with their hardcoded values.

imerad:
### Description

- Adds `AsymptoticConfidenceSequence`, a variance-adaptive anytime-valid confidence sequence that consumes known per-batch standard errors instead of treating each batch estimate as an arbitrary point in `[0, 1]`, tightening drift-detection bounds when batch estimates are precise. `ConfidenceSequence` is converted from a structural protocol into a concrete dataclass base class shared by both `EmpiricalBernsteinConfidenceSequence` and the new `AsymptoticConfidenceSequence`.
- Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=214000263) and Closes #348 

### Noteworthy choices 

Factorizing AsymptoticConfidenceSequence and EmpiricalConfidenceSequence into a single class would have implied a significant refactor and removed the opportunity to give details in these two classes' docstrings. It has been set aside and can be considered in the future.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/confidence_sequences/asymptotic.py
+++ b/glide/confidence_sequences/asymptotic.py
@@ -1,10 +1,11 @@
+from dataclasses import dataclass
 from typing import Tuple
 
 import numpy as np
 from numpy.typing import NDArray
 
 from glide.confidence_sequences.base import ConfidenceSequence
-from glide.core.validation import _validate_bounds
+from glide.core.validation import _validate_bounds, _validate_equal_lengths, _validate_is_integer, _validate_non_empty
 
 
 def _compute_asymptotic_bounds(
@@ -13,9 +14,18 @@ def _compute_asymptotic_bounds(
     miscoverage: float,
     tightest_at_batch: int,
 ) -> Tuple[NDArray, NDArray]:
+    _validate_non_empty(batch_estimates, "batch_estimates")
+    _validate_equal_lengths(batch_estimates, batch_std_estimates, names=["batch_estimates", "batch_std_estimates"])
+    _validate_bounds(
+        batch_std_estimates,
+        "batch_std_estimates",
+        lower=0.0,
+        error_message=f"'batch_std_estimates' must be non-negative; got {batch_std_estimates.min()!r}.",
+    )
     # The one-sided boundary uses the two-sided formula at twice the miscoverage,
     # which is only defined below 1, hence the 0.5 upper bound.
     _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=0.5, left_inclusive=False, right_inclusive=False)
+    _validate_is_integer(tightest_at_batch, "tightest_at_batch")
     _validate_bounds(tightest_at_batch, "tightest_at_batch", lower=1)
     n_batches = len(batch_estimates)
     batch_counts = np.arange(1, n_batches + 1)
@@ -49,6 +59,7 @@ def _compute_asymptotic_bounds(
     return running_mean_estimates, lower_bounds
 
 
+@dataclass
 class AsymptoticConfidenceSequence(ConfidenceSequence):
     """Anytime-valid asymptotic confidence sequence on a running mean.
 
--- a/glide/confidence_sequences/asymptotic.py
+++ b/glide/confidence_sequences/asymptotic.py
@@ -1,0 +1,88 @@
+from typing import Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.confidence_sequences.base import ConfidenceSequence
+from glide.core.validation import _validate_bounds
+
+
+def _compute_asymptotic_bounds(
+    batch_estimates: NDArray,
+    batch_std_estimates: NDArray,
+    miscoverage: float,
+    tightest_at_batch: int,
+) -> Tuple[NDArray, NDArray]:
+    # The one-sided boundary uses the two-sided formula at twice the miscoverage,
+    # which is only defined below 1, hence the 0.5 upper bound.
+    _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=0.5, left_inclusive=False, right_inclusive=False)
+    _validate_bounds(tightest_at_batch, "tightest_at_batch", lower=1)
+    n_batches = len(batch_estimates)
+    batch_counts = np.arange(1, n_batches + 1)
+    running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
+    # Intrinsic time of the Gaussian approximation: the accumulated squared
+    # standard errors of the batch estimates, not their realized scatter.
+    variance_process = np.cumsum(batch_std_estimates**2)
+    target_position = min(tightest_at_batch, n_batches) - 1
+    target_intrinsic_time = variance_process[target_position]
+    _validate_bounds(
+        target_intrinsic_time,
+        "batch_std_estimates",
+        lower=0.0,
+        left_inclusive=False,
+        error_message=(
+            f"'batch_std_estimates' must accumulate a positive variance by batch {target_position + 1}; "
+            f"got {target_intrinsic_time!r}."
+        ),
+    )
+    doubled_miscoverage = 2.0 * miscoverage
+    log_term = -2.0 * np.log(doubled_miscoverage)
+    tuning_scale_squared = (log_term + np.log(log_term + 1.0)) / target_intrinsic_time
+    scaled_times = variance_process * tuning_scale_squared + 1.0
+    boundary_widths = np.sqrt(
+        2.0
+        * scaled_times
+        / (batch_counts**2 * tuning_scale_squared)
+        * np.log(1.0 + np.sqrt(scaled_times) / doubled_miscoverage)
+    )
+    lower_bounds = running_mean_estimates - boundary_widths
+    return running_mean_estimates, lower_bounds
+
+
+class AsymptoticConfidenceSequence(ConfidenceSequence):
+    """Anytime-valid asymptotic confidence sequence on a running mean.
+
+    A Gaussian-mixture confidence sequence whose width scales with the known
+    standard errors of the per-batch estimates instead of a worst-case bounded
+    range, so the bounds stay tight when per-batch estimates are precise. The
+    price is an asymptotic (rather than exact) time-uniform guarantee: per-batch
+    estimates must be approximately Gaussian with consistently estimated
+    variances.
+
+    Parameters
+    ----------
+    running_mean_estimates : NDArray
+        Per-look running mean of the per-batch estimates, in original metric units.
+    confidence_bounds : NDArray
+        Per-look harmful-side anytime-valid bound, in original metric units.
+
+    References
+    ----------
+    Waudby-Smith, Ian, David Arbour, Ritwik Sinha, Edward H. Kennedy, and Aaditya
+    Ramdas. "Time-uniform central limit theory and asymptotic confidence
+    sequences." arXiv preprint arXiv:2103.06476 (2024).
+
+    Robbins, Herbert. "Statistical methods related to the law of the iterated
+    logarithm." The Annals of Mathematical Statistics 41, no. 5 (1970): 1397-1409.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from glide.confidence_sequences import AsymptoticConfidenceSequence
+    >>> sequence = AsymptoticConfidenceSequence(
+    ...     running_mean_estimates=np.array([0.4, 0.6]),
+    ...     confidence_bounds=np.array([0.1, 0.55]),
+    ... )
+    >>> sequence.test_null_hypothesis(0.5, alternative="larger")
+    array([False,  True])
+    """
--- a/glide/confidence_sequences/asymptotic.py
+++ b/glide/confidence_sequences/asymptotic.py
@@ -22,16 +22,13 @@ def _compute_asymptotic_bounds(
         lower=0.0,
         error_message=f"'batch_std_estimates' must be non-negative; got {batch_std_estimates.min()!r}.",
     )
-    # The one-sided boundary uses the two-sided formula at twice the miscoverage,
-    # which is only defined below 1, hence the 0.5 upper bound.
     _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=0.5, left_inclusive=False, right_inclusive=False)
     _validate_is_integer(tightest_at_batch, "tightest_at_batch")
     _validate_bounds(tightest_at_batch, "tightest_at_batch", lower=1)
     n_batches = len(batch_estimates)
     batch_counts = np.arange(1, n_batches + 1)
     running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
-    # Intrinsic time of the Gaussian approximation: the accumulated squared
-    # standard errors of the batch estimates, not their realized scatter.
+
     variance_process = np.cumsum(batch_std_estimates**2)
     target_position = min(tightest_at_batch, n_batches) - 1
     target_intrinsic_time = variance_process[target_position]
@@ -81,7 +78,7 @@ class AsymptoticConfidenceSequence(ConfidenceSequence):
     ----------
     Waudby-Smith, Ian, David Arbour, Ritwik Sinha, Edward H. Kennedy, and Aaditya
     Ramdas. "Time-uniform central limit theory and asymptotic confidence
-    sequences." arXiv preprint arXiv:2103.06476 (2024).
+    sequences." The Annals of Statistics 52, no. 6 (2024): 2613-2640.
 
     Robbins, Herbert. "Statistical methods related to the law of the iterated
     logarithm." The Annals of Mathematical Statistics 41, no. 5 (1970): 1397-1409.
--- a/tests/unit/confidence_sequences/test_asymptotic.py
+++ b/tests/unit/confidence_sequences/test_asymptotic.py
@@ -20,21 +20,51 @@ def batch_std_estimates():
 
 
 def test_compute_asymptotic_bounds_delegates_to_validation(batch_estimates, batch_std_estimates):
-    with patch.object(asymptotic_module, "_validate_bounds") as mock_validate_bounds:
+    with (
+        patch.object(asymptotic_module, "_validate_non_empty") as mock_validate_non_empty,
+        patch.object(asymptotic_module, "_validate_equal_lengths") as mock_validate_equal_lengths,
+        patch.object(asymptotic_module, "_validate_is_integer") as mock_validate_is_integer,
+        patch.object(asymptotic_module, "_validate_bounds") as mock_validate_bounds,
+    ):
         _compute_asymptotic_bounds(batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=1)
 
-        assert mock_validate_bounds.call_count == 3
-        assert mock_validate_bounds.call_args_list[0][0] == (0.2, "miscoverage")
-        assert mock_validate_bounds.call_args_list[0][1] == {
+        mock_validate_non_empty.assert_called_once_with(batch_estimates, "batch_estimates")
+        mock_validate_equal_lengths.assert_called_once()
+        np.testing.assert_array_equal(mock_validate_equal_lengths.call_args[0][0], batch_estimates)
+        np.testing.assert_array_equal(mock_validate_equal_lengths.call_args[0][1], batch_std_estimates)
+        assert mock_validate_equal_lengths.call_args[1] == {"names": ["batch_estimates", "batch_std_estimates"]}
+        mock_validate_is_integer.assert_called_once_with(1, "tightest_at_batch")
+
+        assert mock_validate_bounds.call_count == 4
+        np.testing.assert_array_equal(mock_validate_bounds.call_args_list[0][0][0], batch_std_estimates)
+        assert mock_validate_bounds.call_args_list[0][0][1] == "batch_std_estimates"
+        assert mock_validate_bounds.call_args_list[0][1]["lower"] == 0.0
+        assert mock_validate_bounds.call_args_list[1][0] == (0.2, "miscoverage")
+        assert mock_validate_bounds.call_args_list[1][1] == {
             "lower": 0.0,
             "upper": 0.5,
             "left_inclusive": False,
             "right_inclusive": False,
         }
-        assert mock_validate_bounds.call_args_list[1][0] == (1, "tightest_at_batch")
-        assert mock_validate_bounds.call_args_list[1][1] == {"lower": 1}
-        assert mock_validate_bounds.call_args_list[2][0][1] == "batch_std_estimates"
-        assert "must accumulate a positive variance" in mock_validate_bounds.call_args_list[2][1]["error_message"]
+        assert mock_validate_bounds.call_args_list[2][0] == (1, "tightest_at_batch")
+        assert mock_validate_bounds.call_args_list[2][1] == {"lower": 1}
+        assert mock_validate_bounds.call_args_list[3][0][1] == "batch_std_estimates"
+        assert "must accumulate a positive variance" in mock_validate_bounds.call_args_list[3][1]["error_message"]
+
+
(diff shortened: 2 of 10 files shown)


imerad:
## Code review

Ran the diff against `main` through a multi-angle review (correctness, removed-behavior audit, cross-file callers, reuse/simplification/efficiency, altitude/conventions), verifying every candidate directly against the code. `make lint`, `make type-check`, and unit tests/coverage on the touched module all pass, and the ticket's numeric acceptance criterion (`_compute_asymptotic_bounds([0.4,0.6],[0.1,0.2],0.2,1) → [0.181036, 0.247749]`) checks out exactly.

### Real gaps

1. **The "pruned tests" commit (`6b62910`) deleted tests the ticket's own acceptance criteria require.** `test_compute_asymptotic_bounds_width_shrinks_with_batches` and `test_compute_asymptotic_bounds_widens_with_confidence` (both with hardcoded expected values) no longer exist anywhere, and no test in the suite constructs `ConfidenceSequence` / `EmpiricalBernsteinConfidenceSequence` / `AsymptoticConfidenceSequence` to verify `running_mean_estimates`/`confidence_bounds` are stored correctly (`test_sequence_attributes` was dropped and never relocated to `test_base.py`). Coverage stays 100% since these were value-correctness checks, not line coverage — CI won't catch this regression on its own.
2. **`_compute_asymptotic_bounds` silently accepts negative `batch_std_estimates`.** Verified: passing `[-0.1, -0.2]` returns the identical bounds as `[0.1, 0.2]`, with no error, even though a negative standard error is meaningless. `_validate_uncertainties` already exists in `glide/core/validation.py` for exactly this shape of check.
3. **No shape/type validation on `_compute_asymptotic_bounds` inputs.** A length mismatch between `batch_estimates`/`batch_std_estimates` raises a raw numpy broadcast error or an unrelated `IndexError` (if `tightest_at_batch` exceeds the shorter array) instead of a clear `ValueError`; a non-integer `tightest_at_batch` or an empty `batch_estimates` array also crash with low-level `IndexError`s. `_validate_equal_lengths`, `_validate_is_integer`, and `_validate_non_empty` already exist in `glide/core/validation.py` for these exact cases.
4. **`CHANGELOG.md`**: the new "Changed" bullet for `ConfidenceSequence` was appended after the existing "Renamed `PPIMeanMonitor`..." bullet instead of at the top of the list, per the repo's changelog convention.

### Design points (these map onto the two open checkboxes in the ticket's own TODO)

5. `ConfidenceSequence` going from `Protocol` to a concrete `@dataclass` de-duplicates `test_null_hypothesis`, but trades away the structural-typing extensibility the old docstring advertised ("any class implementing this protocol can be used..."). The same de-duplication was also reachable by keeping the `Protocol` and factoring `test_null_hypothesis` into a shared private function both classes call. Worth a short comment in the code explaining the trade-off, since the rationale currently only lives in the ticket text.
6. Neither `EmpiricalBernsteinConfidenceSequence` nor `AsymptoticConfidenceSequence` is re-decorated with `@dataclass`. That's fine today (both are docstring-only), but the ticket explicitly plans future fields on these subclasses (variance process, tuning scale). Verified: adding a subclass field as a bare annotation (mirroring the codebase's existing dataclass-field style) without also decorating the subclass is silently dropped by dataclass machinery — a `TypeError` if passed via constructor kwarg, or a silently-shared mutable class attribute if given a default. Worth a one-line comment flagging this for whoever adds the next field.
7. Minor: the 3-line running-mean computation (`n_batches`/`batch_counts`/`running_mean_estimates`) in `asymptotic.py` is copied verbatim from `_compute_empirical_bernstein_bounds` in `empirical_bernstein.py` — low severity, but a small shared helper would remove the duplication.

Not independently re-verified: the formula's fidelity to arXiv:2103.06476 eq. 50 / footnote 11 (the ticket's first open TODO item) — I confirmed it reproduces the ticket's hardcoded numeric check, which is corroborating but not the same as checking the paper source directly.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

glide/confidence_sequences/asymptotic.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    _validate_bounds(miscoverage, "miscoverage", lower=0.0, upper=0.5, left_inclusive=False, right_inclusive=False)
+    _validate_is_integer(tightest_at_batch, "tightest_at_batch")
+    _validate_bounds(tightest_at_batch, "tightest_at_batch", lower=1)
+    n_batches = len(batch_estimates)
+    batch_counts = np.arange(1, n_batches + 1)
+    running_mean_estimates = np.cumsum(batch_estimates) / batch_counts
+
+    variance_process = np.cumsum(batch_std_estimates**2)
+    target_position = min(tightest_at_batch, n_batches) - 1
```
gmartinon-ed, line 33: Pourquoi "-1" ?

glide/confidence_sequences/asymptotic.py
```diff
@@ -0,0 +1,96 @@
...
+    tuning_scale_squared = (log_term + np.log(log_term + 1.0)) / target_intrinsic_time
+    scaled_times = variance_process * tuning_scale_squared + 1.0
+    boundary_widths = np.sqrt(
+        2.0
+        * scaled_times
+        / (batch_counts**2 * tuning_scale_squared)
+        * np.log(1.0 + np.sqrt(scaled_times) / doubled_miscoverage)
+    )
+    lower_bounds = running_mean_estimates - boundary_widths
+    return running_mean_estimates, lower_bounds
```
gmartinon-ed, line 56: As-tu pu bien vérifier l'accord avec l'article de cette formule ?

glide/confidence_sequences/asymptotic.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@dataclass
+class AsymptoticConfidenceSequence(ConfidenceSequence):
```
gmartinon-ed, line 60: Quelle a été ta reflexion sur cet élément de la TODO: est-ce qu'on garde ces classes ou pas ?

tests/unit/confidence_sequences/test_asymptotic.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test_compute_asymptotic_bounds(batch_estimates, batch_std_estimates):
+    running_means, lower_bounds = _compute_asymptotic_bounds(
+        batch_estimates, batch_std_estimates, miscoverage=0.2, tightest_at_batch=1
+    )
+    np.testing.assert_allclose(running_means, np.array([0.4, 0.5]))
+    np.testing.assert_allclose(lower_bounds, np.array([0.181036, 0.247749]), atol=1e-6)
+
+
+def test_compute_asymptotic_bounds_target_clamped_to_last_batch(batch_estimates, batch_std_estimates):
```
gmartinon-ed, line 63: Pas sûr de comprendre ce test

gmartinon-ed (changes requested):

glide/confidence_sequences/asymptotic.py
imerad: `target_position` sera utilisé comme indice de tableau et l'indexation commence à 0

glide/confidence_sequences/asymptotic.py
imerad: Oui, j'ai regardé de près et ces calculs correspondent aux formules dans l'article. Ce sera définitivement validé lors de la validation scientifique

glide/confidence_sequences/asymptotic.py
imerad: La suppression des classes Asymptotic/EmpiricalBernsteinConfidenceSequence allait entrainer une refacto non négligeable au niveau des moniteurs existants. Ça empêcherait aussi d'avoir une docstring pour chacune. Au cas où un autre type de confidence sequence viendrait dans le futur et nécessiterait des méthodes différentes, c'est plus naturel d'avoir des sous-classes pour chaque cas. Enfin, c'est plus homogène avec les sous classes de confidence interval. Je suis resté sur la conclusion que supprimer ces sous classes créerait un anti-pattern

tests/unit/confidence_sequences/test_asymptotic.py
imerad: Il vérifie que le `min` définissant `target_position` plus haut fait bien son travail i.e. l'indice du batch ciblé est ramené au dernier batch au cas où il dépasse.

gmartinon-ed (approved):

PR #354, opened 2026-07-20
Add AsymptoticClassicalMeanMonitor

### LINKED ISSUE #349: AsymptoticClassicalMeanMonitor: faster label-only drift detection using per-batch error bars
## TODO

- [ ] 

## Background

`EmpiricalClassicalMeanMonitor` (the label-only monitor) suffers from the same conservativeness as its PPI sibling: it feeds per-batch sample means as bare points into the empirical-Bernstein confidence sequence, discarding the standard error each sample mean carries (its sample standard deviation divided by the square root of the batch label count). Users without proxy labels currently have no fast detection option at all.

This ticket adds `AsymptoticClassicalMeanMonitor`, completing a two-by-two family: estimator (classical versus PPI) crossed with confidence sequence (empirical-Bernstein versus asymptotic). It passes the per-batch sample means and their standard errors to the `AsymptoticConfidenceSequence` from `feat-asymptotic-confidence-sequence.md`, which is estimator-agnostic by design, so no new sequence code is needed. On the reference stream used in the classical monitor doctest, the alarm fires at batch index 3 instead of 11 (verified with a prototype).

Depends on `feat-asymptotic-confidence-sequence.md` and on `ref-rename-empirical-ppi-mean-monitor.md`, which also places the shared `_postprocess` in `glide/monitors/core.py`. No dependency on `feat-asymptotic-ppi-mean-monitor.md`: the two asymptotic monitors only share the confidence sequence and the already-extracted `_postprocess`, so they can be built in either order. Documentation updates live in the existing validation, tutorial, and user-guide tickets.

## Design choices

New class in `glide/monitors/asymptotic_classical.py`, mirroring `AsymptoticPPIMeanMonitor` in structure and `EmpiricalClassicalMeanMonitor` in `detect` signature, plus the `tightest_at_batch: int = 10` keyword and the `confidence_level` in (0.5, 1) requirement shared with the asymptotic PPI monitor. What the two classical monitors share moves into a new `glide/monitors/classical_core.py`, mirroring the `ppi_core.py` extraction: `_preprocess` (body moved verbatim from `EmpiricalClassicalMeanMonitor._preprocess`, which is removed) and `_compute_batch_estimates` returning both the per-batch means and their standard errors, fully vectorized via `np.bincount`. Postprocessing reuses the `_postprocess` function already in `glide/monitors/core.py` (extracted from both empirical monitors by `ref-rename-empirical-ppi-mean-monitor.md`); this ticket only imports it, it does not touch `EmpiricalClassicalMeanMonitor`'s postprocessing at all. The result object `ClassicalMeanMonitoringResult` is reused unchanged.

One honest caveat must appear in the docstring: the classical per-batch estimate rests on the labeled samples alone, so the Gaussian approximation behind the asymptotic guarantee is weaker than in the PPI case when batches carry few labels. In particular, a small batch of a binary metric can have all labels identical, giving a zero standard error that understates the uncertainty (and raising once the accumulated variance is zero up to the target batch). The docstring should recommend a practical minimum of labeled samples per batch (tens rather than the hard validation floor of 2) more insistently than the PPI variant does.

### classical_core shared helpers

**Location:** `glide/monitors/classical_core.py`

**Purpose:** Hold the preprocessing and per-batch estimation shared by the two classical monitors.

```python
from typing import Tuple

import numpy as np


def _preprocess(
    y: NDArray,
    batches: NDArray,
    higher_is_better: bool,
    threshold: float,
    confidence_level: float,
    metric_lower_bound: float,
    metric_upper_bound: float,
) -> Tuple[NDArray, float, NDArray, NDArray]:
    # Body moved verbatim from EmpiricalClassicalMeanMonitor._preprocess.
    ...


def _compute_batch_estimates(
    risk_y: NDArray,
    batch_codes: NDArray,
) -> Tuple[NDArray, NDArray]:
    batch_n = np.bincount(batch_codes)
    batch_sums = np.bincount(batch_codes, weights=risk_y)
    batch_mean_estimates = batch_sums / batch_n
    batch_squared_sums = np.bincount(batch_codes, weights=risk_y**2)
    # Guard against tiny negative variances from floating-point cancellation.
    batch_variances = np.maximum(batch_squared_sums - batch_n * batch_mean_estimates**2, 0.0) / (batch_n - 1)
    batch_std_estimates = np.sqrt(batch_variances / batch_n)
    return batch_mean_estimates, batch_std_estimates
```

`EmpiricalClassicalMeanMonitor.detect` is rewritten to call `_preprocess`, `_compute_batch_estimates` (dropping the returned standard errors), and the already-shared `_postprocess` from `glide/monitors/core.py`; its `_preprocess` method is removed and its unit tests move to `tests/unit/monitors/test_classical_core.py` (`_postprocess` has no method of its own left to remove: it was already extracted to `glide/monitors/core.py` and its tests moved to `test_core.py` by `ref-rename-empirical-ppi-mean-monitor.md`).

### AsymptoticClassicalMeanMonitor

**Location:** `glide/monitors/asymptotic_classical.py`

**Purpose:** Anytime-valid label-only drift monitor built on the asymptotic confidence sequence.

```python

from glide.confidence_sequences import AsymptoticConfidenceSequence
from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
from glide.mean_monitoring_results import ClassicalMeanMonitoringResult
from glide.monitors.classical_core import _compute_batch_estimates, _preprocess
from glide.monitors.core import _postprocess


class AsymptoticClassicalMeanMonitor:
    """Anytime-valid label-only drift monitor leveraging each batch estimate's error bar.

    It computes a per-batch sample mean of the labeled values and monitors the
    running mean of those estimates against a user-supplied ``threshold``,
    through a Gaussian-mixture asymptotic confidence sequence whose width scales
    with the standard error of each batch mean, so genuine drifts are detected
    early. The false-alarm guarantee is asymptotic: each batch needs enough
    labeled samples (tens in practice) for its sample mean to be approximately
    Gaussian with a consistently estimated variance.

    References
    ----------

    Examples
    --------
    >>> import numpy as np
    >>> from glide.monitors import AsymptoticClassicalMeanMonitor
    >>> pre_drift_batch = np.array([0.0, 0.2, np.nan, np.nan])
    >>> post_drift_batch = np.array([0.8, 1.0, np.nan, np.nan])
    >>> y = np.hstack([pre_drift_batch, np.tile(post_drift_batch, 50)])
    >>> monitor = AsymptoticClassicalMeanMonitor()
    >>> result = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
    >>> result.drift_detected
    True
    >>> result.first_alarm_index
    3
    """

    def detect(
        self,
        y: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
        metric_name: str = "Metric",
        tightest_at_batch: int = 10,
    ) -> ClassicalMeanMonitoringResult:

        [Parameter documentation mirrored from EmpiricalClassicalMeanMonitor.detect
        plus the two additions of AsymptoticPPIMeanMonitor.detect: the (0.5, 1)
        requirement on ``confidence_level`` and the ``tightest_at_batch`` guidance
        (leave the default of 10 alone; tightness only, never validity).]
        """
        _validate_bounds(
            confidence_level,
            "confidence_level",
            lower=0.5,
            upper=1,
            left_inclusive=False,
            right_inclusive=False,
            error_message=(
                f"'confidence_level' must be in (0.5, 1) for the asymptotic monitor; got {confidence_level!r}."
            ),
        )
        risk_y, risk_threshold, batch_codes, batch_n = _preprocess(
            y, batches, higher_is_better, threshold, confidence_level, metric_lower_bound, metric_upper_bound
        )
        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(risk_y, batch_codes)
        risk_running_means, risk_lower_bounds = _compute_asymptotic_bounds(
        )
        running_means, confidence_bounds, metric_batch_estimates = _postprocess(
            risk_running_means, risk_lower_bounds, batch_mean_estimates,
            higher_is_better, metric_lower_bound, metric_upper_bound,
        )
            running_mean_estimates=running_means,
        )
            metric_name=metric_name,
            alarm_threshold=threshold,
            batch_mean_estimates=metric_batch_estimates,
            batch_n=batch_n,
        )
        return result
```

Export from `glide/monitors/__init__.py` and add a `:::` entry in `docs/api/monitors.md`.

**Unit tests.** The `_compute_batch_estimates` tests live in `tests/unit/monitors/test_classical_core.py` (mirror rule), alongside the `_preprocess` tests moved from `test_empirical_classical.py` (the `_postprocess` tests are not part of this move: they already live in `test_core.py`); the analytical anchor is `risk_y = [0.49, 0.51, 0.5, 0.54]` with `batch_codes = [0, 0, 1, 1]` giving means `[0.5, 0.52]` and standard errors `[0.01, 0.02]`. `tests/unit/monitors/test_asymptotic_classical.py` mirrors the `# --- detect ---` section of `test_empirical_classical.py`, reusing its exact `y`/`batches`/`monitor` fixtures:

- `test_detect_delegates_to_validation`: patch `_validate_bounds` in the `asymptotic_classical` module and check the single direct call with `(0.8, "confidence_level")`, `lower=0.5`, `upper=1`, both sides exclusive, and the `(0.5, 1)` error message.
- `test_detect_is_valid_monitoring_result`: `isinstance` checks for `ClassicalMeanMonitoringResult` and `AsymptoticConfidenceSequence`, `monitor_name == "AsymptoticClassicalMeanMonitor"`, finite running means, running means at or above the bounds.
- `test_detect_metadata`: mirror of the empirical version (metric name, orientation, threshold, confidence level, `batch_n`).
- `test_detect_custom_confidence_level`: hardcoded anchor on the fixture stream (verified against a prototype run):

```python
def test_detect_custom_confidence_level(monitor, y, batches):
    expected_running_means = np.array([0.5, 0.51])
    expected_confidence_bounds = np.array([0.472, 0.484])

    result = monitor.detect(
        y, batches, higher_is_better=False, threshold=0.5, metric_name="perf", confidence_level=0.85
    )

    assert result.confidence_level == 0.85
    np.testing.assert_allclose(result.running_means, expected_running_means, atol=0.001)
    np.testing.assert_allclose(result.confidence_bounds, expected_confidence_bounds, atol=0.001)
```

## Functional tests

`tests/functional/monitors/test_asymptotic_classical.py` mirrors `tests/functional/monitors/test_empirical_classical.py`: same module docstring style, same `y`/`batches` fixtures, `test_detect_*` names, and a one-line docstring per test.

**`test_detect_no_alarm_without_drift`** — "A stable stream below the threshold never alarms." On the shared fixtures with a threshold far above the running risk, `drift_detected` must be `False`.

**`test_detect_prefix_consistency`** — "Detecting on a growing history is prefix-consistent with detecting on the full history." Mirror of the empirical test, passing `tightest_at_batch=2` explicitly on both calls so the tuning target is already reached within the prefix (see corner cases).

**`test_detect_higher_is_better_symmetry`** — "Monitoring a performance is the mirror image of monitoring its complement as a risk." Mirror of the empirical test.

## Corner cases

- `confidence_level <= 0.5`: raise `ValueError` naming the (0.5, 1) requirement.
- A batch whose labeled values are all identical yields a zero standard error; if this holds for every batch up to `tightest_at_batch`, the `ValueError` from the confidence sequence computation propagates.
- A batch with exactly 2 identical labels among varied batches: no error, but its zero standard error contributes nothing to the variance process (document, do not raise).
- `metric_lower_bound`/`metric_upper_bound` different from (0, 1): standard errors are computed in unit-risk space after rescaling, so they must not be rescaled a second time on output (only means and bounds pass through `_postprocess`).
- Successive `detect` calls on growing histories shorter than `tightest_at_batch`: the clamped tuning target moves with the history, so the bounds of early batches recalibrate between calls; from batch `tightest_at_batch` onward they are prefix-consistent. Document this in the `detect` docstring, as in the asymptotic PPI monitor.

## Acceptance criteria

- [ ] `AsymptoticClassicalMeanMonitor` is importable from `glide.monitors`, rendered in `docs/api/monitors.md`, and its doctest reproduces `first_alarm_index == 3` on the reference stream.
- [ ] `EmpiricalClassicalMeanMonitor` produces bit-identical results after the `classical_core` extraction (existing unit and functional tests pass with only import updates).
- [ ] `_compute_batch_estimates` returns both mean and standard-error arrays and matches the analytical anchor in `test_classical_core.py` (means `[0.5, 0.52]`, standard errors `[0.01, 0.02]`).
- [ ] The unit tests mirror the `detect` section of `test_empirical_classical.py` (same fixtures, same test names) and the custom-confidence anchor passes.
- [ ] The five functional tests above pass with the structure of `test_empirical_classical.py` (shared fixtures, `test_detect_*` names, one-line docstrings).

imerad:
### Description

- Adds `AsymptoticClassicalMeanMonitor`, a label-only drift monitor built on `AsymptoticConfidenceSequence`, using the batch estimates' standard errors instead of the empirical-Bernstein bound used by `EmpiricalClassicalMeanMonitor`.
- Factors out the internals shared between `EmpiricalClassicalMeanMonitor` and the new monitor into `glide/monitors/classical_core.py`.
- Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=214000585&issue=EmertonData%7Cglide%7C349) Closes #349

Any noteworthy implementation decisions or trade-offs?
- The asymptotic monitor doesn't actually need the metric lower/upper bounds, which are used elsewhere to scale and unscale the data, since there's no boundedness assumption behind the asymptotic method. The bound-handling code was factorized with the pre-existing `EmpiricalClassicalMeanMonitor` though. This doesn't invalidate the asymptotic monitor's behavior, but it does force users to supply valid lower/upper bounds on their data even though the asymptotic method doesn't strictly need them. Surgically restricting the bound checks/normalization to only the monitors that need them would make the code messy, so that's left to a possible future refactor.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/monitors/asymptotic_classical.py
+++ b/glide/monitors/asymptotic_classical.py
@@ -12,13 +12,13 @@ class AsymptoticClassicalMeanMonitor:
     """Anytime-valid label-only drift monitor leveraging each batch estimate's
     standard deviation.
 
-    It computes a per-batch sample mean of the labeled values and monitors the
-    running mean of those estimates against a user-supplied ``threshold``,
-    through a Gaussian-mixture asymptotic confidence sequence whose width scales
-    with the standard error of each batch mean, so genuine drifts are detected
-    early. The false-alarm guarantee is asymptotic: each batch needs enough
-    labeled samples (preferably tens) for its sample mean to be approximately
-    Gaussian with a consistently estimated variance.
+    It computes a per-batch sample mean of the labeled values, then tracks the
+    running mean of those estimates against a user-supplied ``threshold``. The
+    comparison uses a Gaussian-mixture asymptotic confidence sequence whose
+    width scales with the standard error of each batch mean, so genuine drifts
+    get flagged early. The false-alarm guarantee is asymptotic: each batch
+    needs enough labeled samples (preferably tens) for its sample mean to be
+    approximately Gaussian with a consistently estimated variance.
 
     References
     ----------
--- a/glide/monitors/asymptotic_classical.py
+++ b/glide/monitors/asymptotic_classical.py
@@ -146,7 +146,7 @@ class AsymptoticClassicalMeanMonitor:
         risk_y, _, batch_codes, batch_n = _preprocess(
             y, batches, higher_is_better, threshold, confidence_level, metric_lower_bound, metric_upper_bound
         )
-        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(risk_y, batch_codes)
+        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(risk_y, batch_codes, batch_n)
         miscoverage = 1.0 - confidence_level
         risk_running_means, risk_lower_bounds = _compute_asymptotic_bounds(
             batch_mean_estimates, batch_std_estimates, miscoverage, tightest_at_batch
--- a/glide/monitors/asymptotic_classical.py
+++ b/glide/monitors/asymptotic_classical.py
@@ -68,10 +68,10 @@ class AsymptoticClassicalMeanMonitor:
         every call makes the anytime-valid guarantee hold jointly over all calls.
         Alternatively the data may be restricted to the most recent batches, in which
         case the guarantee holds within each restriction but not across the moving
-        history as a whole. When the history is shorter than ``tightest_at_batch``,
-        the tuning target moves with it, so bounds up to that batch recalibrate
-        between successive calls; from ``tightest_at_batch`` onward they are
-        prefix-consistent.
+        history as a whole. When the history has fewer batches than ``tightest_at_batch``,
+        the tuning target is set to the last batch. Prefix-consistency only holds between
+        calls using the same ``tightest_at_batch`` value and with histories containing
+        at least ``tightest_at_batch`` batches.
 
         Parameters
         ----------
@@ -146,7 +146,7 @@ class AsymptoticClassicalMeanMonitor:
         risk_y, _, batch_codes, batch_n = _preprocess(
             y, batches, higher_is_better, threshold, confidence_level, metric_lower_bound, metric_upper_bound
         )
-        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(risk_y, batch_codes, batch_n)
+        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(risk_y, batch_codes)
         miscoverage = 1.0 - confidence_level
         risk_running_means, risk_lower_bounds = _compute_asymptotic_bounds(
             batch_mean_estimates, batch_std_estimates, miscoverage, tightest_at_batch
--- a/glide/monitors/asymptotic_classical.py
+++ b/glide/monitors/asymptotic_classical.py
@@ -1,53 +1,45 @@
 from numpy.typing import NDArray
 
-from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
-from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
+from glide.confidence_sequences import AsymptoticConfidenceSequence
+from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
+from glide.core.validation import _validate_bounds
 from glide.mean_monitoring_results import ClassicalMeanMonitoringResult
 from glide.monitors.classical_core import _compute_batch_estimates, _preprocess
 from glide.monitors.core import _postprocess
 
 
-class EmpiricalClassicalMeanMonitor:
-    """Anytime-valid drift monitor over a batched dataset of labels.
+class AsymptoticClassicalMeanMonitor:
+    """Anytime-valid label-only drift monitor leveraging each batch estimate's
+    standard deviation.
 
-    It uses the plain sample mean per batch. Unlabeled entries are passed as
-    ``np.nan`` and dropped per batch. To be used on accumulated non-overlapping
-    production batches by calling :meth:`detect` on the whole accumulated dataset at
-    any time. Data is passed oldest batch first, and every batch is monitored. The
-    monitor computes a sample-mean estimate per batch, builds an anytime-valid
-    confidence sequence on the running mean of those estimates, and raises a drift
-    alarm when it crosses a user-supplied ``threshold`` (the worst tolerable metric value).
-    Because the bounds are valid at all times simultaneously, :meth:`detect` may be called
-    after every new batch without inflating the false-alarm probability.
+    It computes a per-batch sample mean of the labeled values and monitors the
+    running mean of those estimates against a user-supplied ``threshold``,
+    through a Gaussian-mixture asymptotic confidence sequence whose width scales
+    with the standard error of each batch mean, so genuine drifts are detected
+    early. The false-alarm guarantee is asymptotic: each batch needs enough
+    labeled samples (preferably tens) for its sample mean to be approximately
+    Gaussian with a consistently estimated variance.
 
     References
     ----------
-    Howard, Steven R., Aaditya Ramdas, Jon McAuliffe, and Jasjeet Sekhon. "Time-uniform,
-    nonparametric, nonasymptotic confidence sequences." The Annals of Statistics 49,
-    no. 2 (2021): 1055-1080.
-
-    Podkopaev, Aleksandr, and Aaditya Ramdas. "Tracking the risk of a deployed model
-    and detecting harmful distribution shifts." International Conference on Learning
-    Representations (ICLR), 2022.
-
-    Waudby-Smith, Ian, and Aaditya Ramdas. "Estimating means of bounded random
-    variables by betting." Journal of the Royal Statistical Society Series B:
-    Statistical Methodology 86, no. 1 (2024): 1-27.
+    Waudby-Smith, Ian, David Arbour, Ritwik Sinha, Edward H. Kennedy, and Aaditya
+    Ramdas. "Time-uniform central limit theory and asymptotic confidence
+    sequences." The Annals of Statistics 52, no. 6 (2024): 2613-2640.
 
     Examples
     --------
     >>> import numpy as np
-    >>> from glide.monitors import EmpiricalClassicalMeanMonitor
+    >>> from glide.monitors import AsymptoticClassicalMeanMonitor
     >>> pre_drift_batch = np.array([0.0, 0.2, np.nan, np.nan])
     >>> post_drift_batch = np.array([0.8, 1.0, np.nan, np.nan])
     >>> y = np.hstack([pre_drift_batch, np.tile(post_drift_batch, 50)])
     >>> batches = np.repeat(np.arange(51), 4)
-    >>> monitor = EmpiricalClassicalMeanMonitor()
+    >>> monitor = AsymptoticClassicalMeanMonitor()
     >>> result = monitor.detect(y, batches, higher_is_better=False, threshold=0.5)
     >>> result.drift_detected
     True
     >>> result.first_alarm_index
-    11
+    3
     """
 
     def detect(
@@ -60,13 +52,14 @@ class EmpiricalClassicalMeanMonitor:
         confidence_level: float = 0.8,
         metric_lower_bound: float = 0.0,
         metric_upper_bound: float = 1.0,
+        tightest_at_batch: int = 10,
     ) -> ClassicalMeanMonitoringResult:
         """Detect a drift of the running mean across a batched dataset.
 
-        Splits the data by batch, computes a sample-mean estimate per batch, and
-        builds an anytime-valid empirical-Bernstein confidence sequence on the
-        running mean of those estimates. An alarm is raised at every batch where the
-        sequence crosses the user-supplied ``threshold``.
+        Splits the data by batch, computes a sample-mean estimate and its standard
+        error per batch, and builds an anytime-valid asymptotic confidence sequence
+        on the running mean of those estimates. An alarm is raised at every batch
+        where the sequence crosses the user-supplied ``threshold``.
 
         Rows must be ordered oldest batch first and grouped into contiguous blocks;
         identifier values are not compared, so any label type works. Batches must be
@@ -75,14 +68,18 @@ class EmpiricalClassicalMeanMonitor:
         every call makes the anytime-valid guarantee hold jointly over all calls.
         Alternatively the data may be restricted to the most recent batches, in which
         case the guarantee holds within each restriction but not across the moving
-        history as a whole.
+        history as a whole. When the history is shorter than ``tightest_at_batch``,
+        the tuning target moves with it, so bounds up to that batch recalibrate
+        between successive calls; from ``tightest_at_batch`` onward they are
+        prefix-consistent.
 
         Parameters
         ----------
         y : NDArray
             Array of observations, shape ``(n_samples,)``. Unlabeled entries are
             ``np.nan`` and are dropped per batch; every batch must keep at least 2
-            labeled entries.
+            labeled entries, though tens are recommended in practice for the
+            asymptotic guarantee to hold.
         batches : NDArray
             Array of batch identifiers, shape ``(n_samples,)``. Rows must be ordered
             oldest batch first and grouped into contiguous blocks. Identifier values
@@ -104,11 +101,15 @@ class EmpiricalClassicalMeanMonitor:
             raises "80% confident" alarms: under no drift it has at most a 20% chance
             of raising a false alarm. Raising this value demands more evidence before
             alarming, so alarms become more trustworthy but arrive later; lowering it
-            detects sooner at the cost of more false alarms.
+            detects sooner at the cost of more false alarms. Must be in ``(0.5, 1)``.
         metric_lower_bound : float, optional
             Known lower bound of the metric. Defaults to ``0.0``.
         metric_upper_bound : float, optional
             Known upper bound of the metric. Defaults to ``1.0``.
+        tightest_at_batch : int, optional
+            The batch index (1-indexed) at which the confidence sequence is tuned to
+            be tightest. Defaults to ``10``. Only affects tightness, not validity:
+            bounds stay anytime-valid at every batch regardless of this choice.
 
         Returns
         -------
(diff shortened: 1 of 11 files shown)


imerad:
## Code review findings

Reviewed the diff across correctness, reuse, simplification, efficiency, altitude, and CLAUDE.md conventions. 9 findings, most severe first.

### 1. Numerical instability in the new variance formula
`glide/monitors/classical_core.py:83`

`_compute_batch_estimates` computes per-batch variance with the naive one-pass "sum-of-squares minus n·mean²" formula, which suffers catastrophic cancellation, instead of the numerically stable two-pass formula (`np.var(ddof=1)`) used everywhere else in the codebase (`classical.py`, `stratified_classical.py`, `clustered_classical.py`).

Reproduced directly: for 1000 samples with true mean 0.7 and true std 1e-8 (clipped to [0,1] as risk-scaled data would be), the naive formula returns variance/std of exactly `0.0`, while `np.var(y, ddof=1)` correctly recovers ~9.55e-17 (std ~3.09e-10). Since `AsymptoticClassicalMeanMonitor` feeds this std directly into `variance_process = cumsum(batch_std_estimates**2)` (`glide/confidence_sequences/asymptotic.py`), a spuriously zeroed std either narrows the confidence-sequence width (breaking the anytime-valid guarantee) or spuriously triggers the "must accumulate a positive variance" `ValueError` for batches that actually have tiny-but-nonzero variance.

### 2. Crash on all-NaN input
`glide/monitors/classical_core.py:61`

`_preprocess`'s `_validate_non_empty(y, "y")` only checks raw length, not the count of non-NaN entries, so an all-NaN `y` crashes with a raw, unhelpful `ValueError` instead of a clear message.

Reproduced directly: calling `_preprocess` with `y=np.array([nan,nan,nan,nan])` raises `"ValueError: zero-size array to reduction operation minimum which has no identity"` from `np.argmin(batch_n)` on an empty array, instead of a `ValueError` naming `'y'` and ending with a period per this repo's error-handling convention. This bug pre-dates the refactor (present in `main`'s `empirical_classical.py`) but its reach is now extended to the new `AsymptoticClassicalMeanMonitor.detect` being introduced in this PR.

### 3. Redundant/dead validation of `confidence_level`
`glide/monitors/asymptotic_classical.py:135`

`AsymptoticClassicalMeanMonitor.detect` validates `confidence_level` against the strict `(0.5, 1)` bound at the public boundary, then calls `_preprocess`, which re-validates the same variable against the strictly wider `(0, 1)` bound (`classical_core.py:22-24`) — a check that can never fail on this call path.

This directly violates this repo's CLAUDE.md rule: *"Only validate at public API boundaries. Centralizing checks into a private `_preprocess` helper is fine; re-checking the same condition in downstream private helpers is not."* Any `(0.5,1)`-valid `confidence_level` trivially satisfies `(0,1)`, so the second check is dead code; a future maintainer loosening `_preprocess`'s generic bound would wrongly believe it affects the asymptotic monitor's actual validation range, which is really governed by the outer check alone.

### 4. `docs/api/index.md` not updated
`docs/api/index.md:57`

The "Monitors > Classical" registry table lists only `EmpiricalClassicalMeanMonitor`; no row was added for the new `AsymptoticClassicalMeanMonitor`, even though `docs/api/monitors.md` gained the corresponding mkdocstrings directive.

A user browsing `docs/api/index.md` (the canonical API listing, kept in sync for `AsymptoticConfidenceSequence` in the prior PR #347) will not discover `AsymptoticClassicalMeanMonitor` exists at all, despite it being fully exported and documented on its own page.

### 5. Duplicated min-samples validation
`glide/monitors/classical_core.py:60`

`_preprocess` hand-rolls a "≥2 non-NaN values per batch" check via `np.bincount` + `np.argmin` + `_validate_bounds` instead of reusing `glide/core/validation._validate_min_samples`, which already implements the identical minimum-sample-count rule and is used by `glide/estimators/classical.py` and `stratified_classical.py`.

If the minimum-sample-count rule or its wording changes, a maintainer is likely to update `_validate_min_samples` and its known call sites but miss this hand-rolled duplicate, leaving monitors silently out of sync with estimators on the same validation rule. (Note: `_validate_min_samples`'s message hardcodes "stratum" terminology and isn't vectorized across groups, so direct reuse would need a small generalization — this is a real but not trivial-to-fix duplication.)

### 6. Duplicated result-construction block
`glide/monitors/asymptotic_classical.py:166` vs `glide/monitors/empirical_classical.py:155`

The `miscoverage = 1.0 - confidence_level` computation and the entire `ClassicalMeanMonitoringResult` construction block are copy-pasted near-verbatim between the two monitor files, despite `classical_core.py` being created in this same diff specifically to hold shared monitor internals.

If `ClassicalMeanMonitoringResult`'s constructor gains a new required field or miscoverage semantics change, a developer updating one monitor and running only its own tests can ship while the other monitor silently drifts or breaks, since the two blocks aren't backed by one shared implementation.

### 7. Wasted computation in the shared helpers
`glide/monitors/classical_core.py:80`

`_compute_batch_estimates` recomputes `batch_n` via `np.bincount(batch_codes)` even though `_preprocess` already computed and returned an identical `batch_n` to both callers, and always computes a per-batch std/variance that `EmpiricalClassicalMeanMonitor` immediately discards (`empirical_classical.py:136`), while `_preprocess` always computes a risk-scaled threshold that `AsymptoticClassicalMeanMonitor` discards (`asymptotic_classical.py:146`).

Every call to `detect()` on either monitor now pays for extra bincount passes and a variance/std computation it doesn't need, purely as a byproduct of both monitors sharing one non-tailored interface — minor per-call inefficiency, and a sign the shared functions weren't re-split to fit both callers' needs after the refactor.

### 8. Docstrings on test functions
`tests/functional/monitors/test_asymptotic_classical.py:32,45,58`

Three new test functions carry docstrings, violating this repo's explicit CLAUDE.md rule: *"No docstrings for test functions or private methods — test code is self-documenting."*

This new file propagates the same violation already present in the sibling `tests/functional/monitors/test_empirical_classical.py`, spreading the inconsistency to the newly added monitor's tests instead of cleaning it up.

### 9. Missing test coverage for the variance floor
`tests/unit/monitors/test_classical_core.py:116`

`test_compute_batch_estimates` only exercises a normal, positive-variance case (expected std `[0.01, 0.02]`); no unit test covers a degenerate batch where computed variance is exactly zero or negative before the `np.maximum(..., 0.0)` floor is applied.

The floor exists precisely to guard against the catastrophic-cancellation scenario in finding #1, but with no test forcing that branch, a future change to the variance formula (or an accidental regression) wouldn't be caught by the test suite even under 100%-coverage enforcement, since `np.maximum` isn't a branch line coverage tracks.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

glide/monitors/asymptotic_classical.py
```diff
@@ -0,0 +1,176 @@
...
+        where the sequence crosses the user-supplied ``threshold``.
+
+        Rows must be ordered oldest batch first and grouped into contiguous blocks;
+        identifier values are not compared, so any label type works. Batches must be
+        non-overlapping (no shared samples), and successive calls must be made on
+        growing histories of the same data; passing the full accumulated dataset at
+        every call makes the anytime-valid guarantee hold jointly over all calls.
+        Alternatively the data may be restricted to the most recent batches, in which
+        case the guarantee holds within each restriction but not across the moving
+        history as a whole. When the history is shorter than ``tightest_at_batch``,
```
gmartinon-ed: Cette dernière phrase sur tightest_at_batch est difficile à comprendre

tests/unit/monitors/test_asymptotic_classical.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    assert result.confidence_level == 0.85
+    np.testing.assert_allclose(result.running_means, expected_running_means, atol=0.001)
+    np.testing.assert_allclose(result.confidence_bounds, expected_confidence_bounds, atol=0.001)
+
+
+def test_detect_invalid_confidence_level(monitor, y, batches):
+    with pytest.raises(ValueError, match=r"'confidence_level' must be in \(0.5, 1\) for the asymptotic monitor"):
+        monitor.detect(y, batches, higher_is_better=False, threshold=0.5, confidence_level=0.5)
```
gmartinon-ed: Ce test n'est pas censé être déjà capturé par le "test_delegate_validation" ?

glide/monitors/classical_core.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+    risk_y = _scale_to_unit_risk(labeled_values, metric_lower_bound, metric_upper_bound, higher_is_better)
+    risk_threshold = _scale_to_unit_risk(threshold, metric_lower_bound, metric_upper_bound, higher_is_better)
+    return risk_y, risk_threshold, batch_codes, batch_n
+
+
+def _compute_batch_estimates(
+    risk_y: NDArray,
+    batch_codes: NDArray,
+    batch_n: NDArray,
```
gmartinon-ed: Je chipote, mais est-ce que ce n'est pas error-prone d'avoir batch_n qui encode une information déjà présente dans batch_codes ? ça laisse ouvert la possibilité d'avoir une incohérence entre les deux

glide/monitors/classical_core.py
```diff
@@ -0,0 +1,94 @@
...
+) -> Tuple[NDArray, NDArray]:
+    batch_sums = np.bincount(batch_codes, weights=risk_y)
+    batch_mean_estimates = batch_sums / batch_n
+    batch_squared_sums = np.bincount(batch_codes, weights=risk_y**2)
+    batch_variances = np.maximum(batch_squared_sums - batch_n * batch_mean_estimates**2, 0.0) / (batch_n - 1)
```
gmartinon-ed: Claude faisait remarquer que ça pouvait être instable numériquement, qu'en penses-tu ?

gmartinon-ed (changes requested):

glide/monitors/asymptotic_classical.py
imerad: Traité

tests/unit/monitors/test_asymptotic_classical.py
imerad: Je l'avais loupé. Supprimé

glide/monitors/classical_core.py
imerad: Claude se plaignait que le calcul de `batch_n` était redondant. Comme ça ne coûte pas cher je conviens de revenir vers la version qui refait le calcul.

glide/monitors/classical_core.py
imerad: Je n'y avais pas accordé d'importance car le risque se limite à avoir des valeurs de l'ordre de 1e-10 qui se retrouvent à 0 quand les échantillons sont très peu dispersés dans un batch. En en rediscutant avec Claude, il a une version du calcul qui reste vectorisée et évite la potentielle instabilité. J'applique ce changement.

gmartinon-ed (approved):

PR #356, opened 2026-07-20
Add AsymptoticPPIMeanMonitor

### LINKED ISSUE #350: AsymptoticPPIMeanMonitor: faster drift detection using per-batch error bars
## TODO

- [ ] 

## Background

`EmpiricalPPIMeanMonitor` computes one PPI estimate per batch and feeds those estimates, as bare points, into an empirical-Bernstein confidence sequence. Each PPI estimate also comes with a standard error (an error bar quantifying its sampling noise, already computable via `_compute_std_estimate` in `glide/estimators/ppi_core.py`), and ignoring it makes the monitor very conservative: on realistic streams the empirical-Bernstein bound is an order of magnitude wider than the error bars justify, so genuine drifts are detected late.

This ticket adds `AsymptoticPPIMeanMonitor`, a sibling monitor with the same inputs and the same result object, which passes the per-batch estimates *and* their standard errors to the `AsymptoticConfidenceSequence` delivered by `feat-asymptotic-confidence-sequence.md`. On the reference stream used in the monitor doctests, the alarm fires at batch index 2 instead of 11 (verified with a prototype). The price is an asymptotic guarantee: the false-alarm budget holds in the large-batch regime rather than exactly, so each batch needs enough samples for its PPI estimate to be approximately Gaussian.

Depends on `feat-asymptotic-confidence-sequence.md` and on the rename in `ref-rename-empirical-ppi-mean-monitor.md`. Documentation (user guide, tutorial, validation notebook) is covered by separate tickets and must not be touched here, except the API reference page.

## Design choices

New class in `glide/monitors/asymptotic_ppi.py`, mirroring `EmpiricalPPIMeanMonitor` in structure and `detect` signature, with one extra keyword `tightest_at_batch: int = 10` forwarded to the confidence sequence. Everything the two PPI monitors share moves into a new `glide/monitors/ppi_core.py` (same pattern as `glide/estimators/ppi_core.py`): input preprocessing and the per-batch estimation loop, which now returns both mean and standard-error arrays (the empirical monitor discards the standard errors). Output postprocessing already lives in `glide/monitors/core.py` as a standalone `_postprocess` function, extracted from the two empirical monitors by `ref-rename-empirical-ppi-mean-monitor.md`; this ticket only imports it, it does not move or delete anything there. Per the extraction rule, `EmpiricalPPIMeanMonitor` loses its `_preprocess` method and calls the shared function directly; behaviour is unchanged.

Two deliberate differences from the empirical monitor: the batch estimates are *not* clipped to [0, 1] before building the sequence (the Gaussian theory does not need boundedness, and clipping would bias the estimates the approximation is centered on), and `confidence_level` must lie in (0.5, 1) because the one-sided tuning formula is undefined at or below 0.5.

The power-tuning parameter must stay fitted on strictly earlier batches (as in the shared loop), even though the asymptotic theory would formally tolerate in-batch fitting: a λ fitted within its own batch correlates with that batch's noise and biases the estimate by O(1/n_labels) with a systematic sign, and since the asymptotic bound shrinks like 1/√t while that bias does not, the sequence would eventually sit on the wrong side of the true running risk. A predictable λ keeps every batch estimate exactly conditionally unbiased at no cost. Do not relax this when touching the shared loop.

### ppi_core shared helpers

**Location:** `glide/monitors/ppi_core.py`

**Purpose:** Hold the preprocessing, postprocessing, and per-batch PPI estimation shared by the two PPI monitors.

```python
from typing import Tuple

import numpy as np

from glide.estimators.ppi_core import _compute_mean_estimate, _compute_std_estimate, _compute_tuning_parameter


def _preprocess(
    y_true: NDArray,
    y_proxy: NDArray,
    batches: NDArray,
    higher_is_better: bool,
    threshold: float,
    confidence_level: float,
    metric_lower_bound: float,
    metric_upper_bound: float,
) -> Tuple[NDArray, NDArray, float, NDArray, NDArray, NDArray]:
    # Body moved verbatim from EmpiricalPPIMeanMonitor._preprocess.
    ...


def _compute_batch_estimates(
    risk_y_true: NDArray,
    risk_y_proxy: NDArray,
    batch_codes: NDArray,
    power_tuning: bool,
) -> Tuple[NDArray, NDArray]:
    n_batches = batch_codes[-1] + 1
    batch_mean_estimates = np.empty(n_batches)
    batch_std_estimates = np.empty(n_batches)
        if position == 0 or (not power_tuning):
            tuning_parameter = 1.0
        else:
            earlier_mask = batch_codes < position
            y_true_earlier, y_proxy_labeled_earlier, y_proxy_unlabeled_earlier, _ = _split_labeled_unlabeled(
                risk_y_true[earlier_mask], risk_y_proxy[earlier_mask]
            )
            tuning_parameter = _compute_tuning_parameter(
                y_true_earlier, y_proxy_labeled_earlier, y_proxy_unlabeled_earlier, power_tuning
            )
        batch_mask = batch_codes == position
        y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, _ = _split_labeled_unlabeled(
            risk_y_true[batch_mask], risk_y_proxy[batch_mask]
        )
        batch_mean_estimates[position] = _compute_mean_estimate(
            y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, tuning_parameter
        )
        batch_std_estimates[position] = _compute_std_estimate(
            y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, tuning_parameter
        )
    return batch_mean_estimates, batch_std_estimates
```

`EmpiricalPPIMeanMonitor.detect` thus calls the two `ppi_core` functions (dropping the returned `batch_std_estimates`) plus the already-shared `_postprocess` from `glide/monitors/core.py`; the removed `_preprocess` method's unit tests move to `tests/unit/monitors/test_ppi_core.py` (the `_postprocess` tests already live in `test_core.py`, moved there by `ref-rename-empirical-ppi-mean-monitor.md`).

### AsymptoticPPIMeanMonitor

**Location:** `glide/monitors/asymptotic_ppi.py`

**Purpose:** Anytime-valid PPI drift monitor built on the asymptotic confidence sequence.

```python

from glide.confidence_sequences import AsymptoticConfidenceSequence
from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
from glide.mean_monitoring_results import PredictionPoweredMeanMonitoringResult
from glide.monitors.core import _postprocess
from glide.monitors.ppi_core import _compute_batch_estimates, _preprocess


class AsymptoticPPIMeanMonitor:
    """Anytime-valid drift monitor leveraging the error bar of each batch estimate.

    It computes a per-batch Prediction-Powered Inference (PPI) estimate with a
    predictable power-tuning parameter and monitors the running mean of those
    estimates against a user-supplied ``threshold``, through a Gaussian-mixture
    asymptotic confidence sequence whose width scales with the standard error of
    each batch estimate, so genuine drifts are detected early. The false-alarm
    guarantee is asymptotic: it requires each batch to hold enough samples for
    its PPI estimate to be approximately Gaussian with a consistently estimated
    variance.

    References
    ----------

    Zhang, Guangyi, Yunlong Cai, Guanding Yu, and Osvaldo Simeone. "Prediction-powered
    risk monitoring of deployed models for detecting harmful distribution shifts."
    arXiv preprint arXiv:2602.02229 (2026).

    Examples
    --------
    >>> import numpy as np
    >>> from glide.monitors import AsymptoticPPIMeanMonitor
    >>> pre_drift_y_true = np.array([0.0, 0.2, np.nan, np.nan])
    >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
    >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
    >>> post_drift_y_proxy = np.array([0.8, 1.0, 0.8, 1.0])
    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
    >>> monitor = AsymptoticPPIMeanMonitor()
    >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
    >>> result.drift_detected
    True
    >>> result.first_alarm_index
    2
    """

    def detect(
        self,
        y_true: NDArray,
        y_proxy: NDArray,
        batches: NDArray,
        higher_is_better: bool,
        threshold: float,
        metric_name: str = "Metric",
        power_tuning: bool = True,
        tightest_at_batch: int = 10,

        [Parameter documentation mirrored from EmpiricalPPIMeanMonitor.detect, with
        two additions: ``confidence_level`` must lie in (0.5, 1), and
        ``tightest_at_batch`` documents the batch index (from 1) at which the
        confidence sequence is tuned to be tightest, clamped to the last batch when
        fewer batches are available. Its docstring must state that the default of 10
        is safe to leave alone: the setting affects tightness only, never the
        false-alarm guarantee, and the width penalty for being off by a factor of 10
        is below 10%. The Raises section adds these two conditions and drops the
        clipping-related wording.]
        """
        _validate_bounds(
            confidence_level,
            "confidence_level",
            lower=0.5,
            upper=1,
            left_inclusive=False,
            right_inclusive=False,
            error_message=(
                f"'confidence_level' must be in (0.5, 1) for the asymptotic monitor; got {confidence_level!r}."
            ),
        )
        risk_y_true, risk_y_proxy, risk_threshold, batch_codes, batch_n_true, batch_n_proxy = _preprocess(
            y_true, y_proxy, batches, higher_is_better, threshold, confidence_level,
            metric_lower_bound, metric_upper_bound,
        )
        batch_mean_estimates, batch_std_estimates = _compute_batch_estimates(
            risk_y_true, risk_y_proxy, batch_codes, power_tuning
        )
        risk_running_means, risk_lower_bounds = _compute_asymptotic_bounds(
        )
        running_means, confidence_bounds, metric_batch_estimates = _postprocess(
            risk_running_means, risk_lower_bounds, batch_mean_estimates,
            higher_is_better, metric_lower_bound, metric_upper_bound,
        )
            running_mean_estimates=running_means,
        )
            metric_name=metric_name,
            alarm_threshold=threshold,
            batch_mean_estimates=metric_batch_estimates,
            batch_n_true=batch_n_true,
            batch_n_proxy=batch_n_proxy,
        )
        return result
```

Export from `glide/monitors/__init__.py` and add a `:::` entry in `docs/api/monitors.md`.

**Unit tests.** `tests/unit/monitors/test_asymptotic_ppi.py` mirrors the `# --- detect ---` section of `test_empirical_ppi.py`, reusing its exact `y_true`/`y_proxy`/`batches`/`monitor` fixtures (the `_preprocess`/`_postprocess` sections have moved to `test_ppi_core.py`/`test_core.py`):

- `test_detect_delegates_to_validation`: patch `_validate_bounds` in the `asymptotic_ppi` module and check the single direct call with `(0.8, "confidence_level")`, `lower=0.5`, `upper=1`, both sides exclusive, and the `(0.5, 1)` error message.
- `test_detect_is_valid_monitoring_result`: `isinstance` checks for `PredictionPoweredMeanMonitoringResult` and `AsymptoticConfidenceSequence`, `monitor_name == "AsymptoticPPIMeanMonitor"`, finite running means, running means at or above the bounds.
- `test_detect_metadata`: mirror of the empirical version (metric name, orientation, threshold, confidence level, `batch_n_true`/`batch_n_proxy`).
- `test_detect_custom_confidence_level`: hardcoded anchor on the fixture stream (verified against a prototype run):

```python
def test_detect_custom_confidence_level(monitor, y_true, y_proxy, batches):
    expected_running_means = np.array([0.52, 0.52])
    expected_confidence_bounds = np.array([0.446, 0.477])

    result = monitor.detect(
        y_true, y_proxy, batches, higher_is_better=False, threshold=0.5, metric_name="risk", confidence_level=0.85
    )

    assert result.confidence_level == 0.85
    np.testing.assert_allclose(result.running_means, expected_running_means, atol=0.001)
    np.testing.assert_allclose(result.confidence_bounds, expected_confidence_bounds, atol=0.001)
```

## Functional tests

`tests/functional/monitors/test_asymptotic_ppi.py` mirrors `tests/functional/monitors/test_empirical_ppi.py`: same module docstring style, the same simulator-built `dataset` fixture (`generate_stratified_binary_dataset` plus `simulate_annotation`, 5 batches of 20 with 8 labels), `test_detect_*` names, and a one-line docstring per test.

**`test_detect_batch_mean_estimates_match_empirical_monitor`** — "The asymptotic and empirical monitors share the same per-batch estimates." Both monitors run on the `dataset` fixture; `batch_mean_estimates` must match exactly (they come from the same shared loop).

**`test_detect_prefix_consistency`** — "Detecting on a growing history is prefix-consistent with detecting on the full history." Mirror of the empirical test, but pass `tightest_at_batch=2` explicitly on both calls: the tuning target must already be reached within the prefix for the two histories to share the same ρ (see corner cases).

**`test_detect_higher_is_better_symmetry`** — "Monitoring a performance is the mirror image of monitoring its negation as a risk." Mirror of the empirical test.

## Corner cases

- `confidence_level <= 0.5`: raise `ValueError` naming the (0.5, 1) requirement (the empirical monitor accepts (0, 1); this monitor is stricter).
- Batches whose PPI estimates all have zero standard error up to `tightest_at_batch`: `ValueError` raised by the confidence sequence computation propagates.
- `power_tuning=False`: standard errors are still computed per batch with a tuning parameter of 1.0.
- Batch estimates outside [0, 1] in risk space (aggressive power tuning): no clipping; the sequence and the reported `batch_mean_estimates` use the raw values.
- Successive `detect` calls on growing histories shorter than `tightest_at_batch`: the clamped tuning target moves with the history, so the bounds of early batches recalibrate between calls; from batch `tightest_at_batch` onward they are prefix-consistent. Document this in the `detect` docstring (it matches the delayed-start framing of the asymptotic guarantee) rather than raising.

## Acceptance criteria

- [ ] `AsymptoticPPIMeanMonitor` is importable from `glide.monitors`, rendered in `docs/api/monitors.md`, and its doctest reproduces `first_alarm_index == 2` on the reference stream.
- [ ] `EmpiricalPPIMeanMonitor` produces bit-identical results after the `ppi_core` extraction (existing unit and functional tests pass with only import updates).
- [ ] `_compute_batch_estimates` returns both mean and standard-error arrays, and the empirical monitor path computes the same means as before.
- [ ] The unit tests mirror the `detect` section of `test_empirical_ppi.py` (same fixtures, same test names) and the custom-confidence anchor passes.
- [ ] The five functional tests above pass with the structure of `test_empirical_ppi.py` (dataset fixture, `test_detect_*` names, one-line docstrings).

imerad:
### Description

- Adds `AsymptoticPPIMeanMonitor`, a prediction-powered drift monitor built on `AsymptoticConfidenceSequence` that uses the batch estimates' standard errors for faster drift detection than the empirical-Bernstein based monitor.
- Extracts the logic shared with `EmpiricalPPIMeanMonitor` into `glide/monitors/ppi_core.py`.
- Handles this [ticket](https://github.com/orgs/EmertonData/projects/26/views/2?pane=issue&itemId=214000832&issue=EmertonData%7Cglide%7C350) Closes #350

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used

### DIFF
--- a/glide/monitors/asymptotic_ppi.py
+++ b/glide/monitors/asymptotic_ppi.py
@@ -9,7 +9,7 @@ from glide.monitors.ppi_core import _compute_batch_estimates, _preprocess
 
 
 class AsymptoticPPIMeanMonitor:
-    """Anytime-valid drift monitor leveraging the error bar of each batch estimate.
+    """Anytime-valid drift monitor leveraging the variance of each batch estimate.
 
     It computes a per-batch Prediction-Powered Inference (PPI) estimate: a small set
     of true labels and a large set of proxy labels, combined with a power-tuning
@@ -40,9 +40,9 @@ class AsymptoticPPIMeanMonitor:
     >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
     >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
     >>> post_drift_y_proxy = np.array([0.8, 1.0, 0.8, 1.0])
-    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
-    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
-    >>> batches = np.repeat(np.arange(51), 4)
+    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 5)])
+    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 5)])
+    >>> batches = np.repeat(np.arange(6), 4)
     >>> monitor = AsymptoticPPIMeanMonitor()
     >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
     >>> result.drift_detected
--- a/glide/monitors/asymptotic_ppi.py
+++ b/glide/monitors/asymptotic_ppi.py
@@ -1,40 +1,41 @@
-import numpy as np
 from numpy.typing import NDArray
 
-from glide.confidence_sequences import EmpiricalBernsteinConfidenceSequence
-from glide.confidence_sequences.empirical_bernstein import _compute_empirical_bernstein_bounds
+from glide.confidence_sequences import AsymptoticConfidenceSequence
+from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
+from glide.core.validation import _validate_bounds
 from glide.mean_monitoring_results import PredictionPoweredMeanMonitoringResult
 from glide.monitors.core import _postprocess
 from glide.monitors.ppi_core import _compute_batch_estimates, _preprocess
 
 
-class EmpiricalPPIMeanMonitor:
-    """Anytime-valid drift monitor over a batched dataset of true and proxy labels.
+class AsymptoticPPIMeanMonitor:
+    """Anytime-valid drift monitor leveraging the error bar of each batch estimate.
 
-    It uses a per-batch Prediction-Powered Inference (PPI) estimate: a small set of
-    true labels and a large set of proxy labels, combined with a power-tuning
+    It computes a per-batch Prediction-Powered Inference (PPI) estimate: a small set
+    of true labels and a large set of proxy labels, combined with a power-tuning
     parameter fitted on the batches that strictly precede it (predictable, as
-    required for validity). The monitor builds an anytime-valid confidence sequence
-    on the running mean of the per-batch PPI estimates, and raises a drift alarm
-    when it crosses a user-supplied ``threshold`` (the worst tolerable metric
-    value). Because the bounds are valid at all times simultaneously, :meth:`detect`
-    may be called after every new batch without inflating the false-alarm
-    probability.
+    required for validity), together with the standard error of that estimate. The
+    monitor tracks the running mean of the per-batch estimates against a
+    user-supplied ``threshold``, through a Gaussian-mixture asymptotic confidence
+    sequence whose width scales with the standard error of each batch estimate, so
+    genuine drifts get flagged early. The false-alarm guarantee is asymptotic: each
+    batch needs enough labeled and proxy samples for its PPI estimate to be
+    approximately Gaussian with a consistently estimated variance.
 
     References
     ----------
+    Waudby-Smith, Ian, David Arbour, Ritwik Sinha, Edward H. Kennedy, and Aaditya
+    Ramdas. "Time-uniform central limit theory and asymptotic confidence
+    sequences." The Annals of Statistics 52, no. 6 (2024): 2613-2640.
+
     Zhang, Guangyi, Yunlong Cai, Guanding Yu, and Osvaldo Simeone. "Prediction-powered
     risk monitoring of deployed models for detecting harmful distribution shifts."
     arXiv preprint arXiv:2602.02229 (2026).
 
-    Howard, Steven R., Aaditya Ramdas, Jon McAuliffe, and Jasjeet Sekhon. "Time-uniform,
-    nonparametric, nonasymptotic confidence sequences." The Annals of Statistics 49,
-    no. 2 (2021): 1055-1080.
-
     Examples
     --------
     >>> import numpy as np
-    >>> from glide.monitors import EmpiricalPPIMeanMonitor
+    >>> from glide.monitors import AsymptoticPPIMeanMonitor
     >>> pre_drift_y_true = np.array([0.0, 0.2, np.nan, np.nan])
     >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
     >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
@@ -42,12 +43,12 @@ class EmpiricalPPIMeanMonitor:
     >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
     >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
     >>> batches = np.repeat(np.arange(51), 4)
-    >>> monitor = EmpiricalPPIMeanMonitor()
+    >>> monitor = AsymptoticPPIMeanMonitor()
     >>> result = monitor.detect(y_true, y_proxy, batches, higher_is_better=False, threshold=0.5)
     >>> result.drift_detected
     True
     >>> result.first_alarm_index
-    11
+    2
     """
 
     def detect(
@@ -62,13 +63,14 @@ class EmpiricalPPIMeanMonitor:
         power_tuning: bool = True,
         metric_lower_bound: float = 0.0,
         metric_upper_bound: float = 1.0,
+        tightest_at_batch: int = 10,
     ) -> PredictionPoweredMeanMonitoringResult:
         """Detect a drift of the running mean across a batched dataset.
 
-        Splits the data by batch, computes a prediction-powered estimate per batch,
-        and builds an anytime-valid empirical-Bernstein confidence sequence on the
-        running mean of those estimates. An alarm is raised at every batch where the
-        sequence crosses the user-supplied ``threshold``.
+        Splits the data by batch, computes a prediction-powered estimate and its
+        standard error per batch, and builds an anytime-valid asymptotic confidence
+        sequence on the running mean of those estimates. An alarm is raised at every
+        batch where the sequence crosses the user-supplied ``threshold``.
 
         Rows must be ordered oldest batch first and grouped into contiguous blocks;
         identifier values are not compared, so any label type works. Batches must be
@@ -77,7 +79,10 @@ class EmpiricalPPIMeanMonitor:
         every call makes the anytime-valid guarantee hold jointly over all calls.
         Alternatively the data may be restricted to the most recent batches, in which
         case the guarantee holds within each restriction but not across the moving
-        history as a whole.
+        history as a whole. When the history has fewer batches than ``tightest_at_batch``,
+        the tuning target is set to the last batch. Prefix-consistency only holds between
+        calls using the same ``tightest_at_batch`` value and with histories containing
+        at least ``tightest_at_batch`` batches.
 
         Parameters
         ----------
@@ -108,7 +113,7 @@ class EmpiricalPPIMeanMonitor:
             raises "80% confident" alarms: under no drift it has at most a 20% chance
             of raising a false alarm. Raising this value demands more evidence before
             alarming, so alarms become more trustworthy but arrive later; lowering it
-            detects sooner at the cost of more false alarms.
+            detects sooner at the cost of more false alarms. Must be in ``(0.5, 1)``.
         power_tuning : bool, optional
             If ``True`` (default), compute the power-tuning parameter of each batch on
             all previous batches (the first batch, having no predecessor, uses
@@ -117,6 +122,11 @@ class EmpiricalPPIMeanMonitor:
             Known lower bound of the metric. Defaults to ``0.0``.
         metric_upper_bound : float, optional
             Known upper bound of the metric. Defaults to ``1.0``.
+        tightest_at_batch : int, optional
+            The batch index (1-indexed) at which the confidence sequence is tuned to
+            be tightest. Defaults to ``10``, which is safe to leave alone: this
+            setting affects tightness only, never the false-alarm guarantee, and the
+            width penalty for being off by a factor of 10 is below 10%.
 
         Returns
         -------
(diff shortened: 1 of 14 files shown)


imerad:
## Self-review follow-up: full list of issues found (including minor)

I ran `make lint`, `make type-check`, `make unit-tests`, `make functional-tests`, `make coverage`, and `mkdocs build --strict` locally — all pass, 100% coverage. Digging further for anything not caught by the gates, here's everything I found, most to least significant:

1. **Docstring bug: `AsymptoticPPIMeanMonitor.detect`'s `Raises` section is missing a real error condition.** `EmpiricalPPIMeanMonitor.detect`'s `Raises` list has a bullet "If any value falls outside `[metric_lower_bound, metric_upper_bound]`." covering the `y_true`/`y_proxy` bound checks performed in the now-shared `ppi_core._preprocess`. That bullet was dropped when writing the new docstring, but the underlying validation is unchanged and still fires. Confirmed live:
   ```python
   >>> AsymptoticPPIMeanMonitor().detect(
   ...     np.array([1.5, 0.2, np.nan, np.nan]), np.array([0.1, 0.2, 0.1, 0.2]),
   ...     np.array([0, 0, 0, 0]), higher_is_better=False, threshold=0.5,
   ... )
   ValueError: 'y_true' values must lie between 'metric_lower_bound'=0.0 and 'metric_upper_bound'=1.0; got values in [0.2, 1.5].
   ```
   This bullet should be added back to `asymptotic_ppi.py`'s docstring.

2. **Minor efficiency: `EmpiricalPPIMeanMonitor` now always computes a standard-error estimate it never uses.** The shared `ppi_core._compute_batch_estimates` always computes both `batch_mean_estimates` and `batch_std_estimates` per batch; `EmpiricalPPIMeanMonitor.detect` discards the latter (`risk_batch_estimates, _ = _compute_batch_estimates(...)`). This roughly doubles the per-batch computation for the empirical monitor versus before the refactor. Looks like a deliberate trade-off for sharing one loop (the ticket calls this out explicitly), so probably fine to accept, but it's a real, measurable extra cost for existing callers who don't need standard errors.

3. **Minor redundancy: `confidence_level` is validated twice on every call to `AsymptoticPPIMeanMonitor.detect`.** It's checked against `(0.5, 1)` at the top of `detect`, then checked again against the wider `(0, 1)` inside `ppi_core._preprocess`; the second check can never fail once the first passes. Not introduced by this PR — the identical pattern already exists in `AsymptoticClassicalMeanMonitor`/`classical_core.py` — so it's consistent with precedent, but both call sites carry the same avoidable redundancy.

4. **Minor duplication: `ppi_core._compute_batch_estimates` recomputes `n_batches`.** It derives `n_batches = batch_codes[-1] + 1` independently, duplicating `len(batch_identifiers)` already computed in `_preprocess`. Safe given `_unique_ordered_batches`'s contiguity guarantee, but could be passed in instead of recomputed.

5. **Inherent, pre-existing non-vectorized loop.** `ppi_core._compute_batch_estimates` iterates batches in a Python `for` loop (needed because each batch's power-tuning parameter depends on an expanding window over all strictly earlier batches), unlike `classical_core._compute_batch_estimates`, which is fully vectorized via `bincount`. Not a regression (the loop already existed in `EmpiricalPPIMeanMonitor` pre-refactor), just worth knowing the PPI monitors won't scale as well as the classical ones for very large batch counts.

Of these, only #1 is a real defect worth fixing before merge; the rest are minor/consistent-with-precedent observations.

glide/monitors/asymptotic_ppi.py
```diff
@@ -0,0 +1,206 @@
...
+from glide.confidence_sequences import AsymptoticConfidenceSequence
+from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
+from glide.core.validation import _validate_bounds
+from glide.mean_monitoring_results import PredictionPoweredMeanMonitoringResult
+from glide.monitors.core import _postprocess
+from glide.monitors.ppi_core import _compute_batch_estimates, _preprocess
+
+
+class AsymptoticPPIMeanMonitor:
+    """Anytime-valid drift monitor leveraging the error bar of each batch estimate.
```
gmartinon-ed: variance plutôt que "error bar" ?

glide/monitors/asymptotic_ppi.py
```diff
@@ -0,0 +1,206 @@
...
+    --------
+    >>> import numpy as np
+    >>> from glide.monitors import AsymptoticPPIMeanMonitor
+    >>> pre_drift_y_true = np.array([0.0, 0.2, np.nan, np.nan])
+    >>> pre_drift_y_proxy = np.array([0.0, 0.2, 0.0, 0.2])
+    >>> post_drift_y_true = np.array([0.8, 1.0, np.nan, np.nan])
+    >>> post_drift_y_proxy = np.array([0.8, 1.0, 0.8, 1.0])
+    >>> y_true = np.hstack([pre_drift_y_true, np.tile(post_drift_y_true, 50)])
+    >>> y_proxy = np.hstack([pre_drift_y_proxy, np.tile(post_drift_y_proxy, 50)])
+    >>> batches = np.repeat(np.arange(51), 4)
```
gmartinon-ed: On peut peut-être faire que 3 batchs si la dérive a lieu au batch 2 ?

gmartinon-ed: En tout cas c'est plutôt bon signe pour la suite ce "2"

glide/monitors/asymptotic_ppi.py
imerad: Appliqué

glide/monitors/asymptotic_ppi.py
imerad: On passe de 51 à 6 batchs, je propage aussi aux autres moniteurs

gmartinon-ed (approved):

PR #390, opened 2026-07-31
Fix landing page sidebar spacing and simulator defaults

imerad:
### Description

- Fixes the landing page sidebar so the GitHub/docs/PyPI/paper/newsletter/Emerton Data icon buttons sit a fixed distance below the logo instead of vertically centering in the leftover sidebar space (previously drifted lower on tall viewports).
- Adjusts default values on the interactive PPI simulator (default correlation now 0.77 instead of 0.81, the maximal possible value, plus updated default human sample size and estimate sliders).
- No issue tracks this; it's a standalone landing page polish pass.
- Noteworthy: we moved the icon buttons closer to the logo with a fixed margin (`margin-top: 2rem` on `.nav-btn-list`) rather than relying on `justify-content: space-between` on the sidebar, and chose 0.77 as the default correlation instead of 0.81, which was the maximal possible value.

### Checklist

- [ ] `make lint` passes
- [ ] `make type-check` passes
- [ ] `make tests` passes
- [ ] `make coverage` reports 100% coverage

### LLM usage

- [ ] No LLM used

### DIFF
--- a/docs/landing/index.html
+++ b/docs/landing/index.html
@@ -43,7 +43,10 @@
       padding: 2.5rem 1.75rem 2rem;
       display: flex;
       flex-direction: column;
-      justify-content: space-between;
+    }
+
+    .sidebar-footer {
+      margin-top: auto;
     }
 
     .main-panel {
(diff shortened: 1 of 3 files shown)


gmartinon-ed (changes requested):
Possible d'homogénéiser la font size des deux notes ? Au plus petit ?

gmartinon-ed (approved):

PR #397, opened 2026-08-04
Extract shared PPI/Classical engine for estimators and monitors

### LINKED ISSUE #393: PPRM Universal Refacto - PPI Chapter
## TODO

- [ ] Investiguer avec Claude si les PPIDataset sont vraiment nécessaires ou s'ils n'introduisent pas une abstraction superflue. Voir si la solution retenue se généralisera bien aux autres cas futurs (ASI, stratified, clusterd, multi)

## Background

Today the per-batch estimate logic lives in two places kept in sync by hand: the estimator body (`glide/estimators/ppi.py`, `glide/estimators/classical.py`, each with its own `_preprocess` plus calls into `glide/estimators/ppi_core.py`) and the monitor body (`glide/monitors/ppi_core.py`, whose `_compute_batch_estimates` re-derives the exact same PPI mean/std with a *predictable* tuning parameter fitted on prior batches). Every estimator family we later want to monitor (Multi-PPI, Stratified, Clustered, ASI) would duplicate this a third, fourth, fifth time.

The goal state: a single **engine** per estimator family that owns the whole per-batch body — validation, transformation, tuning, mean/std — reused unchanged by both the estimator (one unit = the whole dataset) and the monitor (one unit = one batch, tuning fitted on the prefix). This ticket is the **reference implementation** on the simplest pair, PPI and its classical counterpart; the four remaining pairs (Multi-PPI, Stratified, Clustered, ASI) are **out of scope** and follow in later tickets once this shape is approved.

**Prerequisites (assume both already merged):** the removal of `EmpiricalPPRM` and the empirical monitors' shared core, and the removal of risk-scaling from the monitors. This ticket is written as if neither exists: there is no empirical/asymptotic split to preserve, `glide/monitors/ppi_core.py` and `glide/monitors/classical_core.py` are free to delete, and the base performs no scaling.

## Design choices

- **Engine = the shared body, shell = the head.** A `MeanEstimateEngine` exposes three steps: `prepare(y_true, y_proxy, <family metadata>)` (validation + transformation as one atomic step, callable on any unit), `fit_tuning_parameter(dataset, power_tuning)`, and `compute_mean_and_std(dataset, tuning_parameter)`. The estimator shell keeps only confidence-interval construction, `effective_sample_size`, and result packaging; the monitor shell keeps only the batch loop and the confidence sequence. The compute cores in `glide/estimators/ppi_core.py` stay validation-free, array-based, and are reused verbatim by the PPI engine.
- **`prepare` is explicit, never a "batch" blob.** Each engine declares an explicit signature (`PPIMeanEngine.prepare(y_true, y_proxy)`, `ClassicalMeanEngine.prepare(y)`); the monitor base invokes it generically by slicing its named fields per batch and splatting them in prepare-signature order. Chosen over an opaque batch object so the estimator call sites stay readable and typed.
- **Engine location.** Engines live in a new top-level `glide/engines/` package, not under `glide/estimators/`, because the engine is shared by estimators *and* monitors: it is a peer concern, not an estimator sub-concern.
- **Engine naming.** The protocol is `MeanEstimateEngine` (it produces a mean *estimate*; it is not owned by "estimators", since monitors use it too — hence not `MeanEstimatorEngine`). Concrete engines are `PPIMeanEngine` / `ClassicalMeanEngine`, parallel to `PPIMeanEstimator` / `ClassicalMeanEstimator` and shorter than the redundant `PPIMeanEstimatorEngine`. The prepared value each engine carries is a small frozen dataclass named `<Family>Dataset` (`PPIDataset`, `ClassicalDataset`).
- **Engine is selected by monitor class, never inferred from arrays.** `AsymptoticPPRM` stays the PPI monitor; the classical counterpart is a separate class. A future stratified vs clustered monitor is disambiguated by *which class you instantiate*, not by guessing whether an ID column means strata or clusters (an integer ID array is structurally identical for both).
- **One shared monitor base, thin façades.** A base `AsymptoticRM` (in `glide/monitors/base.py`) owns the shared per-batch orchestration in a `_detect` template that returns the per-batch ingredients; each concrete monitor is an `engine` attribute plus a thin `detect` (the family's explicit public signature) that calls `_detect` and assembles the family result inline. Chosen so later families add a ~15-line façade, not a re-implemented loop. The plain name (no `Abstract`/`Base` prefix, no leading underscore) follows GLIDE's `base.py` convention (`ConfidenceSequence`, `MeanMonitoringResult`); neither prefix nor underscore-prefixed class names appear anywhere in the codebase.
- **Monitor naming — settle now, later tickets depend on it.** All asymptotic monitors carry the `RM` (risk monitor) suffix, resolving today's `PPRM` vs `ClassicalMeanMonitor` asymmetry: base `AsymptoticRM`, PP family `Asymptotic<Variant>PPRM` (`AsymptoticPPRM`, later `AsymptoticStratifiedPPRM`, `AsymptoticClusteredPPRM`, `AsymptoticMultiPPRM`, `AsymptoticActivePPRM` for ASI), classical family `Asymptotic<Variant>ClassicalRM` (base classical = `AsymptoticClassicalRM`, renamed from `AsymptoticClassicalMeanMonitor`). **The classical rename is a user-facing change; flag in the PR.**

### Engine layer (new `glide/engines/` package, internal — not re-exported from the top-level namespace)

`glide/engines/base.py` defines the protocol, generic over the dataset and tuning-parameter types so no `object`/`Any` appears. `prepare` is declared variadic in the protocol (all fields are `NDArray`, passed in a documented order); each concrete engine overrides it with an explicit named signature. No docstring: the protocol is internal.

`TuningParameterT` is a `TypeVar`, not `float`, because the tuning parameter genuinely changes shape across families and this protocol must serve all of them: `float` for PPI (scalar λ), `None` for classical, and `NDArray` for Multi-PPI and Stratified (the vector λ from `multi_ppi_core._compute_tuning_parameters`). The generic keeps `fit_tuning_parameter`'s output type tied to `compute_mean_and_std`'s input type per engine. Crucially, the base never inspects the tuning parameter — `_detect` threads it opaquely (`fit → compute`) — so a Multi-PPI vector λ will flow through `AsymptoticRM` unchanged, with no edit to the base. The same holds for the untuned value: it comes from `fit_tuning_parameter(dataset, power_tuning=False)` (`1.0` for PPI, `np.full(n_proxies, 1 / np.sqrt(n_proxies))` for Multi-PPI), never a hardcoded scalar in the base.

```python
from typing import Protocol, Tuple, TypeVar

DatasetT = TypeVar("DatasetT")
TuningParameterT = TypeVar("TuningParameterT")


class MeanEstimateEngine(Protocol[DatasetT, TuningParameterT]):
    def prepare(self, *fields: NDArray) -> DatasetT: ...

    def fit_tuning_parameter(self, dataset: DatasetT, power_tuning: bool) -> TuningParameterT: ...

    def compute_mean_and_std(self, dataset: DatasetT, tuning_parameter: TuningParameterT) -> Tuple[float, float]: ...
```

`glide/engines/ppi.py` wraps the existing PPI cores. `prepare` performs exactly the validation the estimator's `_preprocess` does today and returns a `PPIDataset`; `fit_tuning_parameter` and `compute_mean_and_std` unpack the dataset and call the untouched functions in `glide/estimators/ppi_core.py`.

```python
from typing import Tuple

from glide.core.validation import _validate_equal_lengths, _validate_sample_sizes, _validate_y_proxy, _validate_y_true
from glide.estimators.ppi_core import _compute_mean_estimate, _compute_std_estimate, _compute_tuning_parameter


@dataclass(frozen=True)
class PPIDataset:
    y_true: NDArray
    y_proxy_labeled: NDArray
    y_proxy_unlabeled: NDArray


class PPIMeanEngine:
    def prepare(self, y_true: NDArray, y_proxy: NDArray) -> PPIDataset:
        _validate_equal_lengths(y_true, y_proxy, names=["y_true", "y_proxy"])
        _validate_y_proxy(y_proxy)
        _validate_y_true(y_true)
        y_true_labeled, y_proxy_labeled, y_proxy_unlabeled, labeled_mask = _split_labeled_unlabeled(y_true, y_proxy)
        _validate_sample_sizes(labeled_mask)
        dataset = PPIDataset(y_true_labeled, y_proxy_labeled, y_proxy_unlabeled)
        return dataset

    def fit_tuning_parameter(self, dataset: PPIDataset, power_tuning: bool) -> float:
        tuning_parameter = _compute_tuning_parameter(
            dataset.y_true, dataset.y_proxy_labeled, dataset.y_proxy_unlabeled, power_tuning
        )
        return tuning_parameter

    def compute_mean_and_std(self, dataset: PPIDataset, tuning_parameter: float) -> Tuple[float, float]:
        mean = _compute_mean_estimate(
            dataset.y_true, dataset.y_proxy_labeled, dataset.y_proxy_unlabeled, tuning_parameter
        )
        std = _compute_std_estimate(
            dataset.y_true, dataset.y_proxy_labeled, dataset.y_proxy_unlabeled, tuning_parameter
        )
        return mean, std
```

`glide/engines/classical.py` mirrors this for the classical body. `fit_tuning_parameter` returns `None`: classical has no tuning. We keep it as an explicit one-liner rather than dropping it, because a default on the generic base returning `None` is unsound under the `TuningParameterT` typevar (PPI's is `float`), and the explicit `return None` documents that classical genuinely has no tuning while keeping one uniform interface the base can call.

```python
import numpy as np
from typing import Tuple

from glide.core.validation import _validate_min_samples


@dataclass(frozen=True)
class ClassicalDataset:
    y: NDArray


class ClassicalMeanEngine:
    def prepare(self, y: NDArray) -> ClassicalDataset:
        not_nan_mask = ~np.isnan(y)
        y_valid = y[not_nan_mask]
        _validate_min_samples(y_valid, "y")
        dataset = ClassicalDataset(y_valid)
        return dataset

    def fit_tuning_parameter(self, dataset: ClassicalDataset, power_tuning: bool) -> None:
        return None

    def compute_mean_and_std(self, dataset: ClassicalDataset, tuning_parameter: None) -> Tuple[float, float]:
        n_samples = len(dataset.y)
        mean = np.mean(dataset.y)
        std = np.std(dataset.y, ddof=1) / np.sqrt(n_samples)
        return mean, std
```

**Design note — cores stay array-based (do not pass the dataset into `_compute_*`).** The `glide/estimators/*_core.py` functions remain pure array functions and the engine unpacks the dataset into them. Reason, looking ahead: those cores are the single reuse point across families — `clustered_ppi.py` and `stratified_ppi.py` already call `ppi_core`'s `_compute_mean_estimate` with arrays of *cluster/stratum means*, not a `PPIDataset`. Coupling the core to a specific dataset dataclass would force every such caller to wrap-and-unwrap, and would invert the layering (a low-level core depending on an engine-layer type). Keeping the dataset an engine-layer concern and the cores pure-math keeps the separation clean and preserves the exact-numerics guarantee.

### Estimator shells delegate to the engine

Each estimator holds its engine (and, for the effective-sample-size baseline, the paired classical engine) as instance attributes set in `__init__`. The body becomes three engine calls; the head (CI + `effective_sample_size` + result) is unchanged in behaviour.

```python
class PPIMeanEstimator:
    def __init__(self) -> None:
        self._engine = PPIMeanEngine()
        self._classical_engine = ClassicalMeanEngine()

    def estimate(self, y_true, y_proxy, metric_name="Metric", confidence_level=0.95, power_tuning=True):
        """<unchanged PPIMeanEstimator docstring + doctest>"""
        dataset = self._engine.prepare(y_true, y_proxy)
        n_labeled, n_unlabeled = len(dataset.y_true), len(dataset.y_proxy_unlabeled)
        tuning_parameter = self._engine.fit_tuning_parameter(dataset, power_tuning)
        mean, std = self._engine.compute_mean_and_std(dataset, tuning_parameter)
        confidence_interval = CLTConfidenceInterval(mean=mean, std=std, confidence_level=confidence_level)

        classical_dataset = self._classical_engine.prepare(dataset.y_true)
        _, classical_std = self._classical_engine.compute_mean_and_std(classical_dataset, None)
        effective_sample_size = floor(n_labeled * classical_std**2 / std**2)
        # ... result packaging unchanged ...
```

`ClassicalMeanEstimator.__init__` sets `self._engine = ClassicalMeanEngine()`; its `estimate` collapses to `prepare -> compute_mean_and_std(dataset, None)` then CI + result.

**Effective-sample-size coupling (was: reinstantiate a whole `ClassicalMeanEstimator`).** The shell now reuses the paired **classical engine** for the baseline variance instead of building a full `ClassicalMeanEstimator` and reading its CI. `classical_std**2` is the classical variance; `std**2` is the PPI variance; `effective_sample_size = floor(n_labeled * classical_std**2 / std**2)`. This must reproduce today's `classical_ci.var / ppi_ci.var` exactly (both `.var` are the squared standard errors) — guarded by the unchanged `Effective Sample Size` doctest. It is the reproducible pattern for the later pairs: each PP shell reuses its paired classical engine (ASI ↔ IPW-classical, Stratified ↔ stratified-classical, etc.), keeping the coupling to one engine call and no result-object construction.

### Monitor base and façades

`AsymptoticRM` (new base in `glide/monitors/base.py`) owns `_detect`. It performs monitor-level validation that is *not* an estimator concern (equal lengths including `batches`, `batches` grouped in contiguous blocks via `_unique_ordered_batches`, `confidence_level` in `(0.5, 1)`, metric bounds, `threshold` in range), then runs the predictable per-batch loop and returns the per-batch ingredients (batch codes, mean estimates, confidence sequence) that each façade assembles into its result. Column data validation is delegated to `engine.prepare`. `fields` is the list of per-sample arrays the façade forwards to `engine.prepare`, in prepare-signature order; `_prepare_subset` mutualises the slice-and-prepare used for both the current batch and the tuning prefix.

The base is **generic over the engine's two types** — `AsymptoticRM(Generic[DatasetT, TuningParameterT])`, the same pair as the engine protocol — so the dataset flows out of `_prepare_subset` and the tuning parameter threads through the loop without a single `Any`. `_detect` does not build a result: it returns the per-batch ingredients (`batch_codes`, `batch_mean_estimates`, `confidence_sequence`), and each façade's `detect` assembles its family's result from them inline, exactly as the pre-refactor monitors did. That keeps each façade to a single method (no separate result-builder) and the base to two type parameters (no result typevar). Each façade binds the two type parameters in its class header (below).

```python
from typing import Generic, List, Tuple

from glide.engines.base import DatasetT, MeanEstimateEngine, TuningParameterT


class AsymptoticRM(Generic[DatasetT, TuningParameterT]):
    engine: MeanEstimateEngine[DatasetT, TuningParameterT]  # stateless class attribute set by each concrete façade

    def _prepare_subset(self, fields: List[NDArray], mask: NDArray) -> DatasetT:
        sliced_fields = [field[mask] for field in fields]
        dataset = self.engine.prepare(*sliced_fields)
        return dataset

    def _detect(self, fields: List[NDArray], batches: NDArray, threshold: float, confidence_level: float,
                metric_lower_bound: float, metric_upper_bound: float, tightest_at_batch: int,
                power_tuning: bool) -> Tuple[NDArray, NDArray, ConfidenceSequence]:
        # 1. monitor-level validation (lengths incl. batches, contiguity, confidence_level, metric bounds, threshold)
        batch_identifiers, batch_codes = _unique_ordered_batches(batches)
        batch_mean_estimates = np.empty(n_batches)
        batch_std_estimates = np.empty(n_batches)
            batch_dataset = self._prepare_subset(fields, batch_codes == position)
            if position == 0 or not power_tuning:
                tuning_parameter = self.engine.fit_tuning_parameter(batch_dataset, power_tuning=False)
            else:
                prefix_dataset = self._prepare_subset(fields, batch_codes < position)
                tuning_parameter = self.engine.fit_tuning_parameter(prefix_dataset, power_tuning=True)
            batch_mean_estimates[position], batch_std_estimates[position] = self.engine.compute_mean_and_std(
                batch_dataset, tuning_parameter
            )
        running_mean_estimates, lower_bounds = _compute_asymptotic_bounds(
        )
            running_mean_estimates=running_mean_estimates, confidence_bounds=lower_bounds
        )
        return batch_codes, batch_mean_estimates, confidence_sequence
```

Two things the base does deliberately:

- **No hardcoded `1.0`.** The untuned parameter comes from the engine: `fit_tuning_parameter(batch_dataset, power_tuning=False)` returns `1.0` for PPI (data-independent) and `None` for classical. The base never bakes in a PPI-specific constant.
- **Batch identity on failure.** `_prepare_subset` and `fit_tuning_parameter` are wrapped so any `ValueError` they raise (too few labeled/unlabeled in a batch; constant proxy across the prefix) is re-raised with the offending `batch_identifiers[position]` appended. Engines stay identifier-agnostic.

Each façade is thin: an `engine` attribute and a `detect` with the family's explicit public signature (carrying the **full existing monitor docstring and doctests**, which move here verbatim), which calls `_detect` for the shared computation and then assembles the family-specific per-batch counts and result type inline.

```python
class AsymptoticPPRM(AsymptoticRM[PPIDataset, float]):
    engine = PPIMeanEngine()

    def detect(self, y_true: NDArray, y_proxy: NDArray, batches: NDArray, higher_is_better: bool,
               threshold: float, metric_name: str = "Metric", confidence_level: float = 0.8,
               power_tuning: bool = True, metric_lower_bound: float = 0.0,
               metric_upper_bound: float = 1.0, tightest_at_batch: int = 10,
        """<full existing AsymptoticPPRM docstring + doctests, moved here verbatim>"""
        batch_codes, batch_mean_estimates, confidence_sequence = self._detect(
            fields=[y_true, y_proxy], batches=batches, threshold=threshold, confidence_level=confidence_level,
            metric_lower_bound=metric_lower_bound, metric_upper_bound=metric_upper_bound,
            tightest_at_batch=tightest_at_batch, power_tuning=power_tuning,
        )
        batch_n_true = np.bincount(batch_codes[labeled_mask], minlength=len(batch_mean_estimates))
        batch_n_proxy = np.bincount(batch_codes, minlength=len(batch_mean_estimates))
            metric_name=metric_name, monitor_name=self.__class__.__name__, higher_is_better=higher_is_better,
            alarm_threshold=threshold, confidence_level=confidence_level,
            batch_mean_estimates=batch_mean_estimates, confidence_sequence=confidence_sequence,
            batch_n_true=batch_n_true, batch_n_proxy=batch_n_proxy,
        )
        return result
```

The classical façade is the same shape with `engine = ClassicalMeanEngine()`, a single column `[y]`, `power_tuning=False` fixed (classical never tunes), and inline assembly of the single per-batch count into the classical monitoring result:

```python
class AsymptoticClassicalRM(AsymptoticRM[ClassicalDataset, None]):
    engine = ClassicalMeanEngine()

    def detect(self, y: NDArray, batches: NDArray, higher_is_better: bool, threshold: float,
               metric_name: str = "Metric", confidence_level: float = 0.8,
               metric_lower_bound: float = 0.0, metric_upper_bound: float = 1.0,
               tightest_at_batch: int = 10) -> ClassicalMeanMonitoringResult:
        """<full existing AsymptoticClassicalMeanMonitor docstring + doctests, moved here verbatim>"""
        batch_codes, batch_mean_estimates, confidence_sequence = self._detect(
            fields=[y], batches=batches, threshold=threshold, confidence_level=confidence_level,
            metric_lower_bound=metric_lower_bound, metric_upper_bound=metric_upper_bound,
            tightest_at_batch=tightest_at_batch, power_tuning=False,
        )
        batch_n = np.bincount(batch_codes, minlength=len(batch_mean_estimates))
            metric_name=metric_name, monitor_name=self.__class__.__name__, higher_is_better=higher_is_better,
            alarm_threshold=threshold, confidence_level=confidence_level,
            batch_mean_estimates=batch_mean_estimates, confidence_sequence=confidence_sequence, batch_n=batch_n,
        )
        return result
```

**Files affected:**

| File | What changes |
|---|---|
| `glide/engines/__init__.py` | New internal package init. |
| `glide/engines/base.py` | New `MeanEstimateEngine` protocol (generic, no docstring). |
| `glide/engines/ppi.py` | New `PPIMeanEngine` + `PPIDataset`, wrapping `glide/estimators/ppi_core.py`. |
| `glide/engines/classical.py` | New `ClassicalMeanEngine` + `ClassicalDataset`. |
| `glide/estimators/ppi.py` | Delete `_preprocess`; delegate body to `PPIMeanEngine`; reuse `ClassicalMeanEngine` for the ESS baseline. Public `estimate` signature and result unchanged. |
| `glide/estimators/classical.py` | Delete `_preprocess`; delegate body to `ClassicalMeanEngine`. |
| `glide/monitors/base.py` | New `AsymptoticRM` base with `_detect` (returns per-batch ingredients) and `_prepare_subset`. |
| `glide/monitors/asymptotic_pprm.py` | `AsymptoticPPRM` becomes a thin façade over `AsymptoticRM` + `PPIMeanEngine`. |
| `glide/monitors/asymptotic_classical.py` | `AsymptoticClassicalMeanMonitor` renamed to `AsymptoticClassicalRM`, thin façade over `AsymptoticRM` + `ClassicalMeanEngine`. |
| `glide/monitors/__init__.py` | Update the classical export name. |
| `glide/monitors/ppi_core.py`, `glide/monitors/classical_core.py` | **Deleted** — their `_preprocess`/`_compute_batch_estimates` logic is absorbed into `AsymptoticRM._detect` + the engines. (Safe: the empirical monitors that imported them are already gone per the prerequisite.) |
| `glide/core/validation.py`, `glide/estimators/stratified_core.py` | Rename the validator's `stratum_id` parameter to the generic `identifier` (see corner cases). |
| `CONTRIBUTING.md` | Three updates (see below). |
| `tests/unit/engines/test_ppi.py`, `tests/unit/engines/test_classical.py` | New: engine-level validation, tuning, mean/std against hardcoded expected values. |
| `tests/unit/estimators/`, `tests/unit/monitors/` | Existing tests stay; relocate assertions that tested logic now living in the engine down to `tests/unit/engines/`; keep estimator/monitor tests focused on the shell + delegation. Update the classical monitor test module for the rename. |
| `tests/functional/monitors/` | Keep the "per-batch estimate matches the one-shot estimator" property test; it now guards the engine reuse across both call paths. |

### CONTRIBUTING.md updates (this ticket owns them, since it introduces the pattern)

1. **Architectural overview:** add `engines/` to the directory tree and one sentence to "How the pieces fit together" — estimator and monitor shells delegate their per-unit body (validate → transform → tune → mean/std) to a shared `Engine`; the shell only packages the result.
2. **Recipes:** rewrite "Adding a new estimator" and "Adding a new monitor" to be engine-centric — implement one `Engine` (`prepare` / `fit_tuning_parameter` / `compute_mean_and_std`), then a thin estimator shell (CI + result) and a thin monitor façade (engine attribute + a `detect` that calls `_detect` and packages the result). Call out the exact-numerics doctest guard.
3. **New blurb** "Adding a new estimator family end-to-end" tying the three shells to one engine and stating that the engine is selected by monitor class, never inferred from input arrays.

## Corner cases

- **Exact numerics.** The refactor must not change a single output digit. Existing estimator doctests print hardcoded values (`Point Estimate: 5.618` in `glide/estimators/ppi.py`, `5.500` in `glide/estimators/classical.py`, plus `Effective Sample Size`) and the asymptotic monitor doctests print exact `first_alarm_index` values — all must pass unchanged. This is the primary regression net, and specifically guards the ESS rewrite (classical engine reused instead of a full estimator).
- **Generic identifier in validators.** Rename the `stratum_id` parameter of `_validate_y_proxy` (and any sibling validator) to the self-explanatory `identifier`, so batches, strata, and clusters all pass their context uniformly. Update the `glide/estimators/stratified_core.py` caller in the same change.
- **Batch identity on failure.** The base wraps each per-batch `prepare` / `fit_tuning_parameter` call and re-raises any `ValueError` with the offending batch identifier appended (too few labeled/unlabeled in a batch, or constant proxy across the prefix). Verify an empty/degenerate first batch and a constant-proxy prefix are both reported with the batch name, matching today's messages.
- **Prefix re-preparation cost.** `engine.prepare(prefix)` re-validates and re-splits a growing prefix each batch — mild O(n²). Readability wins here; leave it, optimise later if needed.
- **Type checker and the variadic `prepare`.** The protocol declares `prepare(self, *fields: NDArray)` while concrete engines use explicit named parameters. Confirm `ty` accepts the concrete signatures against the protocol; if it flags the override, type the base's `engine` attribute against the protocol and keep the concrete engines' explicit signatures for the estimator call sites.

## Acceptance criteria

- [ ] A `glide/engines/` package exists with `MeanEstimateEngine` (generic protocol, no `object`/`Any`), `PPIMeanEngine` + `PPIDataset`, and `ClassicalMeanEngine` + `ClassicalDataset`; engines are importable internally but not added to any public top-level `__all__`.
- [ ] `PPIMeanEstimator` and `ClassicalMeanEstimator` contain no per-batch statistical computation of their own: their `estimate` bodies obtain mean/std solely via their engine (set in `__init__`), and `_preprocess` is removed. `PPIMeanEstimator` computes `effective_sample_size` by reusing `ClassicalMeanEngine`, not by instantiating `ClassicalMeanEstimator`.
- [ ] `PPIMeanEngine.compute_mean_and_std` on a full dataset and `AsymptoticPPRM`'s per-batch computation on a single-batch dataset yield identical mean/std for the same inputs.
- [ ] `AsymptoticPPRM` and `AsymptoticClassicalRM` are thin façades over `AsymptoticRM`: each is an `engine` attribute plus a `detect` that assembles `fields`, calls `_detect`, and packages the family result inline (no separate result-builder method). The classical monitor is renamed from `AsymptoticClassicalMeanMonitor` and re-exported under the new name.
- [ ] All pre-existing estimator and monitor doctests pass with their printed numeric values unchanged (no doctest value is edited).
- [ ] A degenerate prior-batch (constant proxy across the prefix) and a too-small batch each raise with a message naming the offending batch.
- [ ] `glide/monitors/ppi_core.py` and `glide/monitors/classical_core.py` are deleted, their logic absorbed into `AsymptoticRM` + the engines.
- [ ] The `stratum_id` validator parameter is renamed to `identifier`, with all call sites updated.
- [ ] `CONTRIBUTING.md` reflects the engine-centric architecture and recipes, including the "add a family end-to-end" blurb and the class-selects-engine rule.

imerad:
### Description

Introduces a shared `MeanEstimateEngine` per family (PPI, Classical) in a new internal `glide/engines/` package: the per-unit body (validate, transform, tune, compute mean/std) is now written once and reused unchanged by both the estimator (one unit = whole dataset) and the monitor (one unit = one batch, via a new `AsymptoticRM` base in `glide/monitors/base.py`). `glide/monitors/ppi_core.py` and `classical_core.py` are deleted, their logic absorbed into the engines + `AsymptoticRM._detect`. This is the reference implementation on the simplest pair; Multi-PPI, Stratified, Clustered, and ASI follow in later tickets once this shape is approved.

Closes #393

**Noteworthy implementation decisions:**

- **Dropped `PPIDataset`/`ClassicalDataset` frozen dataclasses and the second `TuningParameterT` type variable from the ticket's design**, after discussing with the user. The prepared dataset is now a plain tuple of arrays (or a bare array for the single-field classical case) unpacked positionally, matching the existing `_split_labeled_unlabeled`-style convention already used elsewhere in the codebase, instead of introducing a bespoke dataclass per family. The tuning parameter's type collapses from a second generic parameter into one fixed `TuningParameter = Union[float, NDArray, None]` alias, since the shared base never inspects it and `ty` doesn't enforce the stricter per-engine typing either way (verified empirically). Both simplifications avoid compounding boilerplate once Multi-PPI/Stratified/Clustered/ASI each need their own engine.
- **Risk-scaling (`higher_is_better` reorientation) turned out to still be live in the codebase**, contrary to the ticket's assumption that its removal was already a merged prerequisite. Kept it, but relocated entirely into `AsymptoticRM._detect` (reorients the input fields going in, reorients the batch means/confidence sequence coming out), so each façade (`AsymptoticPPRM`, `AsymptoticClassicalRM`) stays pure assembly with no reorientation logic of its own.
- **`AsymptoticClassicalMeanMonitor` renamed to `AsymptoticClassicalRM`** — a user-facing breaking change, called out in the ticket and in `CHANGELOG.md`.
- The `CONTRIBUTING.md` rewrite has been watered down to avoid forcing new contributors to deal with the complexity of the new engine logic.

### Checklist

- [x] `make lint` passes
- [x] `make tests` passes

### LLM usage

- [ ] No LLM used


imerad:
### Code review

Found 1 issue:

1. `AsymptoticClassicalRM.detect()` pre-masks `y`/`batches` to labeled entries *before* calling the shared `AsymptoticRM._detect()`, but the length/NaN validation now happens *inside* `_detect()`, i.e. after that masking. This breaks two guarantees documented in the method's own `Raises` section: a length mismatch between `y` and `batches` now raises a raw `IndexError` ("boolean index did not match indexed array") instead of the documented `ValueError`, and a NaN in `batches` aligned with a NaN in `y` is silently dropped by the mask and never validated, so no error is raised at all. `AsymptoticPPRM.detect()` doesn't have this problem since it passes unmasked arrays into `_detect()`.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/asymptotic_classical.py#L122-L137

Compare with the shared validation call site, which now runs after masking has already occurred:

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/base.py#L38-L48

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

imerad:
### Code review (additional, lower-confidence findings)

Per request, here are the other issues surfaced during review that scored below the usual confirmation bar, included here as minor/FYI items rather than blocking:

1. **Undocumented behavior change: constant-value validation went from whole-series to per-batch/per-prefix for `AsymptoticPPRM`.** Before this PR, `_validate_y_true`/`_validate_y_proxy` (rejecting constant values) ran once on the full dataset. Now `AsymptoticRM._detect` calls `self.engine.preprocess(...)` separately per batch and per accumulated prefix, so a batch of 2 samples that happen to share a value (routine with small batches or low-cardinality metrics) now raises `ValueError` even though the overall series varies fine. Reproduced: the same input succeeds on `main` and raises on this branch. Not mentioned in the CHANGELOG or PR description, and not covered by a test.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/base.py#L16-L19

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/engines/ppi.py#L13-L19

2. **Wrapped error message no longer ends in a period.** `glide/monitors/base.py`'s batch-identity wrapper appends `" (batch '...')"` to validation errors that already end in `.`, producing messages like `"...in dataset. (batch '1')"` instead of terminating with a period, per CLAUDE.md's error-handling convention.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/base.py#L51-L56

3. **Batch-level "too few samples" message lost specificity.** The old per-monitor messages distinguished labeled vs. unlabeled counts and reported the actual number. The new shared `_validate_sample_sizes` message is generic ("Too few labeled or unlabeled samples in dataset.") and doesn't say which count was insufficient or by how much, which is a step back from CLAUDE.md's "always include the bad value" guidance.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/core/validation.py#L137-L145

4. **Naming inconsistency: `engine` (public) in monitors vs. `_engine`/`_classical_engine` (private) in estimators**, for the same internal concept, introduced by this PR.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/asymptotic_classical.py#L43

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/asymptotic_pprm.py#L56

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/estimators/classical.py#L30-L31

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/estimators/ppi.py#L46-L48

5. **`AsymptoticClassicalRM` lost its vectorized `bincount` batch computation**, now routed through `AsymptoticRM._detect`'s per-batch Python loop instead. Classical per-batch mean/std has no cross-batch dependency (unlike PPI's prefix-based tuning), so this is a real, if likely minor in practice, step back from CLAUDE.md's "vectorization whenever possible" — probably an acceptable tradeoff for the code-sharing this PR achieves.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/glide/monitors/base.py#L51-L65

6. **New test uses `assert_called_once_with` on numpy array arguments** instead of the `assert_called_once()` + `np.testing.assert_array_equal(...)` pattern CLAUDE.md requires for array args — the same test file uses the correct pattern a few lines below for other mocks.

https://github.com/EmertonData/glide/blob/12c1fc625db27b79ff3110af6e82a2599410a838/tests/unit/engines/test_ppi.py#L43-L50

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

glide/core/validation.py
```diff
@@ -163,12 +163,12 @@ def _validate_binary_or_nan(array: NDArray, name: str) -> None:
...
 
-def _validate_min_samples(values: NDArray, name: str, stratum_id: Optional[Hashable] = None) -> None:
+def _validate_min_samples(values: NDArray, name: str, identifier: Optional[Hashable] = None) -> None:
     if len(values) < 2:
-        if stratum_id is not None:
+        if identifier is not None:
             raise ValueError(
                 f"'{name}' must have at least 2 non-NaN values per stratum; "
-                f"got {len(values)} in stratum '{stratum_id}'."
+                f"got {len(values)} in stratum '{identifier}'."
```
gmartinon-ed: Ce mot "stratum" dans le message n'est-il pas un peu trop spécifique ? Il me semble que dans un contexte de monitoring, il peut s'agir d'un batch non ? Et dans le futur, qui des batchs stratifiés ?

glide/engines/base.py
```diff
@@ -0,0 +1,14 @@
+from typing import Protocol, Tuple, TypeVar, Union
+
+from numpy.typing import NDArray
+
+DatasetT = TypeVar("DatasetT")
```
gmartinon-ed, line 5: Pourquoi avoir gardé ce dataset type template et pas celui du tuning parameter ?

glide/estimators/clustered_ppi.py
```diff
@@ -4,9 +4,9 @@
 from numpy.typing import NDArray
 
 from glide.confidence_intervals import CLTConfidenceInterval
+from glide.engines.ppi_core import _compute_mean_estimate, _compute_std_estimate, _compute_tuning_parameter
```
gmartinon-ed, line 7: Je me pose une question sur ce déplacement  vers engines/ppi_core.py. Pourquoi ne pas avoir laissé dans estimators ? Ou pourquoi ne pas faire un glide/core avec dedans ppi.py ? Je me trompe peut-être mais j'avais l'impression qu'il fallait bien séparer le "core" des "engines", les "engines" jouant le rôle de middle layer entre le "core" et les classes user-facing. Là, je me demande si on ne fusionne pas la couche "core" et la couche "engine".

glide/engines/ppi.py
```diff
@@ -0,0 +1,29 @@
...
+from glide.core.utils import _split_labeled_unlabeled
+from glide.core.validation import _validate_equal_lengths, _validate_has_no_nan, _validate_sample_sizes
+from glide.engines.ppi_core import _compute_mean_estimate, _compute_std_estimate, _compute_tuning_parameter
+
+PPIDataset = Tuple[NDArray, NDArray, NDArray]
+
+
+class PPIMeanEngine:
+    def preprocess(self, y_true: NDArray, y_proxy: NDArray) -> PPIDataset:
+        _validate_equal_lengths(y_true, y_proxy, names=["y_true", "y_proxy"])
```
gmartinon-ed, line 14: Pourquoi `_validate_y_true` et `validate_y_proxy` sont à part et traités directement dans PPIMeanEstimator ?

glide/estimators/ppi.py
```diff
@@ -115,21 +100,26 @@ def estimate(
...
         """
-        y_true_filtered, y_proxy_labeled, y_proxy_unlabeled = self._preprocess(y_true, y_proxy)
-        n_labeled, n_unlabeled = len(y_true_filtered), len(y_proxy_unlabeled)
-        lambda_ = _compute_tuning_parameter(y_true_filtered, y_proxy_labeled, y_proxy_unlabeled, power_tuning)
-        mean = _compute_mean_estimate(y_true_filtered, y_proxy_labeled, y_proxy_unlabeled, lambda_)
-        std = _compute_std_estimate(y_true_filtered, y_proxy_labeled, y_proxy_unlabeled, lambda_)
+        _validate_y_proxy(y_proxy)
+        _validate_y_true(y_true)
+        dataset = self._engine.preprocess(y_true, y_proxy)
+        y_true_labeled, _, y_proxy_unlabeled = dataset
```
gmartinon-ed: Si je comprends bien, cette ligne sert juste à extraire les `n_labeled` et `n_unlabeled` utilisés plus bas c'est ça ? Dans ce cas, peut-on déplacer ces lignes juste avant l'usage de ces deux variables ? Et peut-on compacter avec des `len(dataset[0])` ? Ou tu penses que c'est moins lisible ?

glide/estimators/ppi.py
```diff
@@ -115,21 +100,26 @@ def estimate(
...
+        mean, std = self._engine.compute_mean_and_std(dataset, tuning_parameter)
         confidence_interval = CLTConfidenceInterval(
             mean=mean,
             std=std,
             confidence_level=confidence_level,
         )
-        classical_confidence_interval = ClassicalMeanEstimator().estimate(y_true_filtered).confidence_interval
-        effective_sample_size = floor(n_labeled * classical_confidence_interval.var / confidence_interval.var)
+        classical_dataset = self._classical_engine.preprocess(y_true)
+        _, classical_std = self._classical_engine.compute_mean_and_std(classical_dataset, None)
```
gmartinon-ed: Ici, ne peut-on pas capitaliser sur le preprocess déjà effectué de y_true par le ppi engine, et donner directement le dataset[0] ? Mais est-ce que ça se généralisera bien aux autres chapitres de la refacto ? En gros, la question que je me pose c'est : est-ce que le preprocessing de l'estimateur prediction-powered sur y_true est toujours totalement redondant avec le preprocessing de l'estimateur classique ?

glide/monitors/base.py
```diff
@@ -0,0 +1,62 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+from glide.confidence_sequences import AsymptoticConfidenceSequence
+from glide.confidence_sequences.asymptotic import _compute_asymptotic_bounds
+from glide.engines.base import DatasetT, MeanEstimationEngine
+from glide.monitors.core import _postprocess, _preprocess
+
+
+class AsymptoticRM(Generic[DatasetT]):
```
gmartinon-ed: A quoi sert la parenthèse ici ? En particulier le `Generic` ?

glide/engines/base.py
```diff
@@ -0,0 +1,14 @@
...
+TuningParameter = Union[float, NDArray, None]
+
+
+class MeanEstimationEngine(Protocol[DatasetT]):
```
gmartinon-ed: Que signifie `Protocol[DatasetT]` ? C'est un protocole paramétré par `DatasetT` c'est ça ?

glide/monitors/base.py
```diff
@@ -0,0 +1,62 @@
...
+    _engine: MeanEstimationEngine[DatasetT]
+
+    def _preprocess_subset(self, fields: List[NDArray], mask: NDArray) -> DatasetT:
```
gmartinon-ed, line 14: Proposition de renaming: `_preprocess_slice` ?

glide/monitors/base.py
```diff
@@ -0,0 +1,62 @@
...
+        batches: NDArray,
+        higher_is_better: bool,
+        confidence_level: float,
+        tightest_at_batch: int,
+        power_tuning: bool,
+    ) -> Tuple[NDArray, NDArray, AsymptoticConfidenceSequence]:
+        risk_fields, batch_identifiers, batch_codes = _preprocess(
+            fields, field_names, batches, higher_is_better, confidence_level
+        )
+        n_batches = len(batch_identifiers)
```
gmartinon-ed, line 32: J'ai l'impression que `batch_identifiers` ne sert plus à grand-chose, à part à calculer `n_batches`, qui peut s'avoir avec un bête `np.unique(batch_codes)`. Peut-on simplifier l'output de `_preprocess` ?

glide/monitors/base.py
```diff
@@ -0,0 +1,62 @@
...
+                tuning_parameter = self._engine.fit_tuning_parameter(prefix_dataset, power_tuning=True)
+
+            batch_mean_estimates[position], batch_std_estimates[position] = self._engine.compute_mean_and_std(
+                batch_dataset, tuning_parameter
+            )
+
+        miscoverage = 1.0 - confidence_level
+        risk_running_means, risk_lower_bounds = _compute_asymptotic_bounds(
+            batch_mean_estimates, batch_std_estimates, miscoverage, tightest_at_batch
+        )
```
gmartinon-ed, line 54: Ici je me mélange un peu les pinceaux en lisant, car notre règle de "préfixe `risk_` signifie normalisé" n'est pas appliquée de manière cohérente, notamment les `batch_mean_estimate` sont dans l'espace normalisé, mais leur nom ne l'indique pas. Donc, je me demande s'il ne faut pas les préfixer `risk_` aussi, ou si tout simplement on abandonne ce préfixe et cette convention, et c'est juste caché dans preprocess et postprocess

glide/monitors/base.py
```diff
@@ -0,0 +1,62 @@
...
+        running_means, confidence_bounds, metric_batch_mean_estimates = _postprocess(
```
gmartinon-ed: idem, ce préfix `metric_` est confusant ?

glide/monitors/asymptotic_classical.py
```diff
@@ -115,48 +115,33 @@ def detect(
...
+        labeled_mask = ~np.isnan(y)
+        batch_n = np.bincount(batch_codes[labeled_mask], minlength=len(batch_mean_estimates))
         result = ClassicalMeanMonitoringResult(
             metric_name=metric_name,
             monitor_name=self.__class__.__name__,
             higher_is_better=higher_is_better,
             alarm_threshold=threshold,
             confidence_level=confidence_level,
-            batch_mean_estimates=metric_batch_estimates,
+            batch_mean_estimates=batch_mean_estimates,
```
gmartinon-ed, line 148: Pour être honnête, je me pose une question: à qui ça sert de mettre les `batch_mean_estimates` dans le monitoring result ? Par exemple, est-ce qu'on s'en sert dans les tutoriels ou dans la validation scientifique, en lisant les attributs du monitoring result ?

Je me pose aussi la question du `batch_codes` en sortie de `_detect`: on dirait qu'on ne s'en sert jamais -> peut-on simplifier la sortie de `_detect` ? En vrai c'est trop bizarre comme type de sorte, un `Tuple[NDArray, NDarray, ConfidenceSequence`...

glide/monitors/asymptotic_classical.py
```diff
@@ -115,48 +115,33 @@ def detect(
...
+            fields=[y],
+            field_names=["y"],
+            batches=batches,
+            higher_is_better=higher_is_better,
+            confidence_level=confidence_level,
+            tightest_at_batch=tightest_at_batch,
+            power_tuning=False,
         )
+        labeled_mask = ~np.isnan(y)
+        batch_n = np.bincount(batch_codes[labeled_mask], minlength=len(batch_mean_estimates))
```
gmartinon-ed: J'ai l'impression que c'est le seul endroit où l'on se sert de `batch_mean_estimates`, peut-on compter les observations en s'en passant ? En fait je me demande plus généralement: est-ce que c'est décomptes `batch_n` et compagnie ne devraient pas être calculés dans `_detect` là où les tableaux sont vraiment traités ? Et dans ce cas, est-ce que ça ne devrait pas être directement dans l'objet `ConfidenceSequence` ?

glide/monitors/asymptotic_pprm.py
```diff
@@ -136,60 +136,36 @@ def detect(
...
+            field_names=["y_true", "y_proxy"],
+            batches=batches,
+            higher_is_better=higher_is_better,
+            confidence_level=confidence_level,
+            tightest_at_batch=tightest_at_batch,
+            power_tuning=power_tuning,
         )
+        labeled_mask = ~np.isnan(y_true)
+        batch_n_true = np.bincount(batch_codes[labeled_mask], minlength=len(batch_mean_estimates))
+        batch_n_proxy = np.bincount(batch_codes, minlength=len(batch_mean_estimates))
```
gmartinon-ed: Cf ma remarque précédent sur les décomptes. En plus je trouve que ces trois lignes sont inélégantes, un peu comme la salade qui dépasse du sandwich, entre nos beaux appels aux méthodes des engines

tests/unit/estimators/test_classical.py
```diff
@@ -20,22 +19,11 @@ def estimator() -> ClassicalMeanEstimator:
...
-    with patch("glide.estimators.classical._validate_min_samples") as mock_validate_min_samples:
-        estimator._preprocess(y_valid)
-    mock_validate_min_samples.assert_called_once()
-    np.testing.assert_array_equal(mock_validate_min_samples.call_args[0][0], y_valid)
-    assert mock_validate_min_samples.call_args[0][1] == "y"
+def test_init_sets_engine(estimator):
+    assert isinstance(estimator._engine, ClassicalMeanEngine)
 
 
 # --- estimate ---
```
gmartinon-ed, line 31: Remarque très générale ici : puisque tous les estimateurs, et tous les moniteurs, délèguent désormais à des engines, est-ce que ces tests ne doivent pas juste se transformer en des tests de délégation ? Et les vrais tests numériques vont dans les tests des engines ?

En tout cas, le fait que ces tests passent valident la refacto en soi, c'est déjà bien !

tests/unit/monitors/test_base.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.fixture
+def monitor():
+    return AsymptoticPPRM()
+
+
+# --- _preprocess_subset ---
+
+
+def test_preprocess_subset(monitor, y_true, y_proxy, batches):
```
gmartinon-ed: Idem: juste un test de délégation non ?

tests/unit/monitors/test_base.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@pytest.fixture
+def batches():
+    return np.array([0, 0, 0, 0, 1, 1, 1, 1])
+
+
+@pytest.fixture
+def monitor():
+    return AsymptoticPPRM()
```
gmartinon-ed: Bizarre ça, dans un fichier `test_base.py`, je pense que dans ce fichier, c'est surtout un sujet de délégation à tester

gmartinon-ed (changes requested):
Nice! Je me demande s'il ne faut pas aller un chouiat plus loin dans la refacto, notamment je tique sur ce type de sortie `Tuple[NDArray, NDarray, ConfidenceSequence]` parfaitement inélégant en sortie de `_detect`. ça fait un peu mélange de torchons et de serviettes.

Aussi je pense que la logique des tests doit être pas mal revue, car on introduit essentiellement une couche de délégation aux engines, et donc pas mal de tests doivent devenir des tests de délégation. A réfléchir avec Claude aussi, car je ne veux pas qu'en découpant les tests en tronçons indépendants (plutôt signe d'une bonne organisation modulaire), on perde de vue le test "fonctionnel" input/expected output sur la classe user-facing.

N.B. les arbitrages de conception qui sont faits dans cette review devront être consignés dans un markdown pour alimenter les prochains tickets de refacto de manière cohérente.

glide/engines/base.py
imerad: Finalement, on définit un type pour le tuning parameter aussi qui devra être fixé par les estimator/monitor comme le DatasetT. L'alternative serait de définir des types union qui incluent toutes les possibilités, mais ce n'est pas intéressant.

glide/estimators/clustered_ppi.py
imerad: Ces fonctions avaient initialement été laissées dans le module estimators, mais cela a provoqué un problème d'import circulaire entre estimators et engines. Claude l'avait rafistolé avec un import dans l'__init__ de PPIMeanEstimator. Je pense que le déplacement de ces fonctions vers engines est l'option la plus cohérente parce qu'elles sont réutilisées par AsymptoticPPRM. Un déplacement vers le module core correspondrait à un traitement de fonctions communément utilisées par les autres modules, cela s'applique bien à la logique de validation/split_labeled_unlabeled/inference result ... mais pas aux calculs spécifiques à PPI. L'autre option est d'inclure ces fonctions dans engines/ppi.py . Mais il faut anticiper qu'elles seront également utilisées par le futur ClusteredPPIMeanEngine. Ce dernier se retrouverait les à importer depuis engines/ppi.py, ce qui n'est pas propre. Une dernière option à laquelle je peux penser est un sous module engines/core/ppi.py, mais ça complique la hiérarchie du répertoire sans réel intérêt à mon avis. Compte tenu de tout ce qui précède, je recommande de rester sur cette organisation.

glide/engines/ppi.py
imerad: Les inférences exécutées par l'estimator et le monitor diffèrent. L'un utilise directement l'intégralité des données tandis que l'autre utilise des ensembles préfixes de batches. Ça change les contraintes à imposer sur les données en entrée. L'état actuel du code vise à factoriser les validations communes dans l'engine tandis que les validations spécifiques sont effectuées par l'estimator/monitor chacun de son côté (avec un peu de redondance pour éviter de disséquer chirurgicalement les fonctions de validation existantes).

glide/estimators/ppi.py
imerad: Je déplace ces lignes plus bas comme recommandé. Je pense qu'un `len(dataset[0])` serait effectivement moins lisible. D'ailleurs je renomme `dataset` en `ppi_dataset` pour éviter une confusion avec `classical_dataset`

glide/monitors/base.py
imerad: Il y a maintenant aussi un `TuningParameterT` dans les crochets. Le mot-clé `Generic` permet de déclarer la généricité en Python. Ça veut dire que la classe `AsymptoticRM` (ou une héritière) peut être instanciée avec un type concret représentant le `DatasetT` (et un autre pour `TuningParameterT`) et ce type concret doit prendre la place de `DatasetT` dans les signatures des méthodes de la classe pour permettre de vérifier le bon typage.

glide/engines/base.py
imerad: Oui, c'est comme ça que je le comprends. La classe `Protocol` est elle même générique et est, en l'occurrence, instanciée avec `DatasetT`

glide/monitors/base.py
imerad: Je crois qu'il vaut mieux garder `_preprocess_subset` car on a applique des masques arbitraires, hors un slice c'est par exemple un `array[start:end]`. Par contre, je renomme `sliced_fields` en `field_subsets`

glide/monitors/base.py
imerad: `batch_identifiers` est encore utilisé dans un éventuel message d'erreur. Malgré son apparente simplicité, `np.unique` peut avoir la complexité d'un tri donc on préfère l'éviter.

glide/monitors/base.py
imerad: En fait, il n'y a plus de normalisation depuis [cette PR](https://github.com/EmertonData/glide/pull/394) (fast track). Les fonctions `_preprocess` et `_postprocess` de monitors.core se contentent de réorienter si `higher_is_better=True`. Compte tenu de cela, le préfixe `risk_` garde du sens. Une convention de ce type est même nécessaire pour distinguer les inputs de l'appel à `_postprocess` de ses outputs. J'ai fait les renommages qu'il faut pour restaurer la cohérence.

glide/monitors/base.py
imerad: Il a été supprimé en faveur de mettre `risk` dans d'autres variables pour distinguer.

glide/monitors/asymptotic_classical.py
imerad: On ne se sert effectivement pas de `batch_mean_estimates` dans les notebooks, mais je pense que ça reste un tableau pertinent à fournir en sortie. En tout cas, ça ne vaut pas le coup de s'en débarrasser pour restreindre le return d'`AsymptoticRM._detect()` à un `ConfidenceSequence` car ce n'est pas possible.

Les `batch_codes` sont nécessaires pour calculer `batch_n` et `batch_n_true`/`proxy` dans `AsymptoticClassicalRM` et `AsymptoticPPRM`.

Ce code factorise les traitements communs à ces deux classes, ce qui donne ce type de sortie hétérogène. Pour y remédier, je propose de ne pas anticiper la création de l'objet `ConfidenceSequence` dans l'`AsymptoticRM` de base et retourner deux arrays en plus pour le créer dans les classes filles. Ça entraîne néanmoins une petite duplication de code.

glide/monitors/asymptotic_classical.py
imerad: En fait, les arguments `minlength` sont inutiles ici car on vérifie au préalable que tous les batchs ont au moins 2 labels donc je les enlève. Mais je pense que `batch_mean_estimates` reste un "nice to have".

Pour ce qui est de déplacer les décomptes de `batch_n` dans le `_detect` d'`AsymptoticRM`, je ne crois pas que ce soit possible car cette classe ne distingue pas le Classical et Prediction-Powered donc elles ne sait pas s'il faut calculer `batch_n` ou la paire `batch_n_true`/`proxy`. De plus, ces derniers sont des attributs de `ClassicalMeanMonitoringResult` et `PredictionPoweredMeanMonitoringResult`, ils ne sont pas destinés à `ConfidenceSequence` qui, elle aussi ne distingue pas classical/prediction-powered. Il ne faut pas oublier de préserver l'homogénéité avec les objets `MeanInferenceResult` dans ces choix.

glide/monitors/asymptotic_pprm.py
imerad: Les `minlength` sont inutiles ici aussi. C'est vrai que ces lignes supplémentaires ne sont pas élégantes, mais il me semble que ces traitements ne peuvent pas se factoriser dans la classe mère.

glide/estimators/ppi.py
imerad: Je remplace `classical_dataset` par `ppi_dataset[0]` qui fonctionne pour le coup. Je pense que ça ne se généralisera pas très naturellement pour les autres estimateurs car ils s'attendent à d'autres arrays. Il y aura probablement de la redondance, mais de manière adaptée au cas par cas. La question qui se pose à l'occasion : est ce qu'on peut créer une classe d'estimateur commune comme celle créée pour les monitors ? Malgré qu'il y a de forts rapprochements entre estimateurs comme PTD/IPW/Stratified, je pense qu'il y a peu de chose à factoriser globalement. Ça se fait bien avec les fonctions de ppi/ptd/stratified/clustered_core je trouve.

tests/unit/estimators/test_classical.py
imerad: Les tests de résultats numériques dans classical et ppi ont été remplacés par des tests de délégation

tests/unit/monitors/test_base.py
imerad: Traité

tests/unit/monitors/test_base.py
imerad: On n'a pas vrament d'autre choix que d'utiliser `AsymptoticPPRM` pour tester le comportement avec power tuning. À part ça, les tests de résultats numériques ont été remplacés par des tests de délégation.

glide/core/validation.py
imerad: Un paramètre identifier_type a été rajouté pour en tenir compte. Pour le moment, la valeur par défaut est "stratum" pour minimiser les instances de code à changer

glide/engines/classical.py
```diff
@@ -16,7 +16,7 @@ def preprocess(self, y: NDArray) -> NDArray:
     def fit_tuning_parameter(self, dataset: NDArray, power_tuning: bool) -> None:
         return None
 
-    def compute_mean_and_std(self, dataset: NDArray, tuning_parameter: Optional[float]) -> Tuple[float, float]:
+    def compute_mean_and_std(self, dataset: NDArray, tuning_parameter: None) -> Tuple[float, float]:
```
gmartinon-ed, line 19: Tiens je ne savais pas que c'était possible ça, ça fait quoi `tuning_parameter: None` ? On ne peut pas juste le virer ?

glide/engines/ppi.py
gmartinon-ed: Je ne vois pas pourquoi `_validate_y_true` et `_validate_y_proxy` ne peuvent pas aller dans PPIMeanEngine ? que ce soit un dataset tout seul ou plusieurs batchs, on veut quand même checker que les y_true et les y_proxy ne sont pas constants et que y_proxy ne contienne pas de NaN, non ?

glide/estimators/ppi.py
```diff
@@ -107,18 +107,19 @@ def estimate(
...
             confidence_level=confidence_level,
         )
-        classical_dataset = self._classical_engine.preprocess(y_true)
-        _, classical_std = self._classical_engine.compute_mean_and_std(classical_dataset, None)
+        _, classical_std = self._classical_engine.compute_mean_and_std(ppi_dataset[0], None)
+
+        y_true_labeled, _, y_proxy_unlabeled = ppi_dataset
+        n_labeled, n_unlabeled = len(y_true_labeled), len(y_proxy_unlabeled)
+
         effective_sample_size = floor(n_labeled * classical_std**2 / std**2)
```
gmartinon-ed, line 123: OK, à mettre juste après la ligne 118 alors, pour avoir un bloc entièrement cohérent et dédié à l'ESS

tests/unit/estimators/test_ppi.py
```diff
@@ -88,22 +102,6 @@ def test_estimate_metadata(estimator, y_arrays):
     assert result.effective_sample_size == 2
 
 
-def test_estimate_custom_confidence_level(estimator, y_arrays):
```
gmartinon-ed, line 91: Est-ce que ce test n'est pas finalement un test fonctionnel à garder, pour s'assurer qu'on teste bien le workflow utilisateur ?

tests/unit/estimators/test_classical.py
```diff
@@ -57,22 +63,6 @@ def test_estimate_metadata(estimator, y_array):
     assert result.n == 4
 
 
-def test_estimate_custom_confidence_level(estimator, y_array):
```
gmartinon-ed, line 60: Est-ce que ce test n'est pas finalement un test fonctionnel à garder, pour s'assurer qu'on teste bien le workflow utilisateur ?

gmartinon-ed (changes requested):
On converge, quelques tests fonctionnels à garder je pense.

glide/engines/classical.py
imerad: Le `tuning_parameter` fait seulement acte de présence ici car c'est le ClassicalMeanEngine. On a besoin de le garder pour respecter le protocole de MeanEstimationEngine

glide/engines/ppi.py
imerad: Finalement, j'ai enlevé `_validate_y_proxy` car c'était redondant. La vérification de non constance se fait déjà dans engines/ppi_core.py dans `_compute_tuning_parameter()` et l'absence de NaNs dans le preprocess de PPIMeanEngine.
Quant à `_validate_y_true`,  la vérification du nombre d'échantillons labellisés était redondante mais on doit vérifier que y_true n'est pas constant pour éviter un bug lors du calcul d'ESS (qui est spécifique au PPI estimator mais absent dans le monitor donc pas factorisable dans PPIMeanEngine). Donc j'ai remplacé par un appel à _validate_non_constant sur la partie labellisée.

glide/estimators/ppi.py
imerad: Fait

tests/unit/estimators/test_ppi.py
imerad: Restauré

tests/unit/estimators/test_classical.py
imerad: Restauré

gmartinon-ed (approved):
Sur les deux tests restaurés, est-ce qu'on ne les déplacerait pas dans functional/ ?

imerad:
Je pense que les tests restaurés restent unitaires en leur nature (vérification de résultats et attributs suite à l'appel d'une méthode sur une instance particulière). Ils diffèrent des tests fonctionnels qui eux vérifient des propriétés mathématiques sous-jacentes. Par conséquent, je préfère les garder dans unit/.

